PR #317, opened 2014-10-02
User-level documentation

linux2400:
Hi!
If anyone is interested in translating, I've written a short sphinx user manual for foodsoft in french.
It's been written from immediate and quite urging use in our new foodcoop in Toulouse, so it's far from exhaustive but I've tried to give a good tour of the main features.
Cheers
Romain


Nalaxon:
Hey there,
I would like to translate the doc, at least into german (since i dont't know french, and it would be usefull for the comunity i guess?)
What is the best way to go?
Forking linux2004'x repo and create a pull request?
Can you point me in a direction?

wvengen:
Hi @Nalaxon, that sounds like a great idea! There is already some German [documentation in the wiki](https://github.com/foodcoops/foodsoft/wiki/Doku), the easiest may be to add pages there.
If you'd rather have something like linux2004's setup, feel free to write something in that style. We'd need to setup some build-system in that case, that's why I'd start with the wiki.

Nalaxon:
Hi @wvengen,
it is not possible to add pictures to the wiki, right?
That what leads me to @linux2400s approach, although I face some problems with Ubuntu 14 in the latexpdf part :(
Also I would rather prefer *.md instead of *.rst (if there is a chance that FC-Wiki support markdown in the future).
I try to start with the wiki first, but in long term, I would really to set up the sphinx thing for easier multilanguage documentation

paroga:
I've created a [new repository](https://github.com/foodcoops/user-docs) for the documentation. I choose a general name for now, since this repo might also contain documentation for [sharedlists](https://github.com/foodcoops/sharedlists) in the future too.

PR #341, opened 2015-01-16
Fixed broken eager loading in orders#edit.

bennibu:
I guess there are more old active record API calls like this. I haven't looked further.


wvengen:
Well found, I guess this was overlooked when upgrading to Rails 4.

wvengen:
This was the only occurence I could find.

bennibu:
Thanks for looking into It.

PR #345, opened 2015-02-05
Add possibility to add an attachment to an invoice

paroga:


wvengen:
Thanks for your pull request! I'll test it later (just came back from a holiday).

wvengen:
Clear and simple!

Have you thought about using [paperclip](https://github.com/thoughtbot/paperclip) or [carrierwave](https://github.com/carrierwaveuploader/carrierwave)? These avoid security problems (XSS) and doesn't let the database grow so much (by storing files outside of it).

paroga:
i had a quick look at it, but didn't want to deal with additional dependencies. storing data outside of the DB makes the backup a little bit more complicated...
if you think it is better I can give it an additional try

wvengen:
I agree about the backups, but I'm not afraid of additional dependencies (especially when attachments will be used for other things like wiki and product images). Still: 3 invoices * 2MB * 50 weeks = 300MB /year, while current database dumps are <1MB /year. Do we want this? I'm open to any option, whatever makes most sense to us. What do you think?

What about restricting mime-types to safe ones? Allowing html as attachment allows people to elevate admin priviliges when an admin views the attachment (spoofed page that submits form adding the member to the admin group).

paroga:
I see no big difference in storing it to DB vs disk. People who do not add attachments still get 1MB/year, the others have to handle the additional data in both cases.
I personally like the idea of only caring about the DB dump for backups.

I think providing an option in the settings page for defining allow mime types would be not a bad idea. Maintaining a white-list in the code would add some unnecessary maintenance.

wvengen:
Sounds good, thanks for putting some thoughts to it :)

I'd be for providing a default-safe list of mime-types with the possibility to override it in the configuration. And add a configuration option when we find that people want to change it in the configuration screen in practice. (I don't think most people would fully understand what the option exactly means.)

wvengen:
Would you be up for adding an integration test?
Also, I'm missing the option to remove an existing attachment.
For the rest, I'll happily test and merge it.

paroga:
I did not find any tests for the invoices and do not want to start from scratch writing tests... :-(
I thought about uploading an empty file as action for removing an attachment, but the whole usability of the attachments is not perfect IMHO.

wvengen:
I understand. When I have time, I'll start writing some invoice tests. Since this is something we implement ourselves, I'd like to test this.

Deleting existing attachments is something I wouldn't want to go without, preferably. Would [this](http://stackoverflow.com/questions/4435826/rails-paperclip-how-to-delete-attachment) be a starting point, perhaps?

paroga:
Checkbox for deletion makes sense to me.
Are we sure that we want to store the data in the database now?

wvengen:
Database is fine with me!

paroga:
Do we want to store the data of #353 then in the DB too? Do we want two different mechanisms?

wvengen:
Fine with me!

wvengen:
What's pending here is security - same as https://github.com/foodcoops/foodsoft/pull/353#issuecomment-91599083

paroga:
ping

app/controllers/finance/invoices_controller.rb
```diff
@@ -62,6 +62,13 @@ def destroy
     redirect_to finance_invoices_url
   end
 
+  def attachment
+    @invoice = Invoice.find(params[:invoice_id])
+    type = MIME::Types[@invoice.attachment_mime].first
+    filename = 'attachment_' + @invoice.id.to_s + '.' + type.preferred_extension
```
wvengen: would `invoice_#{@invoice.id}_attachment.#{type.preferred_extension}` make more sense as filename?

wvengen:
Looking good, great! One tiny question, otherwise mergable!

wvengen:
:+1:

Besides this, I added a conditional to only show the delete box when an attachment is present, and am updating the install documentation to include the new dependency.

PR #347, opened 2015-02-18
Show only suppliers with articles in the dropdown-menu for new orders

paroga:
When there are many suppliers to be able to select them for invoices,
the menu for creating new orders gets unclear.

### DIFF
--- a/app/views/orders/index.html.haml
+++ b/app/views/orders/index.html.haml
@@ -6,7 +6,7 @@
       = t '.new_order'
       %span.caret
     %ul.dropdown-menu
-      - Supplier.undeleted.order('suppliers.name ASC').each do |supplier|
+      - Supplier.where(id: Article.undeleted.select(:supplier_id).distinct).order('suppliers.name ASC').each do |supplier|
         %li= link_to supplier.name, new_order_path(supplier_id: supplier.id), tabindex: -1
 
 .well


wvengen:
Nice idea! Does this generate [N+1 queries](http://www.ibm.com/developerworks/library/os-railsn1/index.html) for the article count? In that case it'd be useful to preload the article count. Or, better, update the controller to only return suppliers that have articles. Could be done using group_by and having.

paroga:
Does group_by and having help, when I need to "join" from an additional table?
In SQL I'd do sth like SELECT * FROM suppliers WHERE id in (SELECT supplier_id FROM articles WHERE undeleted)

wvengen:
Plain SQL would also be an option:
```ruby
Supplier.where(id: Article.undeleted.select(:supplier_id).distinct)
```
Using group by and having:
```ruby
Supplier.joins(:articles).group(:supplier_id).having('COUNT(articles.id) > 0').last
```

Both materialize to one query. The first was faster for me.

app/views/orders/index.html.haml
```diff
@@ -6,7 +6,7 @@
       = t '.new_order'
       %span.caret
     %ul.dropdown-menu
-      - Supplier.undeleted.order('suppliers.name ASC').each do |supplier|
+      - Supplier.where(id: Article.undeleted.select(:supplier_id).distinct).order('suppliers.name ASC').each do |supplier|
```
wvengen: Putting these queries in the view is bad practice. I'll move it to the controller when merged.

PR #348, opened 2015-02-18
Add possibility to copy an order

paroga:
If a supplier has a long list of articles, which are all available, but
the order should contain only a subset of them, selecting them for every
order can take a long time. Starting with a copy of an existing order
can safe a lot of time.

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -63,6 +63,7 @@ class OrdersController < ApplicationController
   # Page to create a new order.
   def new
     @order = Order.new(supplier_id: params[:supplier_id]).init_dates
+    @order.article_ids = Order.find(params[:order_id]).article_ids if params[:order_id]
   end
 
   # Save a new order.
--- a/app/views/orders/new.html.haml
+++ b/app/views/orders/new.html.haml
@@ -1,10 +1,10 @@
 - title t('.title')
 
 = render 'form'
-
-- content_for :javascript do
-  :javascript
-    // select all articles by default
-    $(function() {
-      $('#checkall').click();
-    });
+- if @order.article_ids.empty?
+  - content_for :javascript do
+    :javascript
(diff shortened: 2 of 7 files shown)


app/controllers/orders_controller.rb
```diff
@@ -63,6 +63,11 @@ def show
   # Page to create a new order.
   def new
     @order = Order.new(supplier_id: params[:supplier_id]).init_dates
+    @checkall_articels = true
+    if params[:order_id]
+      @base_order = Order.find(params[:order_id])
+      @checkall_articels = false
+    end
```
wvengen: Could we also just set it with something like:
```ruby
if params[:order_id]
  base_order = Order.find(params[:order_id])
  @order.article_ids = base_order.article_ids
end
```

app/views/orders/new.html.haml
```diff
@@ -1,10 +1,10 @@
...
 
 = render 'form'
-
-- content_for :javascript do
-  :javascript
-    // select all articles by default
-    $(function() {
-      $('#checkall').click();
-    });
+- if @checkall_articels
```
wvengen: this could then become: `if @order.article_ids.empty?`

wvengen:
Nice idea :)

app/controllers/orders_controller.rb
```diff
@@ -63,6 +63,7 @@ def show
   # Page to create a new order.
   def new
     @order = Order.new(supplier_id: params[:supplier_id]).init_dates
+    @order.article_ids = Order.find(params[:order_id]).article_ids if params[:order_id]
```
wvengen: even better :)

Tamriel:
Where can I find this function in the UI?

wvengen:
In the "manage orders" screen.
foodsoft_-_manage_orders_-_2015-07-30_20 01 01

PR #349, opened 2015-02-18
Use FoodsoftConfig[:email_sender] for sender address

paroga:
Not every email provider allows sending messages from an unauthorized server.

### DIFF
--- a/plugins/messages/app/mailers/messages_mailer.rb
+++ b/plugins/messages/app/mailers/messages_mailer.rb
@@ -6,6 +6,7 @@ class MessagesMailer < Mailer
 
     mail subject: "[#{FoodsoftConfig[:name]}] " + message.subject,
          to: recipient.email,
-         from: "#{show_user(message.sender)} <#{message.sender.email}>"
+         from: "#{show_user(message.sender)} via Foodsoft <#{FoodsoftConfig[:email_sender]}>",
+         reply_to: "#{show_user(message.sender)} <#{message.sender.email}>"
   end
 end


wvengen:
Good point: setting the from address for mails from messages is indeed necessary. Still, this breaks replying.

* At the very least: make it clear in the email that replying can only be done online, not by mail.
* Perhaps we can set the Reply-To header, would suffice?
* The most beautiful would be to create an email responder as part of Foodsoft, which processing email replies.

paroga:
What does it break exactly? With the Reply-To header a conforming mail client should work like before...

wvengen:
Oh, somehow I missed that. Sorry!

plugins/messages/app/mailers/messages_mailer.rb
```diff
@@ -6,6 +6,7 @@ def foodsoft_message(message, recipient)
 
     mail subject: "[#{FoodsoftConfig[:name]}] " + message.subject,
          to: recipient.email,
-         from: "#{show_user(message.sender)} <#{message.sender.email}>"
+         from: "#{show_user(message.sender)} via Foodsoft <#{FoodsoftConfig[:email_sender]}>",
```
wvengen: I18n?

paroga: really? isn't "via" universal?

wvengen: For the handful of languages we have right now, yes. In general, no.

paroga: I'd suggest keeping it hardcoded for now, since we have an even bigger problem: The text depends on the language of the sender and not of the recipient.

wvengen: True! Let's keep it like this for now.

PR #350, opened 2015-02-18
Add messaging to order (closes #327)

### LINKED ISSUE #327: Send mail to order participants
A user from foodcoops.net asked for an easy solution, to write an mail to everyone, who participates or hast parcipated in a certain order. At the moment, there is one way only. Write the names manually from the orders show view and copy them into the new message form.

Maybe a simple button on the order show view, which redirects to the message form with all participants preselected should be sufficient.

paroga:


wvengen:
Yay, very good idea!

Messaging is done in [a plugin](https://github.com/foodcoops/foodsoft/tree/master/plugins/messages), so this would need to go into a deface override. See [this example](https://github.com/foodcoops/foodsoft/blob/master/plugins/messages/app/overrides/foodcoop/users/_users/new_message.html.haml.deface).

plugins/messages/app/models/message.rb
```diff
@@ -40,6 +40,15 @@ def group_id=(group_id)
     add_recipients Group.find(group_id).users unless group_id.blank?
   end
 
+  def order_id=(order_id)
+    @order_id = order_id
+    unless order_id.blank?
```
wvengen: why this check?

paroga:
ad deface) i know, but I wasn't able to get it working :-/
ad check) yes, it's not necessary any more

wvengen:
What did you try with deface?

paroga:
create a file in orders/show folder with similar content of your example

wvengen:
Ok, I'm trying now with deface - and run into the old bug where `:javascript` in the original breaks it - https://github.com/spree/deface/issues/125. Solution would be to move the javascript in orders/show.html.haml to a separate file. I'm on it.

wvengen:
Oh, and why not making the "x Ordergroups" a messaging link instead of the mail button?

paroga:
Yes, that was exactly the error I got.
The "x Ordergroups" "expands" to the list of Ordergroups. How do you want to handle that then?

wvengen:
That's the title attribute, a hyperlink could go to the messaging screen.

paroga:
If I do it that way I get some kind of "layering problem": a create_mail_link_for_ordergroups function should only need a list of order groups as parameter. The message page requires the order_id instead and passing a list of order groups/users via URL does not seam like a perfect solution for me either.

wvengen:
I like the design pattern that info displayed on the screen is also a way to interact with it. In this way: seeing the ordergroups allows one to do something with it - sending them a message.

I think the technical problem of how to create the link is solvable. We can still use the order_id (since that can be used to get the same list of ordergroups).

wvengen:
Thanks for your work! Testing this, I'm a bit confused on the message page, since it shows no addressee. I see a number of options:
* Include all recipients in the recipients field
* Add a special field "Order ..." to the Groups field
* Replace the "Order ..." (and recipients?) fields with "Order: ...."
* Allow to put members, groups and orders in the recipients input field (something I'd like to do in the future).

paroga:
at least on my installation it adds all recipients to the recipients field. any idea? i had some "browser caching" problems at the beginning too, but they were gone later :-/

wvengen:
Ok, I'll try again. The code looks ok in that respect.
If you feel like it, you might want to `git pull && git rebase master`

plugins/messages/app/models/message.rb
```diff
@@ -40,6 +40,13 @@ def group_id=(group_id)
     add_recipients Group.find(group_id).users unless group_id.blank?
   end
 
+  def order_id=(order_id)
+    @order_id = order_id
+    for ordergroup in Order.find(order_id).ordergroups
+      add_recipients ordergroup.users
+    end
```
wvengen: This would be shorter and more performant

```ruby
add_recipients Order.find(order_id).ordergroups unless order_id.blank?
```

wvengen: Silly me, that wouldn't work at all. Perhaps another `has_many :through` in Order would work, but let's leave it this way :o

wvengen: Ok, I couldn't resist. If you like:
```ruby
# in app/models/order.rb
has_many :ordergroups, :through => :group_orders
has_many :users_ordered, :through => :ordergroups, :source => :users
```

```ruby
# in plugins/messages/app/models/message.rb#order_id=
add_recipients Order.find(order_id).users_ordered if order_id
```

PR #351, opened 2015-02-18
Unpaid invoices for everybody

paroga:


app/views/finance/invoices/show.html.haml
```diff
@@ -35,6 +35,7 @@
   %b= heading_helper(Invoice, :note) + ':'
   =h @invoice.note
 
-= link_to t('ui.edit'), edit_finance_invoice_path(@invoice)
-|
+- if current_user.role_finance? or (current_user.role_invoices? and not @invoice.paid_on)
```
wvengen: It looks like this could move to a helper method.

app/views/finance/invoices/_invoices.html.haml
```diff
@@ -26,6 +26,10 @@
         %td= link_to Delivery.model_name.human, [invoice.supplier,invoice.delivery] if invoice.delivery
         %td= link_to format_date(invoice.order.ends), new_finance_order_path(order_id: invoice.order_id) if invoice.order
         %td= truncate(invoice.note)
-        %td= link_to t('ui.edit'), edit_finance_invoice_path(invoice), class: 'btn btn-mini'
-        %td= link_to t('ui.delete'), finance_invoice_path(invoice), :data => {:confirm => t('.confirm_delete')}, :method => :delete,
-          class: 'btn btn-danger btn-mini'
+        %td
+          - if current_user.role_finance? or (current_user.role_invoices? and not invoice.paid_on)
```
wvengen: `and` and friends [are control flow operators](http://devblog.avdi.org/2010/08/02/using-and-and-or-in-ruby/), `&` and `!` would suit better here:
```ruby
if current_user.role_finance? || (current_user.role_invoices? & !invoice.paid_on)
```

wvengen: Do we want people to be able to edit each others' invoices? (Not a blocking issue, I'd say.)

wvengen:
Again, a nice feature. Thanks!

app/views/finance/invoices/_invoices.html.haml
paroga: IMHO everything before marking an invoice as paid by the finance people is "unsafe", so editing invoices of other people would be ok for me.

app/views/finance/invoices/show.html.haml
paroga: where should I put it? (and maybe a name for it ;-)) IMHO it's not a member function of User or Invoice...

app/views/finance/invoices/_invoices.html.haml
wvengen: @Cor-Jan not sure if you can make anything out of this discussion, but what do you think of this? The idea is to allow not just financial people to add invoices, and only allow financial people to set invoices to being paid. What would be better?
1. All people who have the invoices permission can edit all unpaid invoices.
2. All people who have the invoices permission can create new invoices and edit their own unpaid ones.

app/views/finance/invoices/show.html.haml
wvengen: That's what [helpers](http://api.rubyonrails.org/classes/ActionController/Helpers.html) can be used for: app/helpers/finance/invoice_helper.rb in this case.

app/views/finance/invoices/_invoices.html.haml
wvengen: Ok, my tension comes from this:
1. As an invoice-user, I create an invoice for something I bought for the foodcoop
2. Another invoice-user edits the invoice (and I'm not aware of it).
3. The finance people see the invoice accept it.
4. Later someone checks the invoices and finds an error.
5. I'm held accountable, since I created the invoice. (Or, if the finance people see something and contact me, I'm at loss why the invoice is different from what I entered before.)

This is a kind of worst-case scenario, but I'd like to avoid these things when possible. The occasional time that someone else needs to edit the invoice that I entered can go through communication with me.

Now perhaps you might work more in groups entering invoices, where the responsibility is shared. If that's so, I'd like it to be clear in the interface that the group is responsible (and others than the group can't edit it).

That's why I'd like to restrict editing to your own invoice. Or implement a system where it's clear who makes changes.

paroga: agree, I'll update the pull request

wvengen: thanks, and I'd be happy to merge when done

paroga:
committed in 1315103a7d4ed62f16d15bd6246fe4301a7282d1

PR #352, opened 2015-02-18
Add bank accounts

paroga:


wvengen:
Wow, using scraping to get the transactions. Since this is quite specific to the banks involved, and a feature not all foodcoops may want to use, I think this belongs in a plugin. See [Plugin development](https://github.com/foodcoops/foodsoft/wiki/Plugin-development). What do you think?

paroga:
Yes it should go into a plugin. Question is how to implement the plugin interface, make it useful for people without an implemented scraper and do the configuration of the credentials

paroga:
Maybe I should be a little more precise: The scraping should go into a plugin. The basic bank_account stuff should go into the foodsoft code, since I'd like to do a tight integration with financial_transactions and invoices (some kind of cross references), which will be nearly impossible with the whole implementation in a plugin.

wvengen:
Ah. Ideally, I'd see the scraping as a separate gem, that can be used by others projects as well. But let's first have something that works here. Did you check if there are existing projects? Until we have a separate gem, I would put it into `lib/banking` or something similar.

Personally, I'd start with uploading transactions files. That can also be used when scraping fails (which is bound to happen at some point when the bank changes its interface), and doesn't need credentials passing through Foodsoft. When that works, enhance it by scraping the bank website.

I can see that this binds deeply into Foodsoft, e.g. when matching transactions with ordergroups. I'm happy to develop this as part of Foodsoft core, and when it's ready decide if it belongs in a plugin or not.

wvengen:
e.g. [Banker](https://github.com/kylewelsby/Banker)

paroga:
I checked Banker and a few others too, but did not use them because the interface does not provide everything I need for my transactions and extending that libraries and finding a generic solution is too much work.

wvengen:
I see. It doesn't feel right for me to have it here, something as generic as scraping a bank website. But as a first step towards a general framework (preferably relating to others gems used by other projects), I'm fine. That probably means that if the banking framework catches on, it will be refactored quite a bit.
For now, fine. (So that you know what to expect :) )

paroga:
What do you think about the current schema? If it looks ok to you, I'll create a version with CSV import instead of the scarping stuff.

db/schema.rb
```diff
@@ -66,6 +66,31 @@
...
+    t.string  "booking_type"
+    t.string  "booking_number"
+    t.text    "text"
+    t.string  "iban"
+    t.string  "bic"
+    t.string  "name"
+    t.string  "reference"
+    t.text    "receipt"
+    t.binary  "image", limit: 1.megabyte
+  end
```
wvengen: Looks good to me in general. Some questions:
* What are receipt and image for?
* 'reference' is called 'memo' in OFX, which I find slightly broader (it may contain general comments too, right?).
* What's the different between booking and value date?

wvengen:
Sounds good - than we can add a scraper later when all is in place. This could be quite a big change - putting banking information in Foodsoft, I don't quite grasp the consequences of this: if the information is there, what makes sense to do with it? A double accounting system? Would be great, but not really my area of expertise.
In any case, I'm excited to see how we can improve Foodsoft at the financial side.

paroga:
could you give me some basic feedback? (IMHO only the first two commits are ready for integration, but the other show already what I'll want to build on top of it)

wvengen:
First commit looking good! Notes:
* Some code could be shared with `lib/foodsoft_file.rb`, it would be good to have uniform handling of spreadsheet uploads.
* Does the API mention anything about currencies?
* I'd love to have clear documentation of the properties, especially since this is about interaction with other systems that developers need to reference outside of the codebase (e.g. see foodcoop-adam's [financial_transaction.rb](https://github.com/foodcoop-adam/foodsoft/blob/beta/app/models/financial_transaction.rb).
* Is a bank account's `last_import_id` the same as the bank's last transaction's `import_id`?
* For this functionality, I consider specs to be essential. I wouldn't want transactions to be missed after a refactoring.
* What about starting to use a bootstrap form layout with static texts for show?

app/models/bank_account.rb
```diff
@@ -34,4 +34,7 @@ def import_from_file(file, options = {})
     end
     count
   end
+
+  def find_import_method
+  end
```
wvengen: explicit `nil` or at least a comment that explains what would happen here?

wvengen:
And of course, I'd expect bank communication methods to reside in a plugin (unless there's a common standard for that, perhaps). I'm not yet sure how to fit in `assign_unchecked_transactions` - it's quite specific; perhaps in a plugin and/or some general mechanism to assign bank transactions to ordergroup transactions.

paroga:
@wvengen I've addressed most of your points. Would be great if you could do an early review again.

app/models/bank_account.rb
```diff
@@ -0,0 +1,28 @@
+require 'roo'
```
wvengen: I think `SpreadsheetFile` would depend on this instead.

app/models/bank_account.rb
```diff
@@ -0,0 +1,28 @@
...
+
+class BankAccount < ActiveRecord::Base
+
+  has_many :bank_transactions, dependent: :destroy
+
+  normalize_attributes :name, :iban, :description
+
+  validates :name, :presence => true, :uniqueness => true, :length => { :minimum => 2 }
+  validates :iban, :presence => true, :uniqueness => true
+  validates_format_of :iban, :with => /\A[A-Z]{2}[0-9]{2}[0-9A-Z]{,30}\z/
```
wvengen: What about a full IBAN check with one of the gems (like [iban-tools](https://github.com/alphasights/iban-tools))?
I've seen users in the USA or Israël, do we need to think about them regarding IBAN? I guess being able to have it not (so) visible would be enough (and given enough interest, they could submit a PR to generalise is).

app/models/bank_account.rb
```diff
@@ -0,0 +1,28 @@
...
+  def import_from_file(file, options = {})
+    SpreadsheetFile.parse file, options do |row|
+      bank_transactions.where(:external_id => row[0]).first_or_create.update(
+                              :date => row[1],
+                              :amount => row[2],
+                              :iban => row[3],
+                              :text => row[4],
+                              :reference => row[5],
+                              :receipt => row[6],
+                              :image => row[7].nil? ? nil : Base64.decode64(row[7]))
```
wvengen: It would be good to document this file format, and provide an example download either in the docs or in Foodsoft.

app/views/finance/bank_transactions/_transactions.html.haml
```diff
@@ -0,0 +1,25 @@
...
+      %th= heading_helper(BankTransaction, :reference)
+      %th= sort_link_helper heading_helper(BankTransaction, :amount), "amount"
+      %th= sort_link_helper heading_helper(BankTransaction, :financial_link), "financial_link"
+  %tbody
+    - @bank_transactions.each do |t|
+      %tr
+        %td= h link_to format_date(t.date), finance_bank_transaction_path(t)
+        %td= h(t.text).gsub("\n", "<br>").html_safe
+        %td= h t.reference
+        %td.currency{:style => "color:#{t.amount < 0 ? 'red' : 'black'}; width:5em"}= number_to_currency(t.amount)
```
wvengen: Would be good to put this in CSS.

app/views/finance/bank_transactions/show.html.haml
```diff
@@ -0,0 +1,39 @@
+%p
+  %b= heading_helper(BankTransaction, :external_id) + ':'
+  = @bank_transaction.external_id
```
wvengen: I know we have this style elsewhere, I'd prefer some more declarative style with classes, or a real table. Could also be a bootstrap form with static 'input's. Unless styled, it somewhat hurts my eyes :s

app/models/bank_transaction.rb
```diff
@@ -33,4 +33,56 @@ class BankTransaction < ActiveRecord::Base
...
+      link = FinancialLink.new
+      invoices.each {|i| i.update_attributes financial_link: link, paid_on: date}
+      update_attribute :financial_link, link
+    end
+
+    return true
+  end
+
+  def assign_to_ordergroup
+    m = /FS(?<group>\d+)(\.(?<user>\d+))?(?<parts>([A-Za-z]+\d+(\.\d+)?)+)/.match(reference)
```
wvengen: Would it be good to include some kind of checksum here? A small typo may result in a big impact, for a user. Of course the reference would then need to be generated from Foodsoft.

app/controllers/finance/bank_accounts_controller.rb
```diff
@@ -5,6 +5,14 @@ def index
     redirect_to finance_bank_account_transactions_url(@bank_accounts.first) if @bank_accounts.count == 1
   end
 
+  def assign_unchecked_transactions
+    @bank_account = BankAccount.find(params[:id])
+    count = @bank_account.assign_unchecked_transactions
+    redirect_to finance_bank_account_transactions_url(@bank_account), notice: t('finance.bank_accounts.controller.import.notice', :count => count)
```
wvengen: I would somehow expect to be able to review all transactions. If I were a foodcoop beginning to use this feature, I would feel uneasy if the system does something I may not fully comprehend yet. Seeing what it does, and checking it, would make me feel a lot better about it.

app/models/bank_account.rb
```diff
@@ -28,4 +28,14 @@ def import_from_file(file, options = {})
...
   def find_import_method
   end
+
+  def assign_unchecked_transactions
+    count = 0
+    bank_transactions.without_financial_link.includes(:supplier, :user).each do |t|
+      if t.assign_to_ordergroup || t.assign_to_invoice
+        count += 1
+      end
+    end
```
wvengen: I wonder if this should be wrapped in a transaction: the iteration is done outside of the update transaction.

app/models/bank_transaction.rb
```diff
@@ -33,4 +33,56 @@ class BankTransaction < ActiveRecord::Base
   def image_url
     'data:image/png;base64,' + Base64.encode64(self.image)
   end
+
+  def assign_to_invoice
+    return false unless supplier
+
+    content = text
+    content += "\n" + reference if reference.present?
+    invoices = supplier.invoices.unpaid.select {|i| content.include? i.number}
```
wvengen: Is `include?` exact enough? What it there are invoices for number `12` and `123`? I might introduce a certain format for this.

config/schedule.rb
```diff
@@ -7,6 +7,11 @@
   rake "multicoops:run TASK=foodsoft:notify_users_of_weekly_task"
 end
 
+# Import and assign bank transactions
+every :weekday, :at => %w(5:56am 6:04pm) do
```
wvengen: Just to check: does this mean between these times or twice a day?

lib/tasks/foodsoft.rake
```diff
@@ -63,6 +63,19 @@ namespace :foodsoft do
...
+    if FoodsoftConfig[:use_bank_accounts]
+      for ba in BankAccount.all
+        import_method = ba.find_import_method
+        next unless import_method
+        import_count = import_method.call(ba)
+        assign_count = ba.assign_unchecked_transactions
+        rake_say "#{ba.name}: imported #{import_count}, assigned #{assign_count}"
+      end
+    end
+  end
```
wvengen: I wonder if there would be a way to emit a notice that `use_bank_accounts` is not enabled when running this task manually.

app/models/bank_account.rb
```diff
@@ -27,6 +27,10 @@ def import_from_file(file, options = {})
   end
 
   def find_import_method
+    # TODO: Move lookup for import function into plugin
+    return method(:import_from_easybank) if /^AT\d{2}14200\d{11}$/.match(iban)
+    return method(:import_from_sparkasse) if /^AT\d{2}20111\d{11}$/.match(iban)
+    return method(:import_from_holvi) if /^FI\d{2}799779\d{7}[0-9A-Z]$/.match(iban)
```
wvengen: This is really cool :) :)
I do think these specific implementations belongs in a plugin, though: no need to load them for all countries in all instances. One per continent or so? We could enable these plugins by default, but at least there is the option to get rid of them.
What do you think?

app/models/bank_account.rb
paroga: it's the same format we accept also at user and supplier. iban-tools also accepts ibans with spaces, witch is a problem when comparing them. IMHO the basic format is enough for now, so people could also insert "invalid IBANs" for countries without defined format

app/views/finance/bank_transactions/show.html.haml
paroga: i agree, but i don't want to start a new style for now. we could do this later too

app/models/bank_transaction.rb
paroga: NO. the format is designed that people can generate it by hand. Since the sum of the parts must match the amount of the transaction we have some kind of "checksum".

app/models/bank_account.rb
paroga: i see no need for transactions, since they are independent at this level. there are transactions for the financial_links

app/models/bank_transaction.rb
paroga: i agree that it's not 100% safe, but the usability must be high at this place, from my experience. if multiple users are responsible for doing the bank transfers it's hard that everybody does it with the exact same format.
since we have an additional check with the amount of the invoices it should be enough

config/schedule.rb
paroga: twice a day

lib/tasks/foodsoft.rake
paroga: for what? if there is no UI for seeing the transactions nobody would expect that something gets imported...

app/models/bank_account.rb
paroga: i would prefer one per country and everything enabled by default

paroga:
> Good steps! Documentation will be quite important, but that's for later, of course. For users, but also in the code itself at some places.

The user documentation will go (linke for the article upload) into the translation files.

> I do think it's also important to think of the impacts of integration Foodsoft with banks: what are the risks of accepting PINs etc. for bank login? I suggest we take this up on a Foodsoft mailing list.

For the final version the pin/passwords must be stored ENCRYPTED in the database at `bank_account`, but I'm still not 100% sure for the format and it still some time until I'm going to upstream that part. I just want to get the basic stuff into upstream for no, so I don't need to worry about database migrations anymore.

app/models/bank_transaction.rb
carchrae: if there is any chance an automated match could be wrong, you should probably not do it.  it can be fine as something in the ui to suggest relevant matches, but if there is any chance of error. 

as an aside, one 'hack' we do in our co-op is asking people to always put some extra cents on the end to make their transaction unique.  with e-transfers, etc, it isn't always clear where the money comes from when looking at the bank statement.

config/schedule.rb
carchrae: perhaps add `if FoodsoftConfig[:use_bank_accounts]` before scheduling this.

lib/tasks/foodsoft.rake
carchrae: from a logging/debugging point of view, it might be useful.   i'm not sure if `rake_say` or `Rails.logger` is the best way.

`rake_say "bank accounts are not enabled so skipping import_and_assign_bank_transactions"`

paroga:
in our setup we use only read-only credentials.

doing it automatically is a big benefit of this, since finance-people do not need to check the bank transactions every day and the members still see their money in financial transaction history the same day.

next step is to implement the parsing code first in a clean way with tests anyway, but i have to updated the PR according to the feedback

carchrae:
> in our setup we use only read-only credentials.
> 
> doing it automatically is a big benefit of this, since finance-people do not need to check the bank transactions every day and the members still see their money in financial transaction history the same day.

wow.  that is fantastic.  it is a tedious job to do otherwise.  unfortunately, our bank doesn't have any nice api - it isn't even on mint.com.  

i was just trying the bank account feature - but i can't use it because i don't have an IBAN number to put in.  :/  i will go and look at the logic so i can create a fake IBAN.  (or really, i should make a PR that allows other bank account numbers)

paroga:
our Austrian bank don't have any nice API too, but I implemented some screenscrapers with a few lines of code 😉 

i'd like to stick at the IBAN schema. you can just use a fake IBAN, since there is no checksum verification: e.g. CA00123456789

PR #353, opened 2015-02-18
Add documents

paroga:


wvengen:
Interesting idea!

I'm curious what others think: is this the right approach? Do we want to store documents in the database? Does something like this belong in the wiki, or is this ok?

wvengen:
I'm ok with the database, as a first simple step forward.

db/schema.rb
```diff
@@ -75,6 +75,15 @@
 
   add_index "deliveries", ["supplier_id"], name: "index_deliveries_on_supplier_id", using: :btree
 
+  create_table "documents", force: true do |t|
+    t.string "name"
+    t.binary "content", limit: 16.megabyte
```
wvengen: content-type?

paroga: I use the filename extension in this case.
IMHO storing the "unsafe" content-type provided by the browser is not much better than using the extension of the file.

wvengen:
It seems to me that some attachment code is bound to be duplicated in invoice attachments and here. Mime-type-checking, removing attachments, perhaps file-size error handling. I think we'd need to either our own concerns, or use some existing gem. Perhaps this [attachment_fu](https://github.com/pothoven/attachment_fu) is useful, it's one that is still being maintained (though we may need to add a full database storage type).

db/schema.rb
wvengen: True :+1:

wvengen:
Ok, testing it now and doing some small changes. I've done some changes and rebased on master (whoops - can't do a pull request anymore to you) - it's on https://github.com/wvengen/foodsoft/tree/feature/documents . You may consider starting from there.

One thing that I found very unintuitive is that mime-type is derived from the name. I entered a description there without extension, which meant the mimetype couldn't be guessed. So I think it works better when the mimetype is derived either from what the browser sends, or from the original filename. I guess we'd better save it.

paroga:
i'd prefer the browser content type solution. should I do this change?
one other point in your changes: adding the translations to the en file in the plugins directory had no effect at my installation. only the en file in the core worked. the de translations worked correctly in the plugin folder. Is this a problem with my installation or a general problem?

wvengen:
I'm for the browser content-type solution as well. Would be great, yes.
Hmmm I could see the changes in the plugin locales right away, strange.
(The idea is to put the English locales in the plugin, and then to import them into main Foodsoft as well so that they can easily be translated using localeapp.)

wvengen:
fyi: for the Travis CI build to succeed, you need to bundle install and commit the Gemfile.lock changes.

Tamriel:
What has to be done to complete this issue?

wvengen:
Thanks for asking, I see the following issues:
* Security - see OWASP on [Unrestricted File Upload](https://www.owasp.org/index.php/Unrestricted_File_Upload)
  * Never accept a filename and its extension directly without having a white-list filter. 
  * All the control characters and Unicode ones should be removed from the filenames and their extensions without any exception
* Rebase on master

For the rest, I'd say - all set.
Would it be ok to not enable this plugin by default just yet? I'd like to have more clarity first if sysadmins agree on putting files in the database. You may ask it on the -dev or -discuss mailing list yourself, if you like.

paroga:
ping

plugins/documents/app/controllers/documents_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def create
+    @document = Document.new
+    @document.data = params[:document][:data].read
+    @document.mime = FileMagic.new(FileMagic::MAGIC_MIME).buffer(@document.data)
+    if params[:document][:name] == ''
+      name = params[:document][:data].original_filename
+      name = File.basename(name)
+      @document.name = name.gsub(/[^\w\.\-]/, '_')
```
wvengen: Great you added a check. Still it's easy to bypass when supplying a name parameter directly.
What about doing this substitution on serving the file? That way we can even change what characters to allow afterwards.

plugins/documents/app/controllers/documents_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  rescue => error
+    redirect_to documents_path, alert: t('documents.destroy.error', error: error.message)
+  end
+
+  def show
+    @document = Document.find(params[:id])
+    filename = @document.name
+    unless filename.include? '.'
+      filename += '.' + MIME::Types[@document.mime].first.preferred_extension
+    end
```
wvengen: It would still be possible to trick (older) browsers or operating systems that rely on file extension instead of mime-type to open dangerous files. I'd prefer to always add or replace the detected extension.

plugins/documents/app/controllers/documents_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def new
+    @document = Document.new
+  end
+
+  def create
+    @document = Document.new
+    @document.data = params[:document][:data].read
+    @document.mime = FileMagic.new(FileMagic::MAGIC_MIME).buffer(@document.data)
```
wvengen: What do you think of including mimetype whitelisting here as well? This one can be a lot broader, but I'd be careful with executable files, for example. (Think of someone uploading an html file with javascript.)

plugins/documents/app/controllers/documents_controller.rb
paroga: i do not want to store the original_filename in the DB, so storing a more or less valid name (if not provided by the user already) seams more reasonable to me. This is not related to security. It's just for providing nice filenames.

plugins/documents/app/controllers/documents_controller.rb
paroga: Yes, but e.g. a DOC file can contain malicious code too, but it's still a valid filetype I want to be able to store. This has nothing to do with security, it's only a feature to get nice filenames (if not provided by the uploading user)

plugins/documents/app/controllers/documents_controller.rb
wvengen: Ah, I see.
I'd really like to follow OWASP and filter filename characters to known safe ones.

plugins/documents/app/controllers/documents_controller.rb
paroga: i do not think that whitelisting will have any value. As long as we do not include a "malicious code checker" (which does not exist), more or less every file (HTML, PDF, DOC) can contain bad code. Then the whitelist contains only TXT, which makes the whole documents useless again. IMHO since documents (a) are an optional feature and (b) can only be accessed by a authenticated user, this is not such a big problem. IMHO the important part is, that an attacker can not get any information from the server. When/if an other user will download a mad file can not be avoided to 100%.

plugins/documents/app/controllers/documents_controller.rb
wvengen: Ah! Indeed known filetypes we'd like to accept may contain malicious code. Still I'd like to help users to avoid saving a .exe file to their desktop, and later opening it.

plugins/documents/app/controllers/documents_controller.rb
paroga: blacklisting as bad practice and whitelisting all unicode characters gets ugly very quickly. What's the problem with storing mad filenames? The only place where it could be problematic is when the browser reads the filename from the HTTP header and when the user uses a browser which is not able to to the without security impact (s)he might have even bigger problems than that!

plugins/documents/app/controllers/documents_controller.rb
paroga: then they should deactivate documents (and the foodcoop has real problems with their members). IMHO this only creates a wrong impression of security and a exe might be a valid document.

wvengen: Things like this happen without the intention to do harm by members using the system. While you and me understand the difference, many don't have a clue. Since OWASP recommends basic file upload security practices, and many existing systems do it (like WordPress), it's the least we can do, isn't it?

paroga: please don't get me wrong, i'm not against security. I just want to know what kind of security we want to provide. OWASP list mainly server-side-issues, which do not apply to implementation since we just pass the file trough. If you say that accidentally uploaded EXE files is something we want to address I'm ok with implementing some kind of whitelisting.

wvengen: Ah, now I understand our different approaches :) And I think you have a point with OWASP being mostly server-side.
Indeed I was thinking more about user-issues than server-security.

I'd be happy to include this as a plugin not loaded by default as-is, with a warning in its README (next to the one about putting files in the database). If it's ever going to be loaded by default, then I'd like filename sanatization and a whitelist to be in place.

Thanks for your patience with me :)

paroga: adding it without enabling seams like a good idea to me. we can improve it the in the master branch (to get it enabled by default).

wvengen:
Great! Disabled it by default, updated the plugin's README and merged.
Thanks a lot!

PR #354, opened 2015-02-18
Show recipients of messages

paroga:


wvengen:
Great! Solves #42 :) I'll test and merge.

PR #355, opened 2015-02-25
Add expected delivery date

paroga:


wvengen:
Ah! Would you care to explain a bit behind the need for this? I like addition of relevant dates, but I'm not sure yet where these could be placed best.

(In the foodcoop-adam fork there's a similar "pickup date" field, more as a sign to people at/from which date they can pickup their goods. This is also shown on the fax pdf as a "deliver before" date.)

paroga:
sounds like the pickup date is the same.

the additional page is for people who check the delivery: the people rotate in our food coop and with the simple list of orders it was not clear, which PDF they should print out (some of our orders hang in the closed state for over a month since we get only a invoice for them once per month)

wvengen:
It seems we're on the same page. For the foodcoop-adam fork, I experimented with a current_orders plugin, which lists all finished_but_not_closed orders, and also allows to modify articles and ordergroups for those orders. I've described it a bit in #358. Is that a direction that would be suitable for you?

paroga:
imho your solution is too complex for us. we need only a list, which list the "group PDF" and "fax PDF" grouped by the pickup date. having additional option might confuse people who do not know the whole system in detail

wvengen:
So you only need to download the PDFs, right? And you don't use the receive functionality?

paroga:
yes, exactly

wvengen:
I'm wondering how Foodsoft can be suitable for your situation, as well as ours. Something to think about.

For your current case: would a download button (menu) in the order overview screen also work? There you already have a list of the currently relevant orders.

paroga:
there are too many closed orders. we get a delivery every week, but only an invoice per month. we had problems in the last weeks that people did not find all relevant orders.
so, unfortunately: no
also the download should be possible for every user, but creating and changing of orders only for the people with the order role

paroga:
maybe we can only add the pickup date to the database and the usual order pages for now? and think about the other points in a next step?

wvengen:
Ah! I've been thinking of introducing something like a "pickup day" context. Each order then belongs to a specific pickup day, so that Foodsoft knows which orders belong together. Some screens could then group by pickup day. Like the ordering, order management and finance screen.

Perhaps we need a pickup-day/distribution "role", so that members in this group can view a screen like you suggested in this pull request. And these could also do the other stuff in #358, if so desired.

wvengen:
Agree to add pickup date now, and leave the rest for later. Thanks for keeping the scope managable :)

config/locales/en.yml
```diff
@@ -1332,7 +1333,11 @@ en:
         Do you really want to close the order %{order}?
         There is no going back.
       create_invoice: Add invoice
-      description1: "%{state} order from %{supplier} created by %{who}, open from %{starts} until %{ends}."
+      description1_order: "%{state} order from %{supplier} created by %{who}, "
+      description1_period:
+        starts: "open from %{starts}"
+        starts_ends: "open from %{starts} until %{ends}"
+        pickup: " and can be picked up on %{pickup}"
```
wvengen: let's put the leading space in the code - can be hard for translators to see at all

config/locales/en.yml
```diff
@@ -1332,7 +1333,11 @@ en:
         Do you really want to close the order %{order}?
         There is no going back.
       create_invoice: Add invoice
-      description1: "%{state} order from %{supplier} created by %{who}, open from %{starts} until %{ends}."
+      description1_order: "%{state} order from %{supplier} created by %{who}, "
```
wvengen: same here about the space

paroga:
ping

wvengen:
Cool, I'll have a look.

app/views/orders/_form.html.haml
```diff
@@ -4,6 +4,7 @@
     = f.input :starts, as: :date_picker_time
     = f.input :boxfill, as: :date_picker_time if @order.is_boxfill_useful?
     = f.input :ends, as: :date_picker_time
+    = f.input :pickup, as: :date_picker
```
wvengen: class `input-small` would be nice here to be in-line with the other date inputs

wvengen:
Looking good. (Too bad `format_date` gives something else than the date in `format_time` - that was already the case I'm afraid.) One little thing, otherwise, mergeable!

wvengen:
Thanks! :)

PR #356, opened 2015-02-25
Add reply_to and group_id to messages schema

paroga:


wvengen:
Nice addition.

It may be useful to learn about [Rails migrations](http://edgeguides.rubyonrails.org/active_record_migrations.html). Generating a migration for this is as easy as

```sh
rails g migration AddReplytoAndGroupidToMessages reply_to:integer group_id:integer
```

If you have a database that is equal to the original state (before the migration), the generated db/schema.rb will be correct after running `rake db:migrate`. If you have a different database structure than foodsoft master, schema.rb can be regenerated using

```sh
git checkout foodcoops/master
RAILS_ENV=test rake db:drop db:create db:schema:load
git checkout my/branch
RAILS_ENV=test rake db:migrate
```

and you'll have a proper db/schema.rb.

I'd be happy to do this as part of merging the pull request. But then you know if you need it.

wvengen:
Merged after fixing migrations. Thanks!

paroga:
thx for the info. i know already about the rails migration, but i skipped the whole migration generation in all my push request, since i generated the changes with a different workflow.
would be great if you could it it during the merge

PR #357, opened 2015-02-26
Add last user login and activity

paroga:


wvengen:
Nice idea!

wvengen:
I think the admin user list is getting a bit wide. Shall we just show the last activity column? We could show both in the user details screen.

paroga:
as you like.... just say

wvengen:
yes please :)

PR #363, opened 2015-04-23
Run on containers, and make use of cache

BanzaiMan:


wvengen:
Wow, this is the best reply I could wish for, thank you!

PR #364, opened 2015-04-26
Docker support

bennibu:
Looking into some issues required running the foodsoft again on my laptop. And as we use at aboutsource Docker for development purposes, I added docker support for the foodsoft.

I have not touched the general README for developers but added a README_DOCKER instead. See this file for further explanations.

Docker makes dealing with the multiple dependencies (mysql, redis, mailcatcher) very easy. Please try it out and give me your feedback.

Warning, I have tested it only on a up to date linux machine. There are maybe some issues on MacOS with the mysql volume.

As the docker containers are not having a real X system I added virtual X support. See file `bin/test`.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,12 +1,17 @@
 #!/bin/sh
 
-# Set test variables
-export DATABASE_URL=mysql2://root:secret@mysql/test
+# We use TEST_DATABASE_URL to make sure we don't accidentaly overwrite dev/prod db
+unset DATABASE_URL; export DATABASE_URL
+[ "$TEST_DATABASE_URL" ] && export DATABASE_URL="$TEST_DATABASE_URL"
 export RAILS_ENV=test
 
 # Start virtuals X environment to allow integration testing via firefox/iceweasel
 export DISPLAY=:99
-Xvfb :99 &
+Xvfb $DISPLAY -nolisten tcp &
+XVFB_PID=$!
 
 # Start tests
 rspec
+
+# Cleanup
+kill $XVFB_PID
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 .idea
 .sass-cache
 log
+config/database.yml
 public/assets
 public/system
 tmp/*
--- a/doc/SETUP_DEVELOPMENT.md
+++ b/doc/SETUP_DEVELOPMENT.md
@@ -5,6 +5,10 @@ Gratulations, if you read this file locally, you have successfully cloned the
 foodsoft project from the git repository. Now you are only a few steps away
 from trying it out and then jumping into development.
 
+This document describes how to setup Foodsoft for development on your local system.
+Alternatively, you [run Foodsoft using Docker](SETUP_DOCKER.md).
+
+
 **System requirements**:
 [RVM](https://rvm.io/rvm/install),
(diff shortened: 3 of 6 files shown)


wvengen:
Looks great, thanks! I'm testing it out, will merge if that works.
I'd like to move the new Docker readme to doc/, if that's ok.

Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM ruby:2.1-slim
```
wvengen: Any strong reason why you're using the slim variant (the [Ruby Docker page](https://registry.hub.docker.com/_/ruby/) seems to discourage this)? I can even imagine that some `apt-get install`-ed packages are already already present when using the normal image, speeding up the process and using less space when many images are present. Haven't measured it, though.

README_DOCKER.md
```diff
@@ -0,0 +1,68 @@
...
+
+    mkdir -p ~/.docker-volumes/foodsoft/mysql
+
+Setup foodsoft development data: (This will take some time, containers needs
+to be pulled from docker registry and a lot dependencies needs to be installed.)
+
+    docker-compose run app rake foodsoft:setup_development
+
+TODO: Right know this is not gonna work because of the new database
+configuration via ENV variable. See Notes below.
```
wvengen: What about skipping the database setup step when DATABASE_URL is defined?
(And perhaps the Dockerfile might even remove it before starting, to make local a development install work with Docker as well.)

bin/test
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# Set test variables
+export DATABASE_URL=mysql2://root:secret@mysql/test
```
wvengen:
```bash
if [ "$DATABASE_URL" ]; then
  if echo "$DATABASE_URL" | grep -q dev; then
    export DATABASE_URL=`echo "$DATABASE_URL" | sed 's/dev\(elopment\)\?/test/g'`
  else
    export DATABASE_URL="${DATABASE_URL}_test"
  fi
fi
```

wvengen:
Please see https://github.com/aboutsource/foodsoft/pull/1 for some improvements.

config/environments/development.rb.SAMPLE
```diff
@@ -39,7 +39,7 @@ Foodsoft::Application.configure do
   # Mailcatcher config, start mailcatcher from console with 'mailcatcher'
   # Mailcatcher can be installed by gem install mailcatcher
   config.action_mailer.delivery_method = :smtp
-  config.action_mailer.smtp_settings = { address: "localhost", port: 1025 }
+  config.action_mailer.smtp_settings = { address: "mailcatcher", port: 1025 }
```
wvengen: Quickly looking over the changes again, should we get this from an environment variable with a `localhost` default, so that it works both with and without Docker?

bennibu: Yep, thats a good idea. We also could do this in application.rb for all
environments at once, so we don't have to maintain multiple action
mailer configurations.

wvengen: I like moving system-dependent config to environment variables. But since the defaults for each rails env are different, I would think this still needs to go into the environment's settings, right?

PR #365, opened 2015-05-03
Add color marking to items in group orders that have missing items

akirk:
In our foodcoop the request occurred to make it more obvious when items just need a few more items, this patch does this by adding background-colors to the `tr`s.

I am a RoR newbie, so feel free to improve this patch or ask for improvements before merging.

more than 1 item missing:
missing-many

1 item missing:
missing-few

no items missing (just used through JavaScript when the count reaches 0)
missing-none

### DIFF
--- a/app/assets/javascripts/ordering.js
+++ b/app/assets/javascripts/ordering.js
@@ -118,11 +118,25 @@ function update(item, quantity, tolerance) {
     $('#price_' + item + '_display').html(I18n.l("currency", itemTotal[item]));
 
     // update missing units
-    var missing_units = unit[item] - (((quantityOthers[item] + Number(quantity)) % unit[item]) + Number(tolerance) + toleranceOthers[item])
-    if (missing_units < 0 || missing_units == unit[item]) {
+    var missing_units = unit[item] - (((quantityOthers[item] + Number(quantity)) % unit[item]) + Number(tolerance) + toleranceOthers[item]),
+        missing_units_css = '';
+    if (missing_units < 0) {
         missing_units = 0;
+        missing_units_css = '';
+    } else if (missing_units == unit[item]) {
+        missing_units = 0;
+        missing_units_css = 'missing-none';
+    } else if (missing_units == 1) {
+        missing_units_css = 'missing-few';
+    } else {
+        missing_units_css = 'missing-many';
     }
-    $('#missing_units_' + item).html(String(missing_units));
+    $('#missing_units_' + item)
+        .html(String(missing_units))
+        .closest('tr')
+        .removeClass('missing-many missing-few missing-none')
+        .addClass(missing_units_css);
+
 
     // update balance
     updateBalance();
--- a/app/assets/javascripts/ordering.js
+++ b/app/assets/javascripts/ordering.js
@@ -118,14 +118,16 @@ function update(item, quantity, tolerance) {
     $('#price_' + item + '_display').html(I18n.l("currency", itemTotal[item]));
 
     // update missing units
-    var missing_units = unit[item] - (((quantityOthers[item] + Number(quantity)) % unit[item]) + Number(tolerance) + toleranceOthers[item]),
+    var missing_units = calcMissingItems(unit[item], quantityOthers[item] + Number(quantity), toleranceOthers[item] + Number(tolerance)),
         missing_units_css = '';
-    if (missing_units < 0) {
+
+    if (missing_units <= 0 || missing_units == unit[item]) {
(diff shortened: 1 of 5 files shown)


wvengen:
Thanks for your contribution!

I'm not using this member ordering screen myself (#), could someone else please comment here if it would make sense for others as well?

There are some questions that I have when seeing the interface:
* What do you think should happen when I've ordered in the past, I've secured my articles (my numbers are in green, not red), but there are still missing units for the foodcoop as a whole?
* What do you think should happen when I have secured a portion of my amount (both green and red numbers are >0)?
* Would it be useful to merge this with the background-coloring of the order result screen?
* There may be some overlap with:
  * the order result screen  - `GroupOrdersHelper#order_article_cass_name`
  * the orders screen - `OrdersHelper#order_article_class` (though that's about an `OrderArticle` instead of a `GroupOrderArticle`) (*)

(*) I realise that something is a bit convoluted here (!)

akirk:
The main reason behind this change is to make rows that contain missing items more visible.

Oftentimes people ordering later would join in on something where there is just one or a few items missing but they miss the need for just one more order because the number in the Missing column does not stand out visually. 

So think of this patch as introducing a visual representation of the question: We need more orders on this to get it fulfilled, do you want to join?

Regarding your questions:
> What do you think should happen when I've ordered in the past, I've secured my articles (my numbers are in green, not red), but there are still missing units for the foodcoop as a whole?

The behavior should be the same as implemented. Maybe I am willing to order even one more item to help the order for others be fulfilled. The background is meant to draw my attention to the line. If I see that my order will be fulfilled, I know that I don't have to do something.

I do understand that people might get this wrong and think that the order might not be fulfilled because it has a red background. The "Amount" column stays the same as now, so people could see that they don't have to act.

Still it would probably make sense to add another indicator or text, to explain that I will get my own order and don't have to order more. Maybe we should try this change and then get feedback from people if they were indeed confused or not?

> What do you think should happen when I have secured a portion of my amount (both green and red numbers are >0)?

Also as it is implemented. The securing of my own order is not relevant, we are trying to ask people to order more to get the current batch complete.

> Would it be useful to merge this with the background-coloring of the order result screen?
> There may be some overlap with:
> the order result screen - GroupOrdersHelper#order_article_cass_name
> the orders screen - OrdersHelper#order_article_class (though that's about an OrderArticle instead of a GroupOrderArticle) (*)

Possible, as I said I am new to RoR so I am not sure where to put it. Maybe a CssClassHelper would make sense, too?

wvengen:
Thanks for clarifying that the main purpose is to get the packages full as a group. Great idea!

I'm seeing something unexpected when using tolerance:
* the background color becomes red when missing units == 0 and tolerance is used by the member
* when surplus tolerance is given and there is secured quantity > 0, the background color isn't green anymore
* when quantity and tolerance are set to zero, background color is green (but should be white in my opinion)
* when quantity is zero and tolerance > 0, background color is also changing. I guess this is a bug in the missing units computation.

And what do you think about using the same color green as in the results screen row highlight (`success` class used in `GroupOrdersHelper#order_article_class_name`), which is equal to the flash-success color?

akirk:
Thanks for your input, (I think) I have fixed the missing units algorithm which fixes your points. Also I have changed the background color. Ready to merge now? :)

wvengen:
Thanks, I'll have a look!

wvengen:
Feels great!! This is much better, thanks.

There's one tiny thing I encountered: when ordering an amount and missing units is zero, the background turns green. So far, so good. Now I save the order, and go back to edit order. Now the background is white. When I press + and - it turns green again. Not a blocker, perhaps just a little confusing. Or did you do this on purpose?

wvengen:
And the code great, especially for a first Rails contribution.

akirk:
Thanks :)

The white/green behavior is somewhat on purpose. Colors are meant to provoke action. When there is nothing missing initially it's white. So now you change something, so an action is needed to fulfil the current batch. Now that you change the value again, the "goal" is achieved, so it turns green again.

That the new value is actually the original state is not checked, but I'd consider that an over-optimization, and rather something that you would encounter while playing around with the + and - buttons while testing. If you insist it should be easy enough to fix, I suppose.

wvengen:
Ah, that actually makes a lot of sense. Thanks for your explanation, and putting care into user experience!

Tamriel:
Right now, when items are missing, the number is colored in red.
This is hardly readable, I would like it more if the whole row would be colored (like in the mockup in the first post).

PR #366, opened 2015-05-15
Fixed binding of rails server in Dockerfile and compose config.

bennibu:
As with rails 4.2 the internal server now binds to localhost as default. So we have to overwrite this behaviour to access the app within the docker container.

By the way, thanks for the new release. I updated the foodcoops.net hosting.


wvengen:
Thanks. And also for updating the hosting!

PR #367, opened 2015-06-03
Add cost_type to financial_transactions

paroga:
This is the minimal implementation we need to model our handling of money in the foodsoft.

What would be an implementation which will get accepted in the official version?

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -225,7 +225,7 @@ class Order < ActiveRecord::Base
     transaction do                                        # Start updating account balances
       for group_order in gos
         price = group_order.price * -1                    # decrease! account balance
-        group_order.ordergroup.add_financial_transaction!(price, transaction_note, user)
+        group_order.ordergroup.add_financial_transaction!(:credit, price, transaction_note, user)
       end
 
       if stockit?                                         # Decreases the quantity of stock_articles
@@ -287,4 +287,3 @@ class Order < ActiveRecord::Base
   end
 
 end
-
--- a/app/views/home/index.html.haml
+++ b/app/views/home/index.html.haml
@@ -53,14 +53,24 @@
           %th= heading_helper FinancialTransaction, :created_on
           %th= heading_helper FinancialTransaction, :user
           %th= heading_helper FinancialTransaction, :note
-          %th= heading_helper FinancialTransaction, :amount
+          %th= heading_helper FinancialTransaction, :deposit
+          %th= heading_helper FinancialTransaction, :memebership_fee
+          %th= heading_helper FinancialTransaction, :credit
         - for ft in current_user.ordergroup.financial_transactions.limit(5).order('created_on DESC')
           %tr
             %td= format_time(ft.created_on)
             %td= h(show_user(ft.user))
             %td= h(ft.note)
             - color = ft.amount < 0 ? 'red' : 'black'
-            %td{:style => "color:#{color}; width:5em", :class => "currency"}= number_to_currency(ft.amount)
+            %td.currency{:style => "color:#{color}; width:5em"}
+              - if ft.deposit?
+                = number_to_currency(ft.amount)
+            %td.currency{:style => "color:#{color}; width:5em"}
+              - if ft.membership_fee?
+                = number_to_currency(ft.amount)
+            %td.currency{:style => "color:#{color}; width:5em"}
+              - if ft.credit?
+                = number_to_currency(ft.amount)
       %br/
       %p= link_to t('.my_ordergroup.transactions.view'), my_ordergroup_path
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -77,6 +77,7 @@ ActiveRecord::Schema.define(version: 20140921104907) do
 
   create_table "financial_transactions", force: true do |t|
     t.integer  "ordergroup_id",                         default: 0, null: false
+    t.integer  "cost_type,                              default: 0, null: false
     t.decimal  "amount",        precision: 8, scale: 2, default: 0, null: false
     t.text     "note",                                              null: false
     t.integer  "user_id",                               default: 0, null: false
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ gem 'prawn-table'
 gem 'haml-rails'
 gem 'kaminari'
 gem 'simple_form'
+gem 'enum_help'
 gem 'inherited_resources', git: 'git://github.com/josevalim/inherited_resources.git', branch: 'rails-4-2'
 gem 'localize_input', git: "git://github.com/bennibu/localize_input.git"
 gem 'daemons'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -154,6 +154,7 @@ GEM
     diff-lcs (1.2.5)
     diffy (3.0.7)
     docile (1.1.5)
+    enum_help (0.0.14)
     erubis (2.7.0)
     eventmachine (1.0.4)
     exception_notification (4.0.1)
@@ -473,6 +474,7 @@ DEPENDENCIES
   daemons
   database_cleaner
   date_time_attribute
+  enum_help
   exception_notification
   factory_girl_rails
   faker
--- a/app/controllers/finance/financial_transactions_controller.rb
+++ b/app/controllers/finance/financial_transactions_controller.rb
@@ -58,7 +58,7 @@ class Finance::FinancialTransactionsController < ApplicationController
     params[:financial_transactions].each do |trans|
       # ignore empty amount fields ...
       unless trans[:amount].blank?
-        Ordergroup.find(trans[:ordergroup_id]).add_financial_transaction!(trans[:amount], params[:note], @current_user)
+        Ordergroup.find(trans[:ordergroup_id]).add_financial_transaction!(params[:cost_type].to_i, trans[:amount], params[:note], @current_user)
       end
     end
     redirect_to finance_ordergroups_url, notice: I18n.t('finance.financial_transactions.controller.create_collection.notice')
--- a/app/models/financial_transaction.rb
+++ b/app/models/financial_transaction.rb
@@ -10,8 +10,10 @@ class FinancialTransaction < ActiveRecord::Base
 
   localize_input_of :amount
 
+  enum cost_type: [:credit , :membership_fee, :deposit]
+
   # Use this save method instead of simple save and after callback
   def add_transaction!
-    ordergroup.add_financial_transaction! amount, note, user
+    ordergroup.add_financial_transaction! cost_type, amount, note, user
   end
 end
--- a/app/models/ordergroup.rb
+++ b/app/models/ordergroup.rb
@@ -47,11 +47,11 @@ class Ordergroup < Group
 
   # Creates a new FinancialTransaction for this Ordergroup and updates the account_balance accordingly.
   # Throws an exception if it fails.
-  def add_financial_transaction!(amount, note, user)
-    transaction do      
-      t = FinancialTransaction.new(:ordergroup => self, :amount => amount, :note => note, :user => user)
+  def add_financial_transaction!(cost_type, amount, note, user)
+    transaction do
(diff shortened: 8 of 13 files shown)


wvengen:
Looks fine at first sight! I guess this would be fine. I'll test it later.

paroga:
Did you find some time to test it in the meantime?

app/models/order.rb
```diff
@@ -225,7 +225,7 @@ def close!(user)
     transaction do                                        # Start updating account balances
       for group_order in gos
         price = group_order.price * -1                    # decrease! account balance
-        group_order.ordergroup.add_financial_transaction!(price, transaction_note, user)
+        group_order.ordergroup.add_financial_transaction!(:credit, price, transaction_note, user)
```
wvengen: I'd expect something like `order` here. Why `credit`?

db/schema.rb
```diff
@@ -77,6 +77,7 @@
 
   create_table "financial_transactions", force: true do |t|
     t.integer  "ordergroup_id",                         default: 0, null: false
+    t.integer  "cost_type,                              default: 0, null: false
```
wvengen: Hey, you're missing a [migration](http://edgeguides.rubyonrails.org/active_record_migrations.html) here :o

app/views/home/index.html.haml
```diff
@@ -53,14 +53,24 @@
...
-            %td{:style => "color:#{color}; width:5em", :class => "currency"}= number_to_currency(ft.amount)
+            %td.currency{:style => "color:#{color}; width:5em"}
+              - if ft.deposit?
+                = number_to_currency(ft.amount)
+            %td.currency{:style => "color:#{color}; width:5em"}
+              - if ft.membership_fee?
+                = number_to_currency(ft.amount)
+            %td.currency{:style => "color:#{color}; width:5em"}
+              - if ft.credit?
+                = number_to_currency(ft.amount)
```
wvengen: Do you really want to have a column for each transaction type? Perhaps using icons is somewhat more flexible (though I did have a hard time coming up with suitable ones). Any other ideas?
And I suppose this could be done cleaner, using css classes, for example. I know the original code isn't too great, but this makes it worse.

wvengen:
Thanks for your contribution, it's a useful feature!

Would you be able to tell what you'd like to improve in the process of a foodcoop? That helps me to reason about an implementation. What totals do you need (and why, if you know), for example.

db/schema.rb
paroga: PR is just for early feedback :-)

app/models/order.rb
paroga: because it uses the "credit-subaccount" of the ordergroup

app/views/home/index.html.haml
paroga: it's not really a transaction type - it's more a "subaccount" of the ordergroup: we have threee different "kind of money" members can transfer to the bankaccount of the foodcoop: (1) a deposit, which they get back when they leave the foodcoop, (2) membership fee, which is decreased every month and (3) the credit for "buying" articles

paroga:
i'd like to but everything of our financial stuff into the foodsoft: what's missing ATM is handling for (monthly) membership fees and deposits.
In a later step it would be cool if the foodsoft could show a warning message (or reject orders) if the membership fee has not been paid for the actual month.
So I need a total of the membership fee and the current "normal" credit total.

app/views/home/index.html.haml
wvengen: Hmm, I understand. And wonder if this can be simplified. The membership fee could be a debit of the account, either by a cronjob, or when placing an order.
The deposit is really different. I'd consider this a candidate for a plugin. One approach would be to debit the ordergroup's account with the deposit upon creation. Then when the member leaves, create a credit transaction of the deposit (which you can find by transaction description), and then pay back the remaining funds and create a financial transaction for that; then delete the ordergroup (or only remove its users). In this case you keep a full record, and don't need different "subaccounts". (One more improvement to introduce an ordergroup status in Foodsoft, so that members can become inactive, and the plugin could credit the deposit on their account on that transition.)
What would you think of this approach?

wvengen:
Thanks for explaining. I know another foodcoop using Foodsoft having a yearly membership fee.

About the periodic membership fee, I see two options:
- Cronjob that debits a member when the last membership fee is longer ago than the period (month). Downside is that sleeping members are also charged, and when they get back after two years, find that they have an enormous negative balance, which isn't a nice welcome.
- When a member creates a new group_order, Foodsoft checks if the ordergroup's membership payment is due, and if so, debits the membership fee (either then, or when the order is settled). This needs a warning in the ordering screen that membership payment is due.

I still think it would be useful to introduce transaction types, but merely as tags (not as subaccounts). Membership would be one type, and ordering would be denied if membership fee payment is overdue.

What do you think?

app/views/home/index.html.haml
paroga: Since we do not handle the "normal credit" via the Foodsoft we need different subaccounts.
Maybe more information how we handle money in our foodcoop:
Members give money via SEPA bank transfer to the bank account of the foodcoop. The "Zahlungsreferenz" field (usually used by companies to have a uniqued id for findig the corresponding invoice) must be filled for every transaction with a special code. This code then tells how much money of should be used for the different "subaccounts". This is required to keep the transaction count low to keep the banking fees at a minimum.
Our foodsoft then imports all information directly from our ebanking and adds the different amounts to the different subaccounts.
In our stock there is a folder with a sheet of paper for every ordergroup, which keeps track of "the real account balance". After transferring money to the bank account of the foodcoop a member has to add this transaction also to this sheet of paper (but only the amount for the "credit"-type, the "membership"-cost are handled only in the foodsoft).
When buying stuff the members have to add this their sheet of paper and calculate the new total, so that there is always the current balance of their account visible.
As you can see the amount on the sheet of paper and the amount shown in the foodsoft get out of sync. Because of that we take the latest balance of the paper and insert it into the foodsoft every few months.
The "balancing"-feature for the orders in the foodsoft is not used.

Because of that offline/online-difference and to make the membership fee more transparent to the members we need this distinction.
Sometimes we have to transfer money between the different subaccounts, so splitting this information into various plugins is not an option.

FYI: We are not the only foodcoop with the "paper in the stock"-concept. So maybe find a nice solution for that, which can be accepted in upstream.

paroga:
The membership fee has to be payed every month, not only when you place orders (the rent of the stock is also independent of the placed orders ;-)). We do not want to have "inactive members", since we have a long list of potential new members just waiting to get active (we limited the maximum count of members in our foodcoop). A special working group is responsible to get in contact with "inactive members" and asks them if they maybe want to leave, so a new "active member" can take over.

Tags for transactions are nice too, but what we need is a strong distinction between membership fee and "normal credit" (but with the possibility to transfer money between them for admins).

wvengen:
Makes sense :) (though I also know groups who would want this differently).
I would really like to avoid the need for different kinds of credits, as it adds complexity.
Would a cronjob work for you for the membership fee?

paroga:
I have a cronjob already, which inserts the membership fees every month. That's not the problem. ;-)
But what I need is to to create a sum of the membership fees independent of the "normal credit" and being able to show a warning (or decline orders) in a later step. Do you have a better idea how we could handle that without mixing "offline" with "online" transactions, so it gets accepted upstream?

wvengen:
A, I see. Good job :) And thanks for explaining what you're doing in more detail, the offline paper thing sounds quite a hassle.

To deny ordering with unpaid membership, we somehow need to know if a certain financial transaction is paid or not. 

The subaccount idea would be a way to solve this. Clever thinking. Still it adds complexity in code and in user-interface, which I'd like to avoid when possible.

What about this: we could add a `paid` column to (debit) transactions. On every new transaction, the `paid` status of all unpaid debit transactions are recomputed.
When ordering, Foodsoft would look at the last membership fee transaction (identified by transaction-type in the "tag"-sense) and deny ordering if that hasn't been paid yet.

(Let's try to solve the membership thing first, and when that's clear look at deposit.)

paroga:
I'm also not a fan of this offline paper thing, but since not everybody has a smartphone, we do not want to put a PC in the stock and we also need a possibility to buy articles which are only in stock (and not in foodsoft), it's the best solution for now.

I also do not like the additional complexity, but it was the easiest solution I found (and it took my quite a while to figure it out :-D). Unfortunately the problem is quite complex, but if we find a better solution I'm more than happy.

How does a `paid`column help me? Member need a possibly to say how much money they want to use "offline" and how much "online". A person with the "financial" right must know how much she can use of membership fees (=online). Otherwise the membership fee must be placed on the offline paper too.
Maybe we can say that we have only two categories "online" and "offline" instead of three if that makes it less complex?
There a split in the responsibility: (1) For the "online money": The financial group (or the foodsoft in a later step) must ensure that the the total is positive, and if not "lock" the account. (2) For the "offline money" every ordergroup must keep a positive balance and transfer money to the foodcoops bank account if the under a specific threshold.

I agree with solving the membership thing first.

wvengen:
Yeah, in several foodcoops I know there are also other lists and systems to complement the features. Having the financial records somewhere else than in Foodsoft, and Foodsoft knowing only approximate financial info easily gets a mess, in my experience. But your setup sounds workable, congratulations.

So now I hear a new problem: members want to buy things that are not in Foodsoft. The way for this would be to create a financial transaction (debit) for the member in Foodsoft. I've heard of groups having a list on the table where everyone writes down how much he buys from stock, at the end of the month someone enters the list in Foodsoft (there's a 'enter multiple transactions at once' screen for that).

Back to the membership thing. If the only thing we're solving now is that people need to pay a monthly membership fee (your cronjob) and that a member should be unable to order when the fee hasn't been paid, it could work like this.
1. Cronjob creates a new transaction. Member gets a negative balance.
2. Member wants to order. He cannot, because the membership fee is unpaid (the `paid` column of the last membership fee transaction is false).
3. Member does a SEPA transaction. This is picked up by the bank import script, and a new financial transaction is created in Foodsoft with the amount. If the amount is enough, the membership fee financial transaction is set to paid.
4. Members tries to order, and succeeds.

`FinancialTransaction` would need a new `transaction_type` column for finding the last membership-fee-transaction, and a `paid` column to figure out if it was paid or not.

That would prohibit people from ordering when their membership fee hasn't been paid, right?

paroga:
Buying things not in Foodsoft is not a problem: When picking up an order in the stock a member has to write the cost of that order into her offline paper. If she does add some additional amount makes no difference. We do not track for which articles money has been spent. Placing orders and paying them (by writing the amount to the offline paper) is completely decoupled.

Yes, the `paid` field would help in implementing the "locking orders when no membership fee is paid"-feature, but it does not with decoupling "offline" and "online" money.

If somebody can order without paying the membership fee is not really important to us, what counts is the have clear financial records. Having the offline paper is not the best solution, but I have to deal with it and I'd like to avoid having a second system only for financial records (which gets a mess sooner or later).

wvengen:
Ok, let's shift the topic from handling membership fee to "online" vs. "offline" money.
I suppose the goal is: getting all financial records in Foodsoft, right?

If you credit the Foodsoft account with the full amount from each bank transaction,
if you make a debit transaction in Foodsoft when a member buys something offline,
it sounds like something would be missing. What is that?

wvengen:
(add: ignoring membership and deposit for now)

paroga:
When a member transfers money to the foodcoop the foodsoft balance gets increased by exactly that amount (by a person with financial permission). Same happens for the offline paper (but done by the member herself). When a order has been picked up the member adds the total of the order to the offline paper and decreases the balance with the total of the order. The balance in the foodsoft gets not touched.
Every few month a person of with financial permission looks at every sheet of paper and adds one financial transaction for the corresponding ordergroup in the foodsoft (formula: Amount = CurrentFoodsoftBalance - CurrentOfflinePaperBalance). After that the balance in the foodsoft is correct for that date.
Important is that we skip the balancing step in the foodsoft for every order.

IMHO nothing important is missing here. We lose the information for what articles the money has been spent, but that's also some data we do not want to collect at all. And the other thing is that we have to trust every member in keeping the balance on the offline paper correct. But trust is an important part in our foodcoop.

wvengen:
Closing in favour of #407.

PR #368, opened 2015-06-03
Add forum plugin (WiP)

paroga:
Messages alone do not satisfy our communication requirements, so a minimal forum is needed.
- [ ] Fix dependencies (Gemfile, gemspec, README)
- [ ] Use url helpers in navigation menu
- [ ] Figure out how Forem's groups fit together with Foodsoft groups
- [ ] Use Foodsoft layout
- [ ] Include forem assets on forum pages
- [ ] Styling of forem pages
- [ ] Remove gravatar icon
- [ ] Make sense of the navigation menu when no topics are present yet
- [ ] Highlight forem in navigation menu when showing it
- [ ] Move Forem configuration to Foodsoft's configuration section (?)
- [ ] Use Foodsoft configuration for some forem initializer options
- [ ] Use formatter supporting most of current wiki syntax (?) - so users don't need to learn two syntaxes
- [ ] Add link to dashboard (?)
- [ ] Update Foodsoft wiki: plugins need to specify engine source in url helpers
- [ ] Update Foodsoft wiki: engine plugins need to reset `default_url_options`
- [ ] Update Foodsoft wiki: document the use of decorators for monkey-patching Foodsoft
- [ ] ...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,11 +46,14 @@ gem 'recurring_select'
 gem 'roo', '~> 2.0.0'
 gem 'roo-xls'
 gem 'spreadsheet'
+gem 'forem', :github => "radar/forem", :branch => "rails4"
+gem 'forem-bootstrap', :github => "radar/forem-bootstrap"
 
 # we use the git version of acts_as_versioned, and need to include it in this Gemfile
 gem 'acts_as_versioned', github: 'technoweenie/acts_as_versioned'
 gem 'foodsoft_wiki', path: 'plugins/wiki'
 gem 'foodsoft_messages', path: 'plugins/messages'
+gem 'foodsoft_forum', path: 'plugins/forum'
 
 # plugins not enabled by default
 #gem 'foodsoft_uservoice', path: 'plugins/uservoice'
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -85,6 +85,103 @@ ActiveRecord::Schema.define(version: 20150301000000) do
 
   add_index "financial_transactions", ["ordergroup_id"], name: "index_financial_transactions_on_ordergroup_id", using: :btree
 
+  create_table "forem_categories", force: :cascade do |t|
+    t.string   "name",                   null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
+    t.string   "slug"
+    t.integer  "position",   default: 0
+  end
+
+  add_index "forem_categories", ["slug"], name: "index_forem_categories_on_slug", unique: true
+
+  create_table "forem_forums", force: :cascade do |t|
+    t.string  "name"
+    t.text    "description"
+    t.integer "category_id"
+    t.integer "views_count", default: 0
+    t.string  "slug"
+    t.integer "position",    default: 0
+  end
+
+  add_index "forem_forums", ["slug"], name: "index_forem_forums_on_slug", unique: true
+
+  create_table "forem_groups", force: :cascade do |t|
+    t.string "name"
+  end
+
+  add_index "forem_groups", ["name"], name: "index_forem_groups_on_name"
+
+  create_table "forem_memberships", force: :cascade do |t|
+    t.integer "group_id"
+    t.integer "member_id"
+  end
+
+  add_index "forem_memberships", ["group_id"], name: "index_forem_memberships_on_group_id"
+
+  create_table "forem_moderator_groups", force: :cascade do |t|
+    t.integer "forum_id"
+    t.integer "group_id"
+  end
+
+  add_index "forem_moderator_groups", ["forum_id"], name: "index_forem_moderator_groups_on_forum_id"
+
+  create_table "forem_posts", force: :cascade do |t|
+    t.integer  "topic_id"
+    t.text     "text"
+    t.integer  "user_id"
+    t.datetime "created_at"
+    t.datetime "updated_at"
+    t.integer  "reply_to_id"
+    t.string   "state",       default: "pending_review"
+    t.boolean  "notified",    default: false
+  end
+
+  add_index "forem_posts", ["reply_to_id"], name: "index_forem_posts_on_reply_to_id"
+  add_index "forem_posts", ["state"], name: "index_forem_posts_on_state"
+  add_index "forem_posts", ["topic_id"], name: "index_forem_posts_on_topic_id"
+  add_index "forem_posts", ["user_id"], name: "index_forem_posts_on_user_id"
+
+  create_table "forem_subscriptions", force: :cascade do |t|
+    t.integer "subscriber_id"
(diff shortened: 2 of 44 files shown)


wvengen:
Nice, I'll have a look!
You should be able to use normal Rails' route helpers like xyz_forum_path, or [`url_for`](http://api.rubyonrails.org/classes/ActionDispatch/Routing/UrlFor.html#method-i-url_for).

Gemfile
```diff
@@ -46,11 +46,14 @@ gem 'recurring_select'
 gem 'roo', '~> 2.0.0'
 gem 'roo-xls'
 gem 'spreadsheet'
+gem 'forem', :github => "radar/forem", :branch => "rails4"
```
wvengen: Please add this just above the `gem 'foodsoft_forum'` line with a comment like with acts_as_versioned, because it's specific to this plugin.

plugins/forum/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+This plugin adds a forum to foodsoft. A new 'Forum' menu entry is added below the 'Foodcoops' menu in the navigation bar.
+
+This plugin is enabled by default in foodsoft, so you don't need to do anything
+to install it. If you still want to, for example when it has been disabled,
+add the following to foodsoft's Gemfile:
+
+```Gemfile
+gem 'foodsoft_forum', path: 'lib/foodsoft_forum'
+```
```
wvengen: and then also include the forum gems here. It's not enough that they are mentioned in the gemspec, because we need the github version.

plugins/forum/foodsoft_wiki.gemspec
```diff
@@ -0,0 +1,21 @@
...
+  s.email       = ["paroga@paroga.com"]
+  s.homepage    = "https://github.com/foodcoops/foodsoft"
+  s.summary     = "Forum plugin for foodsoft."
+  s.description = "Adds simple forum to foodsoft."
+
+  s.files = Dir["{app,config,db,lib}/**/*"] + ["Rakefile", "README.md"]
+  s.test_files = Dir["test/**/*"]
+
+  s.add_dependency "rails"
+  s.add_development_dependency "sqlite3"
```
wvengen: I'm missing the forem and forem-bootstrap dependencies here.

db/schema.rb
```diff
@@ -85,6 +85,103 @@
 
   add_index "financial_transactions", ["ordergroup_id"], name: "index_financial_transactions_on_ordergroup_id", using: :btree
 
+  create_table "forem_categories", force: :cascade do |t|
```
wvengen: Please either include database schema changes and all migrations from the plugin (by running `rake foodsoft_forum_engine:install:migrations`), _or_ include neither. The database schema is used when creating a new database (e.g. new installs), the migrations for upgrading existing installations. Both are needed.

In this case, I'd say include neither and add instructions on running `rake foodsoft_forum_engine:install:migrations` to the plugin's README. Then when, after some user demand, this might become a default Foodsoft plugin, we add this to the standard database schema.

plugins/forum/lib/foodsoft_forum/engine.rb
```diff
@@ -0,0 +1,20 @@
+module FoodsoftForum
+  class Engine < ::Rails::Engine
+    def navigation(primary, context)
+      return unless FoodsoftForum.enabled?
+      primary.item :forum, I18n.t('navigation.forum'), '/f/forums', id: nil do |subnav|
+        for category in  Forem::Category.all
+          subnav.item :forem_category, category.name, '/f/forums/categories/' + category.name.downcase, id: nil
+        end
+      end
```
wvengen: see http://guides.rubyonrails.org/engines.html#routes

paroga: Yes, but how do I do that? When I tried the routes the "/f" (foodsoft=f) was missing in the url.

wvengen:
A question regarding a UX decision you made: I was a bit startled to click on an item in the navigation menu, and find that I was in a different website. Or so it seemed. If this was intentional, I'd love to hear your reasoning behind it. It it was the result of the technical question of adding javascript and css for forem-pages, you can also do that e.g. with [content_for_in_controllers](https://github.com/clm-a/content_for_in_controllers) (the wiki does that) (+ [engine assets](http://guides.rubyonrails.org/engines.html#assets)).

plugins/forum/lib/foodsoft_forum/engine.rb
wvengen: Hmm I'll try. Perhaps [`default_url_options`](https://github.com/foodcoops/foodsoft/blob/master/app/controllers/application_controller.rb#L186) are not picked up by the engine.

paroga:
Ad UX: I wasn't able to get the foodsoft layout running. Some variables (maybe the default_url_options) were not defined.

plugins/forum/lib/foodsoft_forum/engine.rb
wvengen: Found it! `context.forem.category_path(category)`

You can replace `'/f/forums'` by `'#'` since it's a parent menu item. Or, if you really want it, use `context.forem_path`.

wvengen:
Perfect!

wvengen:
Shall we make a checkbox list in the PR description? Then we and everybody can see what the status is of this PR.

paroga:
I'm fine with it.

wvengen:
FYI, in the past I've connected a phpBB3 forum to Foodsoft (using the [userinfo plugin](https://github.com/foodcoop-adam/foodsoft/tree/beta/lib/foodsoft_userinfo)). I agree forem would be much easier to install and embed.

paroga:
Did you check the point of using the default foodsoft layout too? What is required to get Forem::ApplicationController.layout working with the normal layout?

wvengen:
I hope to find some time later this weekend otherwise week.

wvengen:
https://github.com/foodcoop1040/foodsoft/pull/2

wvengen:
This doesn't address the js/css yet.

paroga:
Many thanks. So I was not wrong that there was no one-line-fix. :-)

paroga:
Basic layout work with your changes, but is there an easy way to avoid that forem appends "?foodcoop=f" to every url? (IMHO it's not a blocker, but would be nice)

wvengen:
I've updated PR https://github.com/foodcoop1040/foodsoft/pull/2 to include a fix for the foodcoop parameter.

wvengen:
If you keep your forum branch updated, I can keep reviewing it now and then, and check boxes that have been done so that others can see how it's going.

paroga:
in the meantime the messages plugin has been improved, which made this PR obsolete

PR #369, opened 2015-06-05
Gemfile: bundle update gave error as git branch of gem "inherited_resources" doesn't exists any more

marcopenhacking:
This helped me fixing a bundle update error (see details). Maybe this is helpful.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ gem 'prawn-table'
 gem 'haml-rails'
 gem 'kaminari'
 gem 'simple_form'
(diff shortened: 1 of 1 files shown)


wvengen:
Ouch, that's not good! I'll check if we can use the normal [gem](https://rubygems.org/gems/inherited_resources) nowadays.

wvengen:
I'll do a 4.4.1 release after this, because it prevents people from installing Foodsoft.

wvengen:
It works with the upstream gem :)

p.s. when updating `Gemfile`, you need to run `bundle install` and commit `Gemfile.lock` as well. That's what the CI error is caused by.

marcopenhacking:
Thanks, update to 4.4.1 worked like a charm.

PR #371, opened 2015-06-07
Use OpenSans TTF for better utf-8 compability in prawn documents.

bennibu:
Recently I got some errors when users trying to generate the order matrix:

```
A Prawn::Errors::IncompatibleStringEncoding occurred in orders#show:

  Your document includes text that's not compatible with the Windows-1252 character set.
If you need full UTF-8 support, use TTF fonts instead of PDF's built-in fonts
.
  app/documents/order_matrix.rb:78:in `body'
```

I don't know which character causes the problems, but it as I understand so far, prawn can't provide full utf-8 without manually embedding a TTF. So picked up the opensans TTF from http://www.fontsquirrel.com/fonts/open-sans and included the basic styles. This fixed the error.


wvengen:
:+1:

Do we need to include a license?
Another option might be [this gem](https://rubygems.org/gems/beardley-open_sans).

bennibu:
> Do we need to include a license?

Oh yes, youre right. These files come with the apache licence (2.0).
This is compatible with GPL3, but we have to mention it. I don't know
where exactly but I guess, we could expand the licence paragraph in the
README.md. On the other hand, there is no copyright owner mentioned, so
I don't know if we have to mention that these files are distributed
under the apache 2 licence?

> Another option might be this gem
> <https://rubygems.org/gems/beardley-open_sans>.

I would like to keep it as simple as possible. And in this case we only
need some ttf files. Not more. This gem added some java installer for
some purpose I don't understand. Do you?

wvengen:
I've merged it and mentioned the license.
No idea if the beardley jars are included in the gem or not. Including in Foodsoft is no problem, I guess.

PR #372, opened 2015-06-07
Do not fail upon missing users in Message#recipients.

bennibu:
I got some errors, when users tries to open old messages with already deleted users in the recipient list. I don't know if this fix is a good solution, but it fixes the problem.


wvengen:
Looks good. I'll see if I can add a test for this.

PR #378, opened 2015-08-10
Clarification in Getting Started workflow.

MikeiLL:
I think this might help clarify the process.


wvengen:
Super, thank you! I hope username and password can be asked for in the interactive setup script, so this notice can be removed at some point. Until then, it's essential.

MikeiLL:
Glad to be able to contribute. Thanks so much for all your wonderful work.

PR #399, opened 2016-02-06
Improved Docker Setup

bennibu:
* Use ruby version 3 with most dependencies already shipped
* Use bundle container
* Run rails commands as app user
* Set Database encoding

Sadly, I couldn't get the test suite to run without some errors. But I guess, this has nothing to do with the docker update as there are also errors in latest travis build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,38 @@
-FROM ruby:2.1-slim
+# Based on ruby image, packed with a lot of development dependencies
+FROM ruby:2.3
 
+# Install all dependencies for development and testing
 RUN apt-get update && \
     apt-get install --no-install-recommends -y \
       mysql-client \
-      git \
-      make \
-      gcc \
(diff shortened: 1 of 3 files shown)


wvengen:
Thanks, looks good! I'll test and merge (tests are currently broken - oops - #400).

bennibu:
@wvengen I forgot to mention. If you have to migrate from the old docker setup be sure to delete old container stuff `(sudo) rm -rf tmp/* log/*` and rebuild the container with `docker-compose build`. Otherwise rails will fail upon missing permissions (in tmp and log files).

wvengen:
Tested and merged, thanks!

wvengen:
Oh, and I added your notice to the CHANGELOG, so people will (hopefully) catch it when upgrading.

bennibu:
Thanks!

Issue #402, opened 2016-03-04
Add message groups


wvengen:
discussion of b895b70840cea2048efb48391fa7c6c6e5e12653

Thanks @paroga for implementing this!
I've some possible changes that I'd like your input on:
* To avoid menus getting even bigger, I'd like to remove it from the menu and instead add a "Message groups" button to the "Messages" index, as well as list message groups in the user profile (with button to the "Message groups" page for subscribing to new groups).
* Let's add an explanation of message groups to the `admin/messagegroups` index (like other groups)
* I'd like to expand the text for the messagegroups index for users to explain what this is.
* Let's call it "message groups" instead of "messagegroups"
* Other that than, great feature :+1:

paroga:
messagegroup have nothing to do with the messages, which are implemented in the plugin. they are more a "everybody can join/leve when (s)he wants"-groups (and therfore has no attached roles). I just choose this name, since it's my primary use case. maybe you have a better one?
What do you thin about introducing an additonal layer "groups" in the menu:
Foodcoop -> Groups -> Ordergroups
Foodcoop -> Groups -> Workinggroups
Foodcoop -> Groups -> Messagegroups (or whatever it will be called)

a agree with the missing explanations, but first want to know what we expect this feature to become exactly (and which name it will get) ;-)

wvengen:
True, it's something like "ad-hoc groups", or "social groups", "circles", "organic groups".
Anyway, I think for now "Message groups" is quite clear since its only purpose right now is to be able to send messages to a group.

Regarding the layers, I'm not that worried too much about the Admin menu, which is for power-users anyway - but I'd like to be easy on 'normal' members.

wvengen:
But I guess an extra layer "Groups" in the admin menu doesn't hurt :)

paroga:
IMHO the message groups will get a second-class citizen if we "hide" it on the messages page.
What about just adding the menu entry only if Messagegroup.count > 0 ?

wvengen:
That would be an improvement! Though I think it primarily belongs with the other messaging options in the user profile, just like with many other services that's where you tell the system what messages you'd like to receive. As a bonus, we can add a "Subscribe to groups" on the messages index button.
But if you're strong on this, I'm open to adding it in the menu.

paroga:
I have no strong opinion on it, just some additional thoughts. :-)
I'll think about it and will create a PR then.

wvengen:
[...]

PR #403, opened 2016-03-04
Add messagegroups

wvengen:
Implements #402.

@paroga what do you think?


paroga:
i agree with it :-)

wvengen:
Rebased. Some locales may be either missing or be superfluous - need to check.

PR #404, opened 2016-03-04
Add break to ordergroup

paroga:
I like to keep the break-info for unlimited time. If it's irrelevant people can remove the data from the group settings anyway


wvengen:
Ok! Let's see what'll happen with this :)

I'll merge it, and update the edit form to put both boxes on the same line.

wvengen:
foodsoft_-_edit_ordergroup_-_2016-03-11_13 56 26

PR #405, opened 2016-03-04
Remove useless if in invoices form

paroga:
The created_at field has a value all the time, so the check if it is set
make no sense since it evaluates to true all the time.


wvengen:
Well found. Merged in b623b6abbf1615664c6df9bbf3b7275ace5265f9.

PR #407, opened 2016-03-04
Add class and type for financial transactions #367

paroga:
This change introduces two new data types to group the financial
transactions. Now every transaction has a "type", which itself belongs
to a "class".
Types should be used add structured information to an transaction, instead
of writing it into the notice textfield. E.g. this could be used to have
different types depending on the source of money (cash vs. bank transfer).
Classes are shown as different columns in the tables and will be uses to
group transactions of specific types. They should be used if not the whole
amount of ordergroup should be used to order fodd. E.g. if there is a
deposit or membership fee, which is independent of the normal credit.
This will allow us to implement additional features based on classes in
the future. E.g. the sum of transactions in the "membership fee" class
must be positive to allow food orders or show a big warning if it is bellow
a certain value.

### DIFF
--- a/app/controllers/admin/financial_transaction_classes_controller.rb
+++ b/app/controllers/admin/financial_transaction_classes_controller.rb
@@ -1,0 +1,41 @@
+class Admin::FinancialTransactionClassesController < Admin::BaseController
+  inherit_resources
+
+  def new
+    @financial_transaction_class = FinancialTransactionClass.new(params[:financial_transaction_class])
+    render layout: false
+  end
+
+  def create
+    @financial_transaction_class = FinancialTransactionClass.new(params[:financial_transaction_class])
+    if @financial_transaction_class.save
+      redirect_to update_transaction_types_admin_finances_url, status: 303
+    else
+      render action: 'new', layout: false
+    end
+  end
+
+  def edit
+    @financial_transaction_class = FinancialTransactionClass.find(params[:id])
+    render action: 'new', layout: false
+  end
+
+  def update
+    @financial_transaction_class = FinancialTransactionClass.find(params[:id])
+
+    if @financial_transaction_class.update_attributes(params[:financial_transaction_class])
+      redirect_to update_transaction_types_admin_finances_url, status: 303
+    else
+      render action: 'new', layout: false
+    end
+  end
+
+  def destroy
+    @financial_transaction_class = FinancialTransactionClass.find(params[:id])
+    @financial_transaction_class.destroy!
+    redirect_to update_transaction_types_admin_finances_url, status: 303
+  rescue => error
+    flash.now[:alert] = error.message
+    render template: 'shared/alert'
+  end
+end
--- a/app/controllers/admin/financial_transaction_types_controller.rb
+++ b/app/controllers/admin/financial_transaction_types_controller.rb
@@ -1,0 +1,42 @@
+class Admin::FinancialTransactionTypesController < Admin::BaseController
+  inherit_resources
+
+  def new
+    @financial_transaction_type = FinancialTransactionType.new(params[:financial_transaction_type])
+    @financial_transaction_type.financial_transaction_class = FinancialTransactionClass.find_by_id(params[:financial_transaction_class]) if params[:financial_transaction_class]
+    render layout: false
+  end
+
+  def create
+    @financial_transaction_type = FinancialTransactionType.new(params[:financial_transaction_type])
+    if @financial_transaction_type.save
+      redirect_to update_transaction_types_admin_finances_url, status: 303
+    else
+      render action: 'new', layout: false
+    end
+  end
+
+  def edit
+    @financial_transaction_type = FinancialTransactionType.find(params[:id])
+    render action: 'new', layout: false
+  end
+
+  def update
+    @financial_transaction_type = FinancialTransactionType.find(params[:id])
+
+    if @financial_transaction_type.update_attributes(params[:financial_transaction_type])
+      redirect_to update_transaction_types_admin_finances_url, status: 303
+    else
+      render action: 'new', layout: false
+    end
+  end
+
+  def destroy
+    @financial_transaction_type = FinancialTransactionType.find(params[:id])
+    @financial_transaction_type.destroy!
+    redirect_to update_transaction_types_admin_finances_url, status: 303
+  rescue => error
+    flash.now[:alert] = error.message
+    render template: 'shared/alert'
+  end
+end
--- a/app/controllers/finance/ordergroups_controller.rb
+++ b/app/controllers/finance/ordergroups_controller.rb
@@ -1,19 +1,17 @@
 class Finance::OrdergroupsController < Finance::BaseController
 
   def index
-    if params["sort"]
-      sort = case params["sort"]
-               when "name" then "name"
-               when "account_balance" then "account_balance"
-               when "name_reverse" then "name DESC"
-               when "account_balance_reverse" then "account_balance DESC"
-             end
+    m = /^(?<col>name|sum_of_class_\d+)(?<reverse>_reverse)?$/.match params["sort"]
+    if m
+      sort = m[:col]
+      sort += ' DESC' if m[:reverse]
     else
       sort = "name"
     end
 
     @ordergroups = Ordergroup.undeleted.order(sort)
-    @ordergroups = @ordergroups.where('name LIKE ?', "%#{params[:query]}%") unless params[:query].nil?
+    @ordergroups = @ordergroups.include_transaction_class_sum
+    @ordergroups = @ordergroups.where('groups.name LIKE ?', "%#{params[:query]}%") unless params[:query].nil?
 
     @ordergroups = @ordergroups.page(params[:page]).per(@per_page)
   end
--- a/app/models/financial_transaction.rb
+++ b/app/models/financial_transaction.rb
@@ -4,6 +4,7 @@ class FinancialTransaction < ActiveRecord::Base
   belongs_to :ordergroup
   belongs_to :user
   belongs_to :financial_link
+  belongs_to :financial_transaction_type
 
   validates_presence_of :amount, :note, :user_id, :ordergroup_id
   validates_numericality_of :amount, greater_then: -100_000,
@@ -11,8 +12,18 @@ class FinancialTransaction < ActiveRecord::Base
 
   localize_input_of :amount
 
+  after_initialize do
+    initialize_financial_transaction_type
+  end
+
   # Use this save method instead of simple save and after callback
   def add_transaction!
-    ordergroup.add_financial_transaction! amount, note, user
+    ordergroup.add_financial_transaction! amount, note, user, financial_transaction_type
+  end
+
+  protected
+
+  def initialize_financial_transaction_type
+    self.financial_transaction_type ||= FinancialTransactionType.default
   end
 end
--- a/app/models/financial_transaction_class.rb
+++ b/app/models/financial_transaction_class.rb
@@ -3,4 +3,18 @@ class FinancialTransactionClass < ActiveRecord::Base
 
   validates :name, presence: true
   validates_uniqueness_of :name
+
+  scope :sorted, -> { order(name: :asc) }
+
+  def self.has_multiple_classes
+    FinancialTransactionClass.count > 1
+  end
+
+  def display
+    if FinancialTransactionClass.has_multiple_classes
+      name
+    else
+      I18n.t('activerecord.attributes.financial_transaction.amount')
+    end
+  end
 end
--- a/app/models/financial_transaction_class.rb
+++ b/app/models/financial_transaction_class.rb
@@ -1,0 +1,6 @@
+class FinancialTransactionClass < ActiveRecord::Base
+  has_many :financial_transaction_types, dependent: :destroy
+
+  validates :name, presence: true
+  validates_uniqueness_of :name
+end
--- a/app/models/financial_transaction_type.rb
+++ b/app/models/financial_transaction_type.rb
@@ -1,0 +1,25 @@
+class FinancialTransactionType < ActiveRecord::Base
+  belongs_to :financial_transaction_class
+  has_many :financial_transactions, dependent: :restrict_with_exception
+
+  validates :name, presence: true
+  validates_uniqueness_of :name
+  validates :financial_transaction_class, presence: true
+
+  before_destroy :restrict_deleting_last_financial_transaction_type
+
+  def self.default
+    first
+  end
+
+  def self.has_multiple_types
+    self.count > 1
+  end
+
+  protected
+
+  # check if this is the last financial transaction type and deny
+  def restrict_deleting_last_financial_transaction_type
+    raise I18n.t('model.financial_transaction_type.no_delete_last') if FinancialTransactionType.count == 1
+  end
+end
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -231,7 +231,7 @@ class Order < ActiveRecord::Base
   end
 
   # Sets order.status to 'close' and updates all Ordergroup.account_balances
-  def close!(user)
+  def close!(user, transaction_type = nil)
     raise I18n.t('orders.model.error_closed') if closed?
     transaction_note = I18n.t('orders.model.notice_close', :name => name,
                               :ends => ends.strftime(I18n.t('date.formats.default')))
@@ -243,7 +243,7 @@ class Order < ActiveRecord::Base
       for group_order in gos
         if group_order.ordergroup
           price = group_order.price * -1                  # decrease! account balance
-          group_order.ordergroup.add_financial_transaction!(price, transaction_note, user)
+          group_order.ordergroup.add_financial_transaction!(price, transaction_note, user, transaction_type)
         end
       end
 
--- a/app/models/ordergroup.rb
+++ b/app/models/ordergroup.rb
@@ -55,9 +55,9 @@ class Ordergroup < Group
 
   # Creates a new FinancialTransaction for this Ordergroup and updates the account_balance accordingly.
   # Throws an exception if it fails.
-  def add_financial_transaction!(amount, note, user, link = nil)
+  def add_financial_transaction!(amount, note, user, transaction_type, link = nil)
     transaction do
-      t = FinancialTransaction.new(ordergroup: self, amount: amount, note: note, user: user, financial_link: link)
+      t = FinancialTransaction.new(ordergroup: self, amount: amount, note: note, user: user, financial_transaction_type: transaction_type, financial_link: link)
       t.save!
       self.account_balance = financial_transactions.sum('amount')
       save!
--- a/app/models/ordergroup.rb
+++ b/app/models/ordergroup.rb
@@ -27,6 +27,18 @@ class Ordergroup < Group
     User.natural_order.all.reject { |u| (users.include?(u) || u.ordergroup) }
   end
 
+  def self.include_transaction_class_sum
+    columns = ['groups.*']
+    FinancialTransactionClass.all.each do |c|
+      columns << "sum(CASE financial_transaction_types.financial_transaction_class_id WHEN #{c.id} THEN financial_transactions.amount ELSE 0 END) AS sum_of_class_#{c.id}"
+    end
+
+    select(columns.join(', '))
+      .joins('LEFT JOIN financial_transactions ON groups.id = financial_transactions.ordergroup_id')
+      .joins('LEFT JOIN financial_transaction_types ON financial_transaction_types.id = financial_transactions.financial_transaction_type_id')
+      .group('groups.id')
+  end
+
   def last_user_activity
     last_active_user = users.order('users.last_activity DESC').first
     if last_active_user
--- a/app/views/finance/financial_transactions/_transactions.html.haml
+++ b/app/views/finance/financial_transactions/_transactions.html.haml
@@ -16,8 +16,12 @@
       - if with_ordergroup
         %th= heading_helper FinancialTransaction, :ordergroup
       %th= heading_helper FinancialTransaction, :user
+      - if FinancialTransactionType.has_multiple_types
+        %th= heading_helper FinancialTransaction, :financial_transaction_type
       %th= sort_link_helper heading_helper(FinancialTransaction, :note), "note"
-      %th= sort_link_helper heading_helper(FinancialTransaction, :amount), "amount"
+      - FinancialTransactionClass.sorted.each do |c|
+        %th
+          = sort_link_helper c.display, "amount"
   %tbody
     - @financial_transactions.each do |t|
       %tr
@@ -29,5 +33,10 @@
         - if with_ordergroup
           %td= h link_to t.ordergroup.name, finance_ordergroup_transactions_path(t.ordergroup)
         %td= h show_user(t.user)
+        - if FinancialTransactionType.has_multiple_types
+          %td= h t.financial_transaction_type.name
         %td= h t.note
-        %td.currency{:style => "color:#{t.amount < 0 ? 'red' : 'black'}; width:5em"}= number_to_currency(t.amount)
+        - FinancialTransactionClass.sorted.each do |c|
+          %td.currency{:style => "color:#{t.amount < 0 ? 'red' : 'black'}; width:5em"}
+            - if t.financial_transaction_type.financial_transaction_class == c
+              = number_to_currency(t.amount)
--- a/app/views/finance/financial_transactions/new.html.haml
+++ b/app/views/finance/financial_transactions/new.html.haml
@@ -5,6 +5,8 @@
 = simple_form_for @financial_transaction, :url => finance_ordergroup_transactions_path(@ordergroup),
   :validate => true do |f|
   = f.hidden_field :ordergroup_id
+  - if FinancialTransactionType.has_multiple_types
+    = f.association :financial_transaction_type, :as => :radio_buttons
   = f.input :amount
   = f.input :note, :as => :text
   .form-actions
--- a/app/views/finance/financial_transactions/new_collection.html.haml
+++ b/app/views/finance/financial_transactions/new_collection.html.haml
@@ -34,6 +34,10 @@
   .well.well-small= t('.sidebar')
 
 = form_tag finance_create_transaction_collection_path do
+  - if FinancialTransactionType.has_multiple_types
+    %p
+      %b= heading_helper FinancialTransaction, :financial_transaction_type
+      = select_tag :type, options_for_select(FinancialTransactionType.order(:name).map { |t| [ t.name, t.id ] })
   %p
     %b= heading_helper FinancialTransaction, :note
     = text_field_tag :note, params[:note], class: 'input-xlarge', required: 'required'
--- a/app/views/home/index.html.haml
+++ b/app/views/home/index.html.haml
@@ -67,15 +67,24 @@
         %tr
           %th= heading_helper FinancialTransaction, :created_on
           %th= heading_helper FinancialTransaction, :user
+          - if FinancialTransactionType.has_multiple_types
+            %th= heading_helper FinancialTransaction, :financial_transaction_type
           %th= heading_helper FinancialTransaction, :note
-          %th= heading_helper FinancialTransaction, :amount
+          - FinancialTransactionClass.sorted.each do |fc|
+            %th
+              = fc.display
         - for ft in current_user.ordergroup.financial_transactions.limit(5).order('created_on DESC')
           %tr
             %td= format_time(ft.created_on)
             %td= h(show_user(ft.user))
+            - if FinancialTransactionType.has_multiple_types
+              %td= h(ft.financial_transaction_type.name)
             %td= h(ft.note)
             - color = ft.amount < 0 ? 'red' : 'black'
-            %td{:style => "color:#{color}; width:5em", :class => "currency"}= number_to_currency(ft.amount)
+            - FinancialTransactionClass.sorted.each do |fc|
+              %td{:style => "color:#{color}; width:5em", :class => "currency"}
+                - if ft.financial_transaction_type.financial_transaction_class == fc
+                  = number_to_currency(ft.amount)
 
 -# placeholder deface to add content using erb[silent]:contains()
 - '<dashboard_bottom_mark>'
--- a/app/views/home/ordergroup.html.haml
+++ b/app/views/home/ordergroup.html.haml
@@ -13,6 +13,11 @@
         %p
           %b= heading_helper(Ordergroup, :available_funds) + ':'
           = number_to_currency(@ordergroup.get_available_funds())
+      - if FinancialTransactionClass.has_multiple_classes
+        - FinancialTransactionClass.sorted.each do |c|
+          %p
+            %b= c.display + ':'
+            = number_to_currency(@ordergroup["sum_of_class_#{c.id}"])
       %p
         %b= heading_helper(Ordergroup, :user_tokens) + ':'
         = @ordergroup.memberships.map{|m| show_user m.user}.join(', ')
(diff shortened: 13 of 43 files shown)


wvengen:
Wow, this is quite a change. I'm glad you explained it before :) I've skimmed through the code, will need to let it sink in a bit again - hope that's ok :)

wvengen:
Would you mind rebasing it? It's not even possible to run the code now (because of `Messagegroup`'s `type` attribute).

paroga:
I never thought that it will get accepted easily. :) But I want at least know if the direction is ok. If yes, I'll make a clean PR with translations and so on.

app/controllers/finance/ordergroups_controller.rb
```diff
@@ -1,20 +1,27 @@
...
     @ordergroups = Ordergroup.undeleted.order(sort)
-    @ordergroups = @ordergroups.where('name LIKE ?', "%#{params[:query]}%") unless params[:query].nil?
+    @ordergroups = @ordergroups.where('groups.name LIKE ?', "%#{params[:query]}%") unless params[:query].nil?
 
     @ordergroups = @ordergroups.page(params[:page]).per(@per_page)
+
+    select = ['groups.*']
+    FinancialTransactionClass.all.each do |c|
+      select << "sum(CASE financial_transaction_types.financial_transaction_class_id WHEN #{c.id} THEN financial_transactions.amount ELSE 0 END) AS sum_of_class_#{c.id}"
+    end
```
wvengen: I wonder if this could become one SQL statement. Perhaps not if we want to keep compatible with MySQL and PostgreSQL. At the very least, we should move this to a helper method, or rather to the model.

spec/integration/balancing_spec.rb
```diff
@@ -1,6 +1,8 @@
 require_relative '../spec_helper'
 
 feature 'settling an order', js: true do
+  let(:ftc) { create :financial_transaction_class }
+  let(:ftt) { create :financial_transaction_type, financial_transaction_class: ftc }
```
wvengen: creating a default transaction class can move to the transaction factory

wvengen:
Thanks for rebasing! With regards to idea and UX, this came to mind:
* Rough idea looks fine.
* FTtypes would be useful for many.
* Let's have a set of useful defaults for FTtypes.
* Do we want a default FTclass or leave it empty?
* I wonder if we should access FTclasses in the configuration screen instead. Depends on how often we think people want to change it, how much can 'break' if people change it without fully understanding it.
* How will plugins be able to do stuff with specific FTtypes, e.g. online payment, imported bank transactions, etc. Also applies to the type for transactions created when an order is settled.
* Perhaps we need a full and abbreviated transaction class name (for table headers).

With regards to code, I noted:
* There's quite some stuff happening in the controllers. I think more could be moved to the model, or if that becomes to much, to concerns.
* Since it's financial transactions, I'm a bit more security-minded, and wonder if we should use strong parameters more.
* Transaction CSV download needs adaptation too (just to remember).

Good work! I'm curious how it will shape up.

spec/integration/balancing_spec.rb
paroga: where is the transaction factory?

paroga:
> Rough idea looks fine.

THX

> FTtypes would be useful for many.

not sure, if you do not have anything built on top of it, note is IMHO enough

> Let's have a set of useful defaults for FTtypes.

a) I do not see useful defaults (might be very specific per foodcoop)
b) If we add them the UI gets more complex (now I hide the type select everywhere when possible)
c) We would need translations for it, which is a problem for texts in the DB

> Do we want a default FTclass or leave it empty?

I think we should have one class and one type, so we do not need to handle NULL values

> I wonder if we should access FTclasses in the configuration screen instead. Depends on how often we think people want to change it, how much can 'break' if people change it without fully understanding it.

sounds reasonable. any good UI idea? having a ADD/DELETE/EDIT function for all of thats stuff might get complex very easy...

> How will plugins be able to do stuff with specific FTtypes, e.g. online payment, imported bank transactions, etc. Also applies to the type for transactions created when an order is settled.

I think that plugins need a config option, where admins can select FTtypes which should be used for specific actions.

> Perhaps we need a full and abbreviated transaction class name (for table headers).

I had it in the beginning, but removed it again, since I use name only (@ col header)

> There's quite some stuff happening in the controllers. I think more could be moved to the model, or if that becomes to much, to concerns.

should be better now

> Since it's financial transactions, I'm a bit more security-minded, and wonder if we should use strong parameters more.

I do not see a big problem since it's guarded with role_finance anyway.

app/models/ordergroup.rb
```diff
@@ -26,6 +26,17 @@ def non_members
...
   end
 
+  def self.include_transaction_class_sum
+    columns = ['groups.*']
+    FinancialTransactionClass.all.each do |c|
+      columns << "sum(CASE financial_transaction_types.financial_transaction_class_id WHEN #{c.id} THEN financial_transactions.amount ELSE 0 END) AS sum_of_class_#{c.id}"
+    end
+
+    select(columns.join(', '))
+      .joins('LEFT JOIN financial_transactions ON groups.id = financial_transactions.ordergroup_id LEFT JOIN financial_transaction_types ON financial_transaction_types.id = financial_transactions.financial_transaction_type_id')
```
wvengen: split into two joins?

app/views/finance/financial_transactions/_transactions.html.haml
```diff
@@ -16,8 +16,15 @@
...
         %th= heading_helper FinancialTransaction, :ordergroup
       %th= heading_helper FinancialTransaction, :user
+      - if FinancialTransactionType.count > 1
+        %th= heading_helper FinancialTransaction, :financial_transaction_type
       %th= sort_link_helper heading_helper(FinancialTransaction, :note), "note"
-      %th= sort_link_helper heading_helper(FinancialTransaction, :amount), "amount"
+      - FinancialTransactionClass.table_columns.each do |c|
+        %th
+          - if c.name?
+            = sort_link_helper c.name, "amount"
```
wvengen: We could have `.name?` in the model fallback to the translation, would work on more places.

app/views/finance/financial_transactions/new_collection.html.haml
```diff
@@ -34,6 +34,12 @@
   .well.well-small= t('.sidebar')
 
 = form_tag finance_create_transaction_collection_path do
+  - if FinancialTransactionType.count > 1
+    %p
+      %b= heading_helper FinancialTransaction, :financial_transaction_type
+      = select_tag :type, options_for_select(FinancialTransactionType.order(:name).map { |t| [ t.name, t.id ] })
+  - else
+    = hidden_field_tag :type, FinancialTransactionType.default
```
wvengen: :) `FinancialTransactionType.any?` could allow tiny database optimizations.

wvengen:
> > FTtypes would be useful for many.
>
> not sure, if you do not have anything built on top of it, note is IMHO enough

There could be a filter on FTtypes; no typos would be made, and people can just use one of the available types, requiring less instruction on what notes are used.
Building on top of it could also be default for settling an order.

> > Let's have a set of useful defaults for FTtypes.
>
> a) I do not see useful defaults (might be very specific per foodcoop)

True. Still a default set that applies to many could be helpful for starting foodcoops. Like payment and sale. And system transaction types like those created when settling an order (and, in plugins or in the future, automatic membership fee and online payments). Perhaps we could indeed start with one default transaction type, but still have system transaction types. But I'm running ahead of myself: let's just introduce transaction types first, then introduce the notion of system transaction types.

> b) If we add them the UI gets more complex (now I hide the type select everywhere when possible)

Agree.

> c) We would need translations for it, which is a problem for texts in the DB

Yep! Either in the database seed that would become language-dependent, or we add a code to transaction types. This could be used to identify system transaction types, as well as translations (which could perhaps be overridden by the name which would be empty by default).

> >  I wonder if we should access FTclasses in the configuration screen instead. Depends on how often we think people want to change it, how much can 'break' if people change it without fully understanding it.
>
> sounds reasonable. any good UI idea? having a ADD/DELETE/EDIT function for all of thats stuff might get complex very easy...

Not fully yet. I'm seeing a single index for both transaction classes and types, something like this (with some more explanation or so):

    Registration                  [edit] [delete]
    
        Registration Payment      [edit] [delete]
        Registration Refund       [edit] [delete]

    Membership                    [edit] [delete]
    
        Membership Payment        [edit] [delete]
        Membership Monthly Fee    [edit] [delete]

Than something like a "(Financial) Transaction Types" menu item would suffice.

> > How will plugins be able to do stuff with specific FTtypes, e.g. online payment, imported bank transactions, etc. Also applies to the type for transactions created when an order is settled.
>
> I think that plugins need a config option, where admins can select FTtypes which should be used for specific actions.

Hmm. I also would like it to work without having to configure this. Not sure if that's feasible.

> > Perhaps we need a full and abbreviated transaction class name (for table headers).
>
> I had it in the beginning, but removed it again, since I use name only (@ col header)
Ok. (Then we'd need to make the editbox for it small as well.)

> > Do we want a default FTclass or leave it empty?
>
> I think we should have one class and one type, so we do not need to handle NULL values

Agreed. In the db seed and in a migration. And not deletable.

> should be better now (regarding controller complexity)

Yes, thanks!

> I do not see a (security) big problem since it's guarded with role_finance anyway.

True.

paroga:
Can you check the new UI in my last commit? If it's ok, I'll fix the tests and add the missing checks to ensure that the last transaction type and class can not be removed.

paroga:
ping?

paroga:
can you look at this too?

wvengen:
Thanks for revisiting this! I need more time for review, hope to get back in a short while.

paroga:
np, just wanted to make sure, that you know that there is something new :-)

paroga:
Did you have already some time to look at this change? For me it would be very important go get at least a ok on the model changes. In that case it's easy to deploy this feature, since I do not have to worry about later database migrations...
If you want I can split the commit into smaller pieces to easy up the review process.

paroga:
I've split the commit into smaller parts to make the review easier.

app/models/financial_transaction_class.rb
```diff
@@ -0,0 +1,16 @@
+class FinancialTransactionClass < ActiveRecord::Base
+  has_many :financial_transaction_types, dependent: :destroy
+
+  validates :name, presence: true
+  validates_uniqueness_of :name
+
+  scope :table_columns, -> { order(name: :asc) }
```
wvengen: I'd like to avoid default scopes, perhaps have a default `sorted` scope if you want to have a default sorting order. (Not `ordered` because we use it in a different sense elsewhere.)

app/models/financial_transaction_type.rb
```diff
@@ -0,0 +1,21 @@
+class FinancialTransactionType < ActiveRecord::Base
+  belongs_to :financial_transaction_class
+  has_many :financial_transactions, dependent: :restrict_with_exception
+
+  validates :name, presence: true
+  validates_uniqueness_of :name
+  validates :financial_transaction_class, presence: true
+
+  before_destroy :check_last_financial_transaction_type
```
wvengen: Maybe something like `restrict_deleting_last_financial_transaction_type` would be just a bit clearer?

app/models/order.rb
```diff
@@ -229,7 +229,7 @@ def finish!(user)
   end
 
   # Sets order.status to 'close' and updates all Ordergroup.account_balances
-  def close!(user)
+  def close!(user, transaction_type)
```
wvengen: This method is used by some (unofficial?) plugins, I'd recommend to revert to the default transaction type when not given.

app/models/ordergroup.rb
```diff
@@ -54,9 +54,9 @@ def get_available_funds(exclude = nil)
 
   # Creates a new FinancialTransaction for this Ordergroup and updates the account_balance accordingly.
   # Throws an exception if it fails.
-  def add_financial_transaction!(amount, note, user)
-    transaction do      
-      t = FinancialTransaction.new(:ordergroup => self, :amount => amount, :note => note, :user => user)
+  def add_financial_transaction!(amount, note, user, transaction_type)
```
wvengen: same here regarding a default transaction type fallback

app/views/finance/financial_transactions/new.html.haml
```diff
@@ -5,6 +5,10 @@
 = simple_form_for @financial_transaction, :url => finance_ordergroup_transactions_path(@ordergroup),
   :validate => true do |f|
   = f.hidden_field :ordergroup_id
+  - if FinancialTransactionType.count > 1
+    = f.association :financial_transaction_type, :as => :radio_buttons
+  - else
+    = f.association :financial_transaction_type, :as => :hidden
```
wvengen: Interesting that this works as a hidden field, didn't know.

config/locales/de.yml
```diff
@@ -1279,6 +1279,8 @@ de:
   model:
     delivery:
       each_stock_article_must_be_unique: Lieferung darf jeden Lagerartikel höchstens einmal auflisten.
+    financial_transaction_type:
+      no_delete_last: Es muss mindestens ein Finanaztransaktionstyp existieren.
```
wvengen: typo?

config/locales/en.yml
```diff
@@ -1288,6 +1288,8 @@ en:
   model:
     delivery:
       each_stock_article_must_be_unique: Each stock article must not be listed more than once.
+    financial_transaction_type:
+      no_delete_last: At least one financital transaction type must exist.
```
wvengen: typo?

config/routes.rb
```diff
@@ -148,7 +148,7 @@
           put :update_note
 
           get :confirm
-          patch :close
+          post :close
```
wvengen: ok

db/seeds/minimal.seeds.rb
```diff
@@ -21,5 +21,9 @@
     :groups => [administrators]
 )
 
+# First entry for financial transaction types
+financial_transaction_class = FinancialTransactionClass.create(:name => "Other")
+FinancialTransactionType.create(:name => "Foodcoop", :financial_transaction_class_id => financial_transaction_class.id)
```
wvengen: Is it required to have a financial transaction type? Else I'd leave it out in the seeds.
It would be nice to have a database seed showcasing more extensive use of financial transaction classes.

spec/integration/balancing_spec.rb
wvengen: Ah, sorry - I meant something else. See comment in factory.

spec/factories/financial_transaction_type.rb
```diff
@@ -0,0 +1,13 @@
+require 'factory_girl'
+
+FactoryGirl.define do
+
+  factory :financial_transaction_class do
+    sequence(:name) { |n| Faker::Lorem.characters(rand(2..12)) + " ##{n}" }
+  end
+
+  factory :financial_transaction_type do
+    sequence(:name) { |n| Faker::Lorem.words(rand(2..4)).join(' ') + " ##{n}" }
```
wvengen: Since a type isn't valid without a class, I'd add the line `financial_transaction_class` to it. See [FactoryGirl#associations](http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md#Associations).

db/seeds/minimal.seeds.rb
paroga: financial transaction type is mandatory

app/models/order.rb
paroga: i've added support for a default type

paroga:
i've updated the first commit, cleaned it a little bit up and rebase the other two commits
ad seeds) yes, i like to have also more stuff but (a) i don't want to make this feature too visible in the first step, (b) don't want to change existing behaviour (people do not need to care about the type ATM) and (c) what about translations? IMHO we can make this in a later step...

app/models/ordergroup.rb
```diff
@@ -54,9 +54,10 @@ def get_available_funds(exclude = nil)
 
   # Creates a new FinancialTransaction for this Ordergroup and updates the account_balance accordingly.
   # Throws an exception if it fails.
-  def add_financial_transaction!(amount, note, user)
-    transaction do      
-      t = FinancialTransaction.new(:ordergroup => self, :amount => amount, :note => note, :user => user)
+  def add_financial_transaction!(amount, note, user, transaction_type = nil)
+    transaction_type = FinancialTransactionType.default if transaction_type.nil?
+    transaction do
+      t = FinancialTransaction.new(:ordergroup => self, :amount => amount, :note => note, :user => user, :financial_transaction_type => transaction_type)
```
wvengen: What about moving the default transaction type to `FinancialTransaction` itself? In that way plugins creating a financial transaction using the constructor (I've done it at some point, but not sure if that survived or not in the final plugin), would still work. Perhaps an `after_initialize` there to set the default transaction, documenting that it is for backward compatibility.
What do you think?

db/seeds/minimal.seeds.rb
wvengen: Ah, I understand now.

wvengen:
I'll continue with the other commits later. Thanks for your patience!

app/models/ordergroup.rb
paroga: i like the idea and updated the commit

paroga:
ping?

app/views/finance/financial_transactions/_transactions.html.haml
```diff
@@ -16,14 +16,23 @@
       - if with_ordergroup
         %th= heading_helper FinancialTransaction, :ordergroup
       %th= heading_helper FinancialTransaction, :user
+      - if FinancialTransactionType.count > 1
```
wvengen: `.any?` (more simple database call)

app/models/financial_transaction_class.rb
```diff
@@ -3,4 +3,14 @@ class FinancialTransactionClass < ActiveRecord::Base
 
   validates :name, presence: true
   validates_uniqueness_of :name
+
+  scope :sorted, -> { order(name: :asc) }
+
+  def display
+    if FinancialTransactionClass.count > 1
```
wvengen: `any?` can result in a simpler db-call

app/views/home/index.html.haml
```diff
@@ -53,15 +53,24 @@
         %tr
           %th= heading_helper FinancialTransaction, :created_on
           %th= heading_helper FinancialTransaction, :user
+          - if FinancialTransactionType.count > 1
```
wvengen: same here

app/views/home/index.html.haml
```diff
@@ -53,15 +53,24 @@
...
           %th= heading_helper FinancialTransaction, :note
-          %th= heading_helper FinancialTransaction, :amount
+          - FinancialTransactionClass.sorted.each do |fc|
+            %th
+              = fc.display
         - for ft in current_user.ordergroup.financial_transactions.limit(5).order('created_on DESC')
           %tr
             %td= format_time(ft.created_on)
             %td= h(show_user(ft.user))
+            - if FinancialTransactionType.count > 1
```
wvengen: same here

app/views/home/ordergroup.html.haml
```diff
@@ -9,6 +9,11 @@
       %p
         %b= heading_helper(Ordergroup, :available_funds) + ':'
         = number_to_currency(@ordergroup.get_available_funds())
+      - if FinancialTransactionClass.count > 1
+        - FinancialTransactionClass.sortd.each do |c|
```
wvengen: typo?

app/views/home/ordergroup.html.haml
```diff
@@ -9,6 +9,11 @@
       %p
         %b= heading_helper(Ordergroup, :available_funds) + ':'
         = number_to_currency(@ordergroup.get_available_funds())
+      - if FinancialTransactionClass.count > 1
```
wvengen: same here

app/controllers/admin/financial_transaction_classes_controller.rb
```diff
@@ -0,0 +1,41 @@
+class Admin::FinancialTransactionClassesController < Admin::BaseController
+  inherit_resources
+
+  def new
+    @financial_transaction_class = FinancialTransactionClass.new(params[:financial_transaction_class])
+    render :layout => false
```
wvengen: I would like to transition to new-style options `layout: false`, any objections?

app/controllers/admin/financial_transaction_classes_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def create
+    @financial_transaction_class = FinancialTransactionClass.new(params[:financial_transaction_class])
+    if @financial_transaction_class.valid? && @financial_transaction_class.save
```
wvengen: I think that [`save`](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save) already checks for validity, and returns that.

app/controllers/admin/financial_transaction_classes_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    if @financial_transaction_class.update_attributes(params[:financial_transaction_class])
+      redirect_to update_transaction_types_admin_finances_url, :status => 303
+    else
+      render :action => 'new', :layout => false
+    end
+  end
+
+  def destroy
+    @financial_transaction_class = FinancialTransactionClass.find(params[:id])
+    @financial_transaction_class.destroy
```
wvengen: `destroy!`?

app/controllers/admin/financial_transaction_types_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  inherit_resources
+
+  def new
+    @financial_transaction_type = FinancialTransactionType.new(params[:financial_transaction_type])
+    render :layout => false
+  end
+
+  def create
+    @financial_transaction_type = FinancialTransactionType.new(params[:financial_transaction_type])
+    if @financial_transaction_type.valid? && @financial_transaction_type.save
```
wvengen: same as before

app/controllers/admin/financial_transaction_types_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    if @financial_transaction_type.update_attributes(params[:financial_transaction_type])
+      redirect_to update_transaction_types_admin_finances_url, :status => 303
+    else
+      render :action => 'new', :layout => false
+    end
+  end
+
+  def destroy
+    @financial_transaction_type = FinancialTransactionType.find(params[:id])
+    @financial_transaction_type.destroy
```
wvengen: same here `destroy!`?

app/views/finance/financial_transactions/_transactions.html.haml
paroga: any is not >1 it's >0

app/controllers/admin/financial_transaction_classes_controller.rb
paroga: agree

app/views/finance/financial_transactions/_transactions.html.haml
wvengen: oops I was sleeping :o, sorry

app/views/home/ordergroup.html.haml
wvengen: What about introducing a method on `FinancialTransactionClass` that tells whether the financial transactions class feature is needed?

paroga: +1, any good name?

paroga:
new commits pushed

paroga:
ping?

app/controllers/admin/financial_transaction_classes_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def new
+    @financial_transaction_class = FinancialTransactionClass.new(params[:financial_transaction_class])
+    render layout: false
+  end
+
+  def create
+    @financial_transaction_class = FinancialTransactionClass.new(params[:financial_transaction_class])
+    if @financial_transaction_class.save
+      redirect_to update_transaction_types_admin_finances_url, status: 303
```
wvengen: I don't think we use 303 status elsewhere, seems like a good idea to me.

db/migrate/20160219123220_create_financial_transaction_class_and_types.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.references :financial_transaction_class, :null => false
+    end
+
+    change_table :financial_transactions do |t|
+      t.references :financial_transaction_type
+    end
+
+    execute "INSERT INTO financial_transaction_classes (id, name) VALUES (1, 'Other')"
+    execute "INSERT INTO financial_transaction_types (id, name, financial_transaction_class_id) VALUES (1, 'Foodsoft', 1)"
+    execute "UPDATE financial_transactions SET financial_transaction_type_id = 1"
```
wvengen: this isn't really reversible, is it?

wvengen:
I think in terms of functionality, it's there. Great!
In terms of user-experience, I think there is some way to go. Right now, as a user seeing the new screen, I have no clue what to dot. Adding explaining texts would already help. It is a bit of a difficult concept, so I'm not really sure yet how to keep it short and clear at the same time. But let's see where we get.

Small remarks and ideas that could improve it:
* _Do you really want to delete ..._ could use an indication of what kind is being deleted, "_Do you really want to delete Foodsoft?_" can be a bit frightening.
* Does there need to be a default transaction type for settling an order?
* I don't think the user needs to fully understand _class_ and _type_, UX could help a lot here. E.g.
  - a `+` button behind the class name to add a new child type
  - above the table an 'Add new class' button. 
* What happens when a financial type is deleted? Should it be possible when there are transactions linked to it?
* Perhaps _Default_ is a more suitable name for the transaction instance created on migration?
* Do you think a description for each class (and maybe type) would add comprehensibility? Person setting up the classes would enter them. Could be shown in small font below the name.
* Could also add a user-visible description on hover in account statement to explain. (But if you don't think this is necessary, please ignore.)
* Just a reminder: don't forget to complete i18n :)
* Nice to have: provide a link to the financial transactions list of each type

All in all, good work. Thanks!

paroga:
> this isn't really reversible, is it?

It's reversible if you delete the new tables and the column.

> In terms of user-experience, I think there is some way to go. Right now, as a user seeing the new screen, I have no clue what to dot. Adding explaining texts would already help. It is a bit of a difficult concept, so I'm not really sure yet how to keep it short and clear at the same time. But let's see where we get.

It's the i18n for the "paragraph"...

> Do you really want to delete ... could use an indication of what kind is being deleted, "Do you really want to delete Foodsoft?" can be a bit frightening.

IMHO getting rid of all the different delete messages in the past was a good idea. I don't like to reintroduce them.

> Does there need to be a default transaction type for settling an order?

You need at least one transaction type, which will be hidden if it's only one.

> a + button behind the class name to add a new child type

Ok for me.

> What happens when a financial type is deleted? Should it be possible when there are transactions linked to it?

ATM you can not delete a type, which is already used.

> Perhaps Default is a more suitable name for the transaction instance created on migration?

"Default" is a language-specific term, but ok for me.

> Do you think a description for each class (and maybe type) would add comprehensibility?

IMHO an improvement for the next steps. (I didn't want to make the already big change even bigger^^)

> Just a reminder: don't forget to complete i18n :)

No, as written in the first statement. It's also necessary for explanation.

> Nice to have: provide a link to the financial transactions list of each type

An additional improvement for a coming PR.

db/migrate/20160219123220_create_financial_transaction_class_and_types.rb
wvengen:
```
$ rake db:migrate:down VERSION=20160219123220
== 20160219123220 CreateFinancialTransactionClassAndTypes: reverting ==========
rake aborted!
StandardError: An error has occurred:

ActiveRecord::IrreversibleMigration
```

wvengen:
Thanks for the answers. Sounds good. Remaining:

> > Does there need to be a default transaction type for settling an order?
> 
> You need at least one transaction type, which will be hidden if it's only one.

If there are multiple transactions types, it might be useful to have a default (not just the first). But this can be an improvement later, if needed.


> "Default" is a language-specific term, but ok for me.

If this only applies to English, that's ok. "Other" sounds like "things outside of the ordinary" to me, while "Default" is more like "unless you have a better idea, this is what to use".



> >  Do you think a description for each class (and maybe type) would add comprehensibility?
> 
> IMHO an improvement for the next steps. (I didn't want to make the already big change even bigger^^)

:+1:

paroga:
>> "Default" is a language-specific term, but ok for me.

>If this only applies to English, that's ok. "Other" sounds like "things outside of the ordinary" to me, while "Default" is more like "unless you have a better idea, this is what to use".

I mean that it would require different database migrations for each language. (e.g. is should be "Standard" in German)

wvengen:
Hmmm ... that's a good one. Wouldn't _Standard_ make sense, as it is similar to some other languages?
For database migrations, I think getting the language (not using the settings infrastructure because that may change in the future, but by a direct query) and choosing one of the translations defined in the migration (because also locales may change) would actually work.
Now database seeds may be localized as well (though they really aren't right now, and one cannot provide a language for database creation). So for the future, translations would be ok, I think.

paroga:
I've updated the PR. Fixed the migration, changed the default class name and adopted the UI.
IMHO the only missing part ist the I18N now (which I don't want to do more than once). Can you crosscheck that? If you agree with that I'll add the translations and hope to finish this PR then finally. :-)

app/models/financial_transaction.rb
```diff
@@ -3,15 +3,20 @@
...
+  belongs_to :financial_transaction_type
 
   validates_presence_of :amount, :note, :user_id, :ordergroup_id
   validates_numericality_of :amount, greater_then: -100_000,
     less_than: 100_000
 
   localize_input_of :amount
 
+  after_initialize do
+    self.financial_transaction_type = FinancialTransactionType.default unless self.financial_transaction_type
```
wvengen: Shorter would be
```ruby
self.financial_transaction_type ||= FinancialTransactionType.default
```

wvengen: Moving this to a (private) method and using `after_initialize` with a symbol would make it just even a bit more flexible (allow a plugin to override it).

app/models/financial_transaction_type.rb
```diff
@@ -0,0 +1,25 @@
+class FinancialTransactionType < ActiveRecord::Base
+  belongs_to :financial_transaction_class
+  has_many :financial_transactions, dependent: :restrict_with_exception
+
+  validates :name, presence: true
+  validates_uniqueness_of :name
+  validates :financial_transaction_class, presence: true
+
+  before_destroy :restrict_deleting_last_financial_transaction_type
```
wvengen: :+1:

paroga:
@wvengen ping?

wvengen:
Would prefer to look at this after the foodcoops.net transition is done, ok?

paroga:
sure, just wanted to ping, for the case you didn't see the new commits...

wvengen:
Thanks, I hadn't seen them yet :)

paroga:
4.6 released, so it should be ok to merge it into master finally?

wvengen:
What's going on with the specs?

wvengen:
What's going on with Travis?
Also, did you consider where to write specs for? It adds complexity, and would be good for new contributions to not break this inadvertently.

paroga:
It mainly adds complexity to the views and not to the model, so I do not see a lot of code to test. For what do you want a test for?

I'll check travis, if the rest is fine.

wvengen:
After going through the code again, I agree that there's not actually that much new functionality.
I don't have this PR fully clear in my head, but I'll try: maybe `include_transaction_class_sum` or is that really internal stuff?

wvengen:
Other than these, I'd say, go!

paroga:
i've added a test for `include_transaction_class_sum` and fixed travis. imho it should be ready to merge now

paroga:
@wvengen ping

paroga:
@wvengen ping

PR #410, opened 2016-03-04
Count only undeleted users for the user_count expansion varibale

paroga:


wvengen:
Hadn't thought of that! :+1:

PR #414, opened 2016-03-06
Add a button to add all ordergroups to a new finanacial transaction c…

paroga:
…ollection

This function can be used if we need to add the same financial transaction
to all ordergroups at the same time. E.g. the monthly membership fee.


wvengen:
:)

PR #416, opened 2016-03-06
Add config option to disable the invites

paroga:
Some foodcoops require that an admin adds all users to have more control
over the users.


wvengen:
Nice!

Issue #398, opened 2016-01-26
Wrong available Credit shwon on Dashboard


wedel:
**The Dashboard lists the available Credit of your ordergroup. I experienced, that in some cases there is a wrong credit shown, while the account statement lists the right balance of the account.**
Credit and Balance should be the same. Seems like a bug to me.

Here is a screenshot of the credit, listed at my dashboard:
![dashboard_guthaben](https://cloud.githubusercontent.com/assets/1165337/12581337/3ae641ca-c436-11e5-91cc-b0dd1c205027.png)

and, at the same time, here is a screenshot which shows the Balance of the same account in the account overview:
![financial_transactions_kontouebersicht](https://cloud.githubusercontent.com/assets/1165337/12581362/662c2b38-c436-11e5-8583-292b43a59da3.png)

wedel:
Just checked the code. [get_available_funds()](https://github.com/foodcoops/foodsoft/blob/61a5314b9557aabd914969ce5a7af9361f614a14/app/models/ordergroup.rb#L42) returns the account balance minus the price of all non-closed Orders. So it's not a bug but somehow very confusingly.

wvengen:
Hi, thanks for your note (incl. screenshots, great!). I have found this quite confusing as well.

In the foodcoop-adam fork, I've tried to hide the different between account balance and available funds like this:

![foodsoft_-_my_ordergroup_-_2016-01-29_11 49 29](https://cloud.githubusercontent.com/assets/503804/12673665/602f14bc-c67e-11e5-9a20-d1bdcd69358d.png)

Here are finished-but-not-closed orders (in foodsoft-model-terms) are listed with a star, with mouseover text: _This transaction is still being processed, it will become final when the order is administrated._

PR #418, opened 2016-03-06
Add option to show account balance instead of available funds

paroga:
Some foodcoops do not use the accounting feature of the foodsoft, which
results in wrong calculation of the available credit. To avoid
confusions show the current account balance instead.

### DIFF
--- a/app/views/finance/balancing/new.html.haml
+++ b/app/views/finance/balancing/new.html.haml
@@ -53,13 +53,16 @@
       class: 'btn'
   .btn-group
     - unless @order.closed?
-      = link_to t('.confirm_order'), confirm_finance_order_path(@order), class: 'btn btn-primary'
-      = link_to '#', data: {toggle: 'dropdown'}, class: 'btn dropdown-toggle' do
-        %span.caret
-      %ul.dropdown-menu
-        %li= link_to t('.confirm_order'), confirm_finance_order_path(@order)
-        %li= link_to t('.close_direct'), close_direct_finance_order_path(@order), method: :patch,
-          data: {confirm: t('.close_direct_confirm')}
+      - if FoodsoftConfig[:charge_members_manually]
+        = link_to t('.confirm_order'), close_direct_finance_order_path(@order), class: 'btn btn-primary', method: :patch
+      - else
+        = link_to t('.confirm_order'), confirm_finance_order_path(@order), class: 'btn btn-primary'
+        = link_to '#', data: {toggle: 'dropdown'}, class: 'btn dropdown-toggle' do
+          %span.caret
+        %ul.dropdown-menu
+          %li= link_to t('.confirm_order'), confirm_finance_order_path(@order)
+          %li= link_to t('.close_direct'), close_direct_finance_order_path(@order), method: :patch,
+            data: {confirm: t('.close_direct_confirm')}
 
   #editOrderNav.btn-group
     = link_to '#', data: {toggle: 'dropdown'}, class: 'btn dropdown-toggle' do
--- a/app/views/group_orders/_form.html.haml
+++ b/app/views/group_orders/_form.html.haml
@@ -1,8 +1,9 @@
 - content_for :javascript do
+  - group_balance = FoodsoftConfig[:charge_members_manually] ? @ordering_data[:account_balance] : @ordering_data[:available_funds]
   :javascript
     $(function() {
       #{data_to_js(@ordering_data)}
-      setGroupBalance(#{@ordering_data[:available_funds]});
+      setGroupBalance(#{group_balance});
       setMinimumBalance(#{FoodsoftConfig[:minimum_balance] or 0});
       setToleranceBehaviour(#{FoodsoftConfig[:tolerance_is_costly]});
       setStockit(#{@order.stockit?});
@@ -45,8 +46,11 @@
         %dd
           = show_user(@group_order.updated_by)
           (#{format_time(@group_order.updated_on)})
-      %dt= heading_helper Ordergroup, :available_funds
-      %dd= number_to_currency(@ordering_data[:available_funds])
+      %dt= heading_helper Ordergroup, :account_balance
+      %dd= number_to_currency(@ordering_data[:account_balance])
+      - unless FoodsoftConfig[:charge_members_manually]
+        %dt= heading_helper Ordergroup, :available_funds
+        %dd= number_to_currency(@ordering_data[:available_funds])
 
   .well.pull-right
     = close_button :alert
@@ -156,13 +160,19 @@
           %td.currency
             %span#total_price= number_to_currency(@group_order.price)
         %tr
-          %td= heading_helper(Ordergroup, :available_funds) + ':'
-          %td.currency= number_to_currency(@ordering_data[:available_funds])
+          - if FoodsoftConfig[:charge_members_manually]
+            - old_balance = @ordering_data[:account_balance]
+            %td= heading_helper(Ordergroup, :account_balance) + ':'
+            %td.currency= number_to_currency(@ordering_data[:account_balance])
+          - else
+            - old_balance = @ordering_data[:available_funds]
+            %td= heading_helper(Ordergroup, :available_funds) + ':'
+            %td.currency= number_to_currency(@ordering_data[:available_funds])
         %tr
           %td= t('.new_funds') + ':'
           %td.currency
             %strong
-              %span#new_balance= number_to_currency(@ordering_data[:available_funds] - @group_order.price)
+              %span#new_balance= number_to_currency(old_balance - @group_order.price)
       #order-button
         = submit_tag( t('.action_save'), id: 'submit_button', class: 'btn btn-primary' )
         #{link_to t('ui.or_cancel'), group_orders_path}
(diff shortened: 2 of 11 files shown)


wvengen:
I like it that support for different workflows is improved in Foodsoft.

Leaving the order forever in the `finished(_but_not_closed)` state is not something I'd like to promote, though. At the end of every order cycle, orders need to be settled.
If foodcoops don't do this, amounts can forever be adjusted in retrospect, which is bound to happen at some point, and it'll be next to impossible to figure out what the correct amounts are. And this has an immediate effect on ordergroups' available funds.
Also I can't guarentee that performance remains sane if the number of unsettled orders keeps increasing.

You may understand I'm hesitant to implement this functionality.
Improving the financial overview: yes #398
What about the option to automatically settle orders a number of days after close/pickup?

wvengen:
I like it that support for different workflows is improved in Foodsoft.

Leaving the order forever in the `finished(_but_not_closed)` state is not something I'd like to promote, though. At the end of every order cycle, orders need to be settled.
If foodcoops don't do this, amounts can forever be adjusted in retrospect, which is bound to happen at some point, and it'll be next to impossible to figure out what the correct amounts are. And this has an immediate effect on ordergroups' available funds.
Also I can't guarentee that performance remains sane if the number of unsettled orders keeps increasing.

You may understand I'm hesitant to implement this functionality.
Improving the financial overview: yes #398
What about the option to automatically settle orders a number of days after close/pickup?

paroga:
We still close our orders and it's important to do that. But it's only relevant for the role_finance. "Normal" users should just see their current account_balance.

wvengen:
That's a relief to hear :) And true, this is not something that normal members need to care about. The only thing that matters, is that they can trust that financial transactions are set in stone, while not-yet-settled orders may change for now. That's why I'd like https://github.com/foodcoops/foodsoft/issues/398#issuecomment-176691743 for the financial transactions.

Then we have the balance in the member ordering screen. When there are multiple open orders, this would help to know how much you're spending across all orders. But that only works when all non-open orders are settled, which is not always the case (or even achievable).

What about we use account_balance - value_of_finished_orders as the amount the user is able to spend in the currently open orders? That we can show at the bottom of the ordering screen, as well as value_of_open_orders (incl. the currently open order) as what the user is spending right now.

paroga:
That's a problem for us: As you know we have our "offline balance" which we regulary import into foodsoft. The following is a valid use case:
1) User transfers 100 EUR
2) New `finacial_transaction` with 100 EURO (foodsoft balance = 100)
3) User adds money to offline paper (paper balance = 100)
4) User orders food for 50 EUR
5) Order gets closed
6) User picks up food and writes the amount into the offline paper (new balance = 50)
7) Current paper balance gets synced with foodsoft (foodsoft balance = 50)
8) User CAN NOT order (`account_balance - finisehd_orders` = 0)
9) Financial team closes order
10) User CAN order AGAIN (`account_balance - finisehd_orders` = 50)

We we close the order when we have a invoice for it, which can take some time sometimes. I do not want to close orders before everything is clear from the invoice.

wvengen:
It took me to read it a couple of times, but, yes, I concur! That means you're using the "settle without updating account balance" function, right?

paroga:
yes, exactly

wvengen:
Summarizing, I see these flows:

1. Everything is handled within Foodsoft, accounts are charged by settling orders.
2. Members are charged by manual transactions.
  - 2a. Finished orders are substracted from the available funds, so that there is as little chance as possible that a member may spend more than he can. The time between entering transactions for charging, and settling orders is short. (During that time, it looks like members are charged about twice.)
  - 2b. Finished orders are not counted in the available funds, so that members can always order when new orders are opened, even when older orders haven't been settled yet. When their account is charged later, the account balance may become negative if the member ordered a lot.

You are talking about situation 2b. I'd like to propose to support 1 and 2b, allow to choose between those in the configuration screen, with a short explanation. This may even control the visibility/function of the "settle without charging members" button in the accounting screen.

In case 2b, we could show finished orders slightly different in the financial transactions screen, so that it's still clear how the numbers add up.

What do you think?

paroga:
agree! do we want to make the workflow an option or keep different options? i also like to have an option to make "settle without updating account balance" the default (or only option)

wvengen:
Great! I think I'd like to move to the workflow configuration options, and even try to see if we can remove the individual ones (because I've found them to be confusing to all but Foodsoft-experts). And changing the default depending on selected workflow, yesss.

paroga:
so this change should be ok for now if i change the option to a "select workflow"? do you have good names for the workflows?

wvengen:
I think this would fit as a new section in the _Finances_ config tab, above schedule. Something like:
~~~
Configuration | Foodcoop | Finances | Tasks | ...
-------------------------'          `---------------

  Foodcoop margin   [    2 ] %
      Default VAT   [    7 ] %
  Minimum balance   € [ -500 ]


* Financial Workflow

  (*) Charge members automatically when settling orders

        Choose this option when you keep the administration of
        who received what up-to-date in Foodsoft. In the balancing
        screen, settling the order will charge the amounts from
        member accounts. (This is the default option.)


  ( ) Charge members manually

        When you keep track of who received what elsewhere (e.g. on paper),
        and you don't want to enter this into Foodsoft, select this.
        You'll need to charge member accounts manually (using `Add new transactions`).
        You still need to settle orders in the balancing screen, but this
        will not charge any member accounts.


* Ordering schedule

        Order ends  [ not recurring____ ] time [     ]
    Box-fill phase  [ ]

paroga:
Cool. THX. I'll update the PR.

wvengen:
Awesome. Looking forward to it!

paroga:
unfortunately i wasn't able to build the proposed UI without a lot of work, but it should be clearer compared to my initial proposal now

paroga:
@wvengen ping

paroga:
@wvengen ping

wvengen:
Thanks for the bump, it's still on my list (but foodcoops.net launch first).

app/views/finance/balancing/new.html.haml
```diff
@@ -53,13 +53,16 @@
...
+      - if FoodsoftConfig[:charge_members_manually]
+        = link_to t('.confirm_order'), close_direct_finance_order_path(@order), class: 'btn btn-primary', method: :patch
+      - else
+        = link_to t('.confirm_order'), confirm_finance_order_path(@order), class: 'btn btn-primary'
+        = link_to '#', data: {toggle: 'dropdown'}, class: 'btn dropdown-toggle' do
+          %span.caret
+        %ul.dropdown-menu
+          %li= link_to t('.confirm_order'), confirm_finance_order_path(@order)
+          %li= link_to t('.close_direct'), close_direct_finance_order_path(@order), method: :patch,
+            data: {confirm: t('.close_direct_confirm')}
```
wvengen: Maybe we better remove the close direct option for automatic charging in a future version.

app/views/group_orders/_form.html.haml
```diff
@@ -156,13 +160,19 @@
...
+          - else
+            - old_balance = @ordering_data[:available_funds]
+            %td= heading_helper(Ordergroup, :available_funds) + ':'
+            %td.currency= number_to_currency(@ordering_data[:available_funds])
         %tr
           %td= t('.new_funds') + ':'
           %td.currency
             %strong
-              %span#new_balance= number_to_currency(@ordering_data[:available_funds] - @group_order.price)
+              %span#new_balance= number_to_currency(old_balance - @group_order.price)
```
wvengen: Would this work when there are multiple open orders? At a glance it looks like only the order currently viewed is substracted from the balance for manual charging.

config/locales/en.yml
```diff
@@ -489,6 +489,7 @@ en:
   config:
     hints:
       applepear_url: Website where the apple and pear system for tasks is explained.
+      charge_members_manually: When you keep track of who received what elsewhere (e.g. on paper), and you don't want to enter this into Foodsoft, select this. You'll need to charge member accounts manually (using "Add new transactions"). You still need to settle orders in the balancing screen, but this will not charge any member accounts.
```
wvengen: Clear :+1:

app/views/finance/balancing/new.html.haml
paroga: IMHO it's still usefull sometmes, when e.g. a ordner wasn't deliverd and nobody should get charged.

app/views/group_orders/_form.html.haml
paroga: The `charge_members_manually` mode works as intended: The calculation uses only the current account balance, so no other orders are involved. For the "normal" mode, it works like without this change.

wvengen: I can see this work well with `charge_members_manually` when there is one open order. In other cases, only the current group order total is substracted from the account balance, while there may be other open orders. As a member, I would find it confusing. If just the old balance would be shown, which doesn't change when the group order changes, that may be clearer.

There have been many situations where ordering users didn't really understand what Foodsoft was doing and why, that's why I'm sensitive to things that may be confusing, especially regarding numbers and amounts.

But then I don't have a clear overview of the whole manual ordering flow, so I leave this to your judgement.

paroga: i agree with your point. IMHO the best solution would be to hide it completely, but that would require a bigger UI change (since members shouldn't be allowed to buy unlimited stuff). The relevant change is, that there is no need to put money to the ordergroup and it's possible to have unlimited `finished` an not `closed` orders (let's just ignore the fact, that having a lot of `finished` orders is not a good idea).

PR #419, opened 2016-03-06
Improve sorting of wiki pages

paroga:
Sorting is completely broken. Also align the sorting UI with other pages.

### DIFF
--- a/plugins/wiki/app/views/pages/all.html.haml
+++ b/plugins/wiki/app/views/pages/all.html.haml
@@ -6,7 +6,6 @@
 .navbar
   .navbar-inner
     %ul.nav
-      %li= link_to t('.recent_changes'), all_pages_path(:view => 'recent_changes')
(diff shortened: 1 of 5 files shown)


plugins/wiki/app/views/pages/all.html.haml
```diff
@@ -6,7 +6,6 @@
 .navbar
   .navbar-inner
     %ul.nav
-      %li= link_to t('.recent_changes'), all_pages_path(:view => 'recent_changes')
```
wvengen: That's sad, to miss this one?

wvengen: Ah, now I get it. I guess it's better like this - hope people will not get confused.

paroga: we do not miss this functionality! Now it's implemented via sort_helper like everywhere else. it's an usability improvement to make all pages behave in a similar way :-)

PR #421, opened 2016-03-07
Add page to edit deliveries and orders linked to an invoice

paroga:


wvengen:
This uses a plain html multi select, right? I think the usability of that control is terrible, perhaps worse than no edit functionality. At the very minimum, let's use select2 here (perhaps making each tag a block-level element).

Ideally, I'd think of something like the top of the new delivery screen:

~~~
Name               starts       ends          pickup
Farmer             2015-12-01   2015-12-10    2015-12-15     (x)
Brewer             2015-12-05   2015-12-09    2015-12-15     (x)

| Add order or delivery____________________________ |v| (select2 box)

paroga:
is there any example implementation which i can use as (easy) base? simple_form is only a few LOC and i do not want to reinvent the wheel ;-)

wvengen:
Not that I know of, I'm afraid :(
The simplest thing now may be to just use select2 on the multiselect, and give it some styling.

wvengen:
I'm wondering - how do we want orders and deliveries to be selected, which ones?
Group by pickup-date (though not yet merged)?
Sort by date desc, I suppose.
The list of orders may grow in the hundreds, so ajax would be preferred (but not absolutely required).

paroga:
Sorting by pickup date is preferred. We only show orders which are `finished` and not assigned. Thats I manageable size of items. I worked with it on our instance with a hundrets of orders and invoices and it only showed the relevant stuff.

wvengen:
Sounds good!

wvengen:
I like the UI, well done!

app/helpers/finance/invoices_helper.rb
```diff
@@ -0,0 +1,8 @@
+module Finance::InvoicesHelper
+  def format_delivery_item delivery
+    format_date(delivery.delivered_on)
+  end
+  def format_order_item order
+    "#{format_date(order.ends)} (#{number_to_currency(order.sum)}"
```
wvengen: Missing end bracket :o

app/controllers/finance/invoices_controller.rb
```diff
@@ -14,9 +14,21 @@ def new
...
+  end
+
+  def form_on_supplier_id_change
+    fill_deliveries_and_orders_collection params[:invoice_id], params[:supplier_id]
+    render :layout => false
+  end
+
+  def fill_deliveries_and_orders_collection(invoice_id, supplier_id)
+    @deliveries_collection = Delivery.where('invoice_id = ? OR invoice_id IS NULL AND supplier_id = ?', invoice_id, supplier_id).order(:delivered_on)
+    @orders_collection = Order.where('invoice_id = ? OR invoice_id IS NULL AND supplier_id = ? AND state = ?', invoice_id, supplier_id, 'finished').order(:ends)
```
wvengen: I think this could be a bit more Railsy:
```ruby
@deliveries_collection = Delivery.where(invoice_id: [invoice_id, nil], supplier_id: supplier_id).order(:delivered_on)
@orders_collection = Order.where(invoice_id: [invoice_id, nil], supplier_id: supplier_id, state: 'finished').order(:ends)
```

paroga: are you sure this results in the same SQL? i would read it as "'(invoice_id = ? OR invoice_id IS NULL) AND supplier_id = ?'. (see the additional parentheses around the OR). AFIK this kind of OR is not supported in Rails 4 and will come with Rails 5

paroga: ... but since all invoices should have the correct supplier (if nobody hacked the DB) this should result in the same result

paroga: for the deliveries it's no problem, but for the orders, since we still want to show closed orders in the edit page, when they have been assigned before

wvengen: Oh, sorry, I misread! I'd prefer to see brackets and not rely on OR/AND operator precedence.
You're right here!

wvengen:
Thanks :)

PR #423, opened 2016-03-11
Add REST API to retrieve financial transactions

paroga:


wvengen:
Awesome! I've been wanting to add a proper API. I would like to move to active model serializer at some point, this is a good start (Rails 5 has settled on that as well).

wvengen:
If you happened to have started on specifications for other endpoints, please feel free to add something to the wiki.

wvengen:
The only thing I'm missing here, are tests. Most important other things have tests now, and starting off the API without tests I'm afraid that won't get any coverage at all. For the rest, I'd be happy to merge it!

paroga:
Will do so, but I want to upstream the basic "infrastructure" for it first (dorrkeeper).
My first usecase is to create a mobil-app for handling the "offline paper". I'd like to do some prototyping first (maybe in a plugin), get then some feedback and then upstream (and document) it.

+1 for adding tests. I only like to upstream the Api::V1::BaseController for now. Are there any good examples for similar tests?

wvengen:
Excerpt from another project (not using doorkeeper):

```ruby
# spec/controllers/api/v1/user/status_controller_spec.rb
require 'rails_helper'

describe API::V1::User::StatusController do
  let(:default_params) { { api_version: "1.0" } }

  describe "GET to :index" do
    let(:user)  { create(:user) }
    let(:token) { user.authentication_token }

    before { get :index, authentication_token: token, format: "json" }

    it { is_expected.to respond_with :success }

    describe "the returned JSON" do
      subject(:json) { result = JSON.parse(response.body) }

      context "with a valid auth token" do

        its(["logged_in"]) { is_expected.to be true }

        describe "the included user" do
          subject { json["user"] }

          its(["id"]) { is_expected.to eq user.id }
        end
      end
    end
  end
end
```

and a serializer spec:

```ruby
# spec/serializers/user_serializer_spec.rb
require 'rails_helper'

describe UserSerializer do
  let(:user) { create :user }
  let(:json) { UserSerializer.new(user).to_json }
  subject(:parsed_json) { JSON.parse(json) }

  context "serializes a user without oauth_hash" do
    its(["email"]) { is_expected.to eq (user.email) }
    its(["can_preview"]) { is_expected.to be false }
  end
end
```

This uses [rspec-its](https://github.com/rspec/rspec-its).

wvengen:
Regarding testing, perhaps [using JSON schema](https://github.com/rails-api/active_model_serializers/blob/master/docs/howto/test.md) may be even better.

wvengen:
Closing in favour of #429. @paroga would the proposed financial_transactions endpoint and serializer be ok for you?

Issue #17, opened 2009-07-03
Stockit: Hide articles with a stock amount of zero


bennibu:
The default stockit view should hide articles with an amount lower than 1. But, viewing all stockarticles should be possible.

PR #429, opened 2016-05-23
REST API v1

wvengen:
Starting off from #423, this is a first version of an API for Foodsoft. By design, it only supports user-facing resources, so API users can only see and update one's own orders. Perhaps in a later version of the API, we may add admin-functionality as well.

_**note** this PR will be split into smaller PRs, so don't merge this one_

_**note** this feature is finished, but I'd rather wait with merging until [foodsoft-shop](https://github.com/foodcoops/foodsoft-shop) is ready (or someone else wants to use the API asap); before merging we can still change the api, afterwards I'd be more careful_

Pending:
- [x] Fix merge conflict with master
- [x] Make CORS work for multiple scopes (instead of a hardcoded `/f`)
- [x] Make sure relevant errors return json with description
- [x] Use current language in Doorkeeper
- [x] Review security
  - [x] CORS settings
  - [x] Make sure ransack searches don't leak data (like who ordered what)
- [x] Document endpoints (at least a bit)
- [x] Include tests (see also https://github.com/foodcoops/foodsoft/pull/423#issuecomment-195322128)
- [x] Use model name as root key (instead of `data`)
  - [x] and include `order_article` to `group_order_article` update endpoint (and maybe others)
- [x] Allow plugins to register themselves for OAuth2 credentials (use `Doorkeeper::Application`)
- [x] Make Doorkeeper work with multicoops (!)

For later: return all orders, not just open ones (and allow filtering by status)


wvengen:
Decided to not return json for server error (status 500) e.a., so that clients are aware that perhaps at some point a json body isn't present - which might be the case if something breaks really terribly.

wvengen:
The url used with petstore.swagger.io in doc/API.md is already the one for the master branch here. To test it now, use [this link instead](http://petstore.swagger.io/?url=http%3A%2F%2Fcdn.rawgit.com%2Fwvengen%2Ffoodsoft%2Ffeature%2Frest-for-shop%2Fdoc%2Fswagger.v1.yml).

app/controllers/admin/base_controller.rb
```diff
@@ -1,10 +1,10 @@
 class Admin::BaseController < ApplicationController
   before_filter :authenticate_admin
-  
+
   def index
-    @user = self.current_user
+    @user = current_user
```
wvengen: I guess it broke since moving `current_user` to a concern, this makes it work again.

wvengen:
One reason I'm not merging this yet, is because a question came up in foodsoft-shop. Now we use `data` as the main json key (with `meta` for pagination and such). This is all fine, except when updating a `group_order_article`. The update returns the new `group_order_article`, but the related `order_article` is necessarily also updated. Foodsoft-shop now does a second request to update this, but this adds latency: it would be much nicer to embed this in the first response. With the `data` key, there is no natural place for this.

If we'd use the model name as the json key (`group_order_article(s)`), the `order_article` could just be included under its own json key. Does it make sense to change this? (Even though it would need more code on the client-side, because there is no standard place to look for the data.)

What do you think?

carchrae:
i think we should prioritize this PR. even if it doesn't have a complete front end. each time i go to update a rails view or fix bootstrap 2 design stuff it makes me wince.

I've fixed many performance issues, most are due to sloppy rails joins that would be better done by front end code.

I'll try and give this some time in the next week to review. i did a big api project with spree/solidus so I have some experience with building a rails api. we actually built exactly the same design, a really fast client in angular and rails just for data and server logic.

carchrae:
in order to ship, i propose you make a milestone for api and split each item in your checklist into an issue in that milestone.  easier to submit and comment on that way too

carchrae:
in order to ship, i propose you make a milestone for api and split each item in your checklist into an issue in that milestone.  easier to submit and comment on that way too

wvengen:
Sounds good! -> [milestone #17](https://github.com/foodcoops/foodsoft/milestone/17)

It would be good to have a discussion on the mailing-list with regards to technologies. Some people like the current 'just-Ruby' approach where everything is in one place, others like a clearer separation between frontend and backend.

carchrae:
the spree approach was to have both. slowly their ui used more api routes over time. this is a flexible approach though, because you can change routes to just api if you like

wvengen:
I guess it would be good, then, to not have a separate front-end repository (yet), but use the Rails-supplied React setup. That way, at least in the experience of Ruby-people, there is nothing extra to setup.

Tamriel:
Our Foodsoft really wants to use the Webshop, since it has the button 'Ordered by everyone'.
We would love to have this branch merged, because this branch with its old gem versions does not work with our web hoster, but the normal Foodsoft does.

wvengen:
Thanks for expressing your interest. Foodsoft-shop still needs some small improvements before it is fully ready. Did you test it locally and did it work well?

Tamriel:
We did two orders with thousands of articles with an old version, it worked great :-)
I just testet the current version locally:
It does not work with ruby 2.4 because rails version of the API-branch is incompatible with 2.4 (but Foodsoft master fixes this).
Installing with ruby 2.3 works, but I get an error for which I opened a seperate ticket.

wvengen:
Wow, that's good to hear! I'll try to find some time to finish integration then, in the coming months (even if it does not have all the bells and whistles yet to make it a full replacement for all existing functions, so it will need to become a configuration item for a while).

Tamriel:
Great to see you continuing :-)
My Food Coop would like to do an order and could test this.
Just say when it's a good time, and we will do the order with this branch + foodsoft-shop.

paroga:
It would be nice to make small logical PR out of it first, but keep this here for the overview where it's heading (like I do with the bank transaction stuff)

paroga:
do we want to add the foodsoft-shop in this repo too or how do we make sure they stay compatible?

wvengen:
Ok, I'll split it in the following PRs:
* API setup (doorkeeper, splitting off controller concerns, api base controller)
* A couple of endpoints, in logical groups
* Specs and docs
* Foodsoft-shop plugin (see https://github.com/foodcoops/foodsoft-shop/issues/4 and [this branch](https://github.com/wvengen/foodsoft/tree/feature/rest-for-shop-with-plugin))

Even though I've packed foodsoft-shop in a Foodsoft plugin (that I'd like to ship within Foodsoft), they should remain working even across versions, as long as the API version is supported. (Even though this might break subtly if the API consumers expect newly introduced attributes to be always present - but at least older API consumers will work with newer Foodsoft versios).

carchrae:
i want to say think you @wvengen and @paroga -  i would love to switch to an api version.  i have many improvements to contribute back but they are mixed up with customisation for my coop, keeping the custom stuff in the front-end, where possible, would help a lot.  hopefully i get a bit of time to do that in the next month.

wvengen:
Closing if favour of the 5 other API PRs.

PR #450, opened 2017-01-18
Add IBAN field to supplier and user model

paroga:


wvengen:
Way to go!
What about adding a config option? I think some users would be uneasy even seeing something mentioned like IBAN. This would allow foodcoops using this to enable it. Or perhaps to auto-enable when plugins using this functionality are enabled.

wvengen:
Thanks! I haven't tested it, but it looks fine.
note to self: test this when releasing a new version

PR #451, opened 2017-01-20
Show button for creating invoices also at finished orders

paroga:
Check if a order is not opened instead of checking if it is closed.


wvengen:
Sounds sensible :+1:

PR #452, opened 2017-01-20
Add missing German translations

paroga:


wvengen:
Thanks!
We keep locales in [localeapp](http://www.localeapp.com/projects/4867) (see also [Translations](https://github.com/foodcoops/foodsoft/wiki/Translation)), it would be slightly handier to keep them updated in there.

PR #459, opened 2017-02-04
Add SMTP server for handling reply emails

paroga:
Handling incoming mail via SMTP is easier to set up than with the rake task. Provide a rake task which will start a simple SMTP server for receiving emails.


wvengen:
Wow, really cool feature :+1:

wvengen:
One thing I'm missing: documentation on usage and setup. Would you care to add that to the plugin's README?

wvengen:
Super, thanks!

PR #460, opened 2017-02-14
Fix create_upcoming_periodic_tasks after 9ae7c05

paroga:


wvengen:
Hmmm that seems like a bug, thanks for fixing :)

PR #465, opened 2017-03-04
Allow ERB syntax in the app_config.yml

paroga:
This change allows us to refer to envrionment variables in the yml file.
This is helpful if passwords are passed as environment variables in
Docker installations to avoid plain passwords in configuration files.


wvengen:
Sounds like a great idea :+1:

PR #471, opened 2017-04-09
Install all dependencies in an authenticated way (https)

ypid:
git:// is unauthenticated!!!

(My firewall does not allow this :wink: )


wvengen:
Thank you, that's much better.

PR #478, opened 2017-06-28
Fix order by groups document after 020b75c

paroga:
Use a OUTER JOIN instead of a INNER JOIN to include also the stock order
which does not have an entry in the groups table.


wvengen:
Thank you, I needed to look twice :)

PR #482, opened 2017-07-26
Dockerfile for production

paroga:


Dockerfile
```diff
@@ -1,40 +1,22 @@
...
-
-# Add Gemfiles and run bundle.
-COPY Gemfile Gemfile.lock ./
-ENV BUNDLE_JOBS=4 BUNDLE_PATH=/home/app/bundle \
-    BUNDLE_APP_CONFIG=/home/app/bundle/config
-RUN bundle install
-
-# Copy the application code. (Excluded files see .dockerignore)
+WORKDIR /usr/src/app
 COPY . ./
```
wvengen: Don't we need something like [`.dockerignore`](https://docs.docker.com/engine/reference/builder/#dockerignore-file), perhaps?

wvengen:
It would be nice to generate a `REVISION` file (which is used [here](https://github.com/foodcoops/foodsoft/blob/master/config/application.rb#L67)) when building a docker image. But this is a nice to have that can be added later.

Dockerfile
wvengen: I realize we already have it :)

paroga:
I'd like to keep the main `Dockerfile` for production. We can add a Dockerfile.dev if we really want that later too.

Improving the `.dockerignore`sounds like a good idea, but is not mandatory to get it working again. There are additional changes too, which should be done to the `Dockerfile` to reduce the resulting size (unistall the dev-packages after bundler run).

I'll added generating the `REVISION` file.

Dockerfile
```diff
@@ -1,40 +1,23 @@
...
-      xvfb \
-      iceweasel && \
+      libmysqlclient-dev && \
     rm -rf /var/lib/apt/lists/* && \
     apt-get clean
 
-# Run app and all commands as user 'app'. This avoids changing permissions
-# for files in mounted volume.
-RUN adduser --gecos GECOS --disabled-password --shell /bin/bash app
-USER app
```
wvengen: Wait a minute, does removing this mean that Foodsoft now runs as root? (Hope this is not a newbie Docker question.)

wvengen:
Thanks! And looking good. Besides the run-as-question, ready to merge, I'd say.
I think we can still remove mailcatcher from `docker-compose.yml` for production.

Did you manage to re-setup the [docker hub build](https://hub.docker.com/r/foodcoops/foodsoft/)?

Dockerfile
paroga: It's very uncommon to create a special user just for running the main entry point. Since it's a Docker container there are no "real root" rights in the container anyway and having an additional user does not offer a lot of additional security to the underlying system.

paroga:
Updating the `docker-compose.yml` file is a completely different task, since it's for development only. I personally prefere to get rid of it anyway and provide an example in the Wiki pages instead to show an example Docker setup for production. I also like to remove the `doc` folder and put that information in the wiki too. What do you think?

Dockerfile
wvengen: Makes sense. Thanks for explaining!

wvengen:
Great :+1:

I know @dan-nl is working on a separate development and production docker setup, let's rework the rest with that.
Regarding the `doc` folder, I think it would be a good idea to have instructions on how to get it running in the source tree, so that it's there with the source code. Also, instructions may change over versions, keeping that in VCS allows one to have the right version (e.g. dependencies).

wvengen:
One thing I didn't see yet: how about asset precompilation?

PR #483, opened 2017-08-03
Use a temporary mysql-server for assets precompilation during Docker …

paroga:
…build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,22 @@ RUN buildDeps='libmagic-dev' && \
     apt-get purge -y --auto-remove $buildDeps && \
     whenever --update-crontab
 
+# Add a temporary mysql-server for assets precompilation
+RUN export DATABASE_URL=mysql2://localhost/temp && \
+    export DEBIAN_FRONTEND=noninteractive && \
+    apt-get update && \
+    apt-get install -y mysql-server && \
+    /etc/init.d/mysql start && \
+    cp config/app_config.yml.SAMPLE config/app_config.yml && \
+    bundle exec rake db:setup && \
+    bundle exec rake assets:precompile && \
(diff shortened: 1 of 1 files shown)


wvengen:
Just came across [multi-stage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/#use-multi-stage-builds), would that be an option for asset precompilation and bundle installation? First stage could install all dependencies needed for building and precompilation, second step would setup the final image and just copy the app directory to the new image. Just need to make sure to have a shared base system layer in the two build stages.

paroga:
Yes, I'm aware of that, but AFAIK Docker Hub does not support this feature ATM. Only Docker Cloud does.

wvengen:
The build is failing, though. I suspect Travis CI problems, but it would be good to mention that you've seen the build failing and why it is acceptable to merge anyway (!)

paroga:
My fault, sty. I didn't look at this build result, just at the build results on hub.docker.comm, since it only affects the Docker build. But it seams that Travis is doing something with the Docker stuff. It also fails if I add the spec folder to the `.dockerignore` file. Do you know more how Travis is configured ATM?

wvengen:
Weird, haven't seen this before. It looks like either the database creation command failed, or Travis changed the setup? [mysql on Travis docs](https://docs.travis-ci.com/user/database-setup/#MySQL)

paroga:
Travis has now problems to build the same revision, which worked already before:
https://travis-ci.org/foodcoops/foodsoft/builds/260397329
https://travis-ci.org/foodcoops/foodsoft/builds/260541712

wvengen:
It looks like something changed at Travis CI. Fixed it in af1243ebeae534722b69e4f537e096f4ebafac63.

PR #484, opened 2017-08-03
Update Dockerfile to support assets precompilation and cron

paroga:


Dockerfile
```diff
@@ -17,7 +18,26 @@ RUN buildDeps='libmagic-dev' && \
     rm -rf /var/lib/apt/lists/* && \
     bundle install --deployment --without development test && \
     apt-get purge -y --auto-remove $buildDeps && \
-    whenever --update-crontab
+    mkdir -p log && \
+    ln -sfn /dev/stdout log/production.log && \
+    bundle exec whenever --update-crontab && \
+    { echo BUNDLE_APP_CONFIG=$BUNDLE_APP_CONFIG; crontab -l; } | crontab -
```
wvengen: I don't quite understand this line.

config/environments/production.rb
```diff
@@ -78,9 +78,25 @@
...
+    # Use sendmail as default to avoid ssl cert problems
+    config.action_mailer.delivery_method = :sendmail
+  end
 
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
+
+  if ENV['SECRET_KEY_BASE'].present?
+    config.secret_key_base = ENV['SECRET_KEY_BASE']
+  end
```
wvengen: Nice ... this is indeed needed. What about moving this to `environment.rb`? I think this would make sense for all environments.

Dockerfile
```diff
@@ -5,7 +5,8 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/* && \
     apt-get clean
 
-ENV RAILS_ENV=production
+ENV RAILS_ENV=production \
+    RAILS_SERVE_STATIC_FILES=true
```
wvengen: Good one :o

config/environments/production.rb
paroga: How do i write that in `environment.rb`? There is not `Foodsoft::Application.configure` in that file.

Dockerfile
paroga: It prints `BUNDLE_APP_CONFIG=/usr/local/bundle` and the current crontab via `-l` to `STDOUT` which gets piped into `crontab -`, which reads from `STDIN` and updates the crontab.

Dockerfile
```diff
@@ -17,7 +18,26 @@ RUN buildDeps='libmagic-dev' && \
     rm -rf /var/lib/apt/lists/* && \
     bundle install --deployment --without development test && \
     apt-get purge -y --auto-remove $buildDeps && \
-    whenever --update-crontab
+    echo "Foodsoft::Application.config.secret_token = ENV['SECRET_KEY_BASE']" > config/initializers/secret_token.rb && \
```
wvengen: Can we do this in the source repository and use `.env`? It's good practice anyway, and makes deployment with other methods easier too.

Dockerfile
wvengen: Thanks. Wouldn't this better be placed in something like `/etc/profile.d` or so?

Dockerfile
paroga: I don't want to break any existing deployment. If we really want to change it (with all the documentation updates) we should switch to [secrets.yml](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#config-secrets-yml), which is more work than just getting the Docker setup running...

config/environments/production.rb
wvengen: Sorry, I meant `config/application.rb`, or perhaps better, the initializer.

wvengen: Sorry, I meant `config/application.rb`.
Some things like `SECRET_KEY_BASE` are probably better off in an initializer.

Dockerfile
wvengen: Ok, good for now.

config/environments/production.rb
paroga: i changed this already anyway...

docker-entrypoint.sh
```diff
@@ -1,6 +1,19 @@
...
+
+if test -n "$BUNDLE_APP_CONFIG"; then
+  CRONTAB_ENV="${CRONTAB_ENV}BUNDLE_APP_CONFIG=$BUNDLE_APP_CONFIG\n"
+fi
+
+if test -n "$DATABASE_URL"; then
+  CRONTAB_ENV="${CRONTAB_ENV}DATABASE_URL=$DATABASE_URL\n"
+fi
+
+echo "$CRONTAB_ENV$CRONTAB" | crontab -
```
wvengen: Does this get run every time the container starts?
Why not do this as part of the Docker build step?

paroga: you do not have the required environment variables available during build :-(

wvengen: Ah, of course. That's a bummer.
Does this end up making the crontab one line longer each time the instance is run (e.g. reboot)?

paroga: yes. we can reduce it to be done only when the arguments contain the string `cron`, but this is some quirks and i do not see a problem in updating the crontab during every start

wvengen: Ok!

PR #489, opened 2017-08-17
Enhanced matrix pdf

FcFlodo:
- sort articles by description for easier localization
- rotate header for enhanced readability and longer descriptions
- add net price for easier check with supplier invoice
- add unit to header for quantity-clarification

A bug in the prawn gem forced me to update the gem
https://github.com/prawnpdf/prawn/issues/409
https://github.com/prawnpdf/prawn-table/pull/32

I have uploaded the Gemfile.lock which has changed, but I am not sure if this has to be changed?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,7 @@ gem 'rails-i18n'
 
 gem 'mysql2'
 gem 'prawn'
-gem 'prawn-table'
+gem 'prawn-table', :git => 'https://github.com/straydogstudio/prawn-table.git', ref: '759a27b6'
 gem 'haml-rails'
 gem 'kaminari'
 gem 'simple_form'
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,7 @@ gem 'rails-i18n'
 
 gem 'mysql2'
 gem 'prawn'
-gem 'prawn-table', :git => 'https://github.com/straydogstudio/prawn-table.git', ref: '759a27b6'
+gem 'prawn-table'
 gem 'haml-rails'
 gem 'kaminari'
 gem 'simple_form'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,13 +5,6 @@ GIT
     localize_input (0.1.0)
 
 GIT
-  remote: https://github.com/straydogstudio/prawn-table.git
-  revision: 759a27b6fef2e682c035fc83f2182e53c3270279
-  ref: 759a27b6
-  specs:
-    prawn-table (0.2.1)
-
-GIT
   remote: https://github.com/technoweenie/acts_as_versioned.git
   revision: 63b1fc8529d028fae632fe80ec0cb25df56cd76b
   specs:
@@ -262,6 +255,8 @@ GEM
     prawn (2.2.2)
       pdf-core (~> 0.7.0)
       ttfunk (~> 1.5)
+    prawn-table (0.2.2)
+      prawn (>= 1.3.0, < 3.0.0)
     protected_attributes (1.1.0)
       activemodel (>= 4.0.1, < 5.0)
     pry (0.10.3)
@@ -517,7 +512,7 @@ DEPENDENCIES
   meta_request
   mysql2
   prawn
-  prawn-table!
+  prawn-table
   protected_attributes (= 1.1.0)
   pry-rescue
   pry-stack_explorer
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,6 +5,13 @@ GIT
     localize_input (0.1.0)
 
 GIT
+  remote: https://github.com/straydogstudio/prawn-table.git
+  revision: 759a27b6fef2e682c035fc83f2182e53c3270279
+  ref: 759a27b6
+  specs:
+    prawn-table (0.2.1)
+
+GIT
   remote: https://github.com/technoweenie/acts_as_versioned.git
   revision: 63b1fc8529d028fae632fe80ec0cb25df56cd76b
   specs:
@@ -217,7 +224,7 @@ GEM
     libv8 (3.16.14.17)
     loofah (2.0.3)
       nokogiri (>= 1.5.9)
-    mail (2.6.6)
+    mail (2.6.4)
       mime-types (>= 1.16, < 4)
     mailcatcher (0.6.4)
       activesupport (~> 4.0)
@@ -237,7 +244,7 @@ GEM
       mime-types-data (~> 3.2015)
     mime-types-data (3.2016.0521)
     mini-smtp-server (0.0.2)
-    mini_portile2 (2.2.0)
+    mini_portile2 (2.1.0)
     minitest (5.10.1)
     mono_logger (1.1.0)
     multi_json (1.12.1)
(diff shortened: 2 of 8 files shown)


config/locales/en.yml
```diff
@@ -734,6 +734,7 @@ en:
       - Article
       - Unit
       - Unit quantity
+      - Netto price
```
paroga: if you compare it to the other translations "Nettopreis" in german is usually "Price" in english

please also update the other languages. the translation should be easy to find

Gemfile.lock
```diff
@@ -217,7 +224,7 @@ GEM
     libv8 (3.16.14.17)
     loofah (2.0.3)
       nokogiri (>= 1.5.9)
-    mail (2.6.6)
+    mail (2.6.4)
```
paroga: are these version changes to a lower version really necessary? i don't think so

app/documents/order_matrix.rb
```diff
@@ -13,22 +13,26 @@ def title
...
-    order_articles = @order.order_articles.ordered
+    order_articles = @order.order_articles.sort_by{|a| a.article.name.downcase}
 
     text I18n.t('documents.order_matrix.heading'), style: :bold
     move_down 5
     text I18n.t('documents.order_matrix.total', :count => order_articles.size), size: fontsize(8)
     move_down 10
 
     order_articles_data = [I18n.t('documents.order_matrix.rows')]
+
```
paroga: please avoid unnecessary whitespace changes

app/documents/order_matrix.rb
```diff
@@ -13,22 +13,26 @@ def title
...
     order_articles.each do |a|
       order_articles_data << [a.article.name,
                               a.article.unit,
                               a.price.unit_quantity,
+                              number_to_currency(a.price.price * a.price.unit_quantity),
                               number_with_precision(article_price(a), precision: 2),
                               a.units]
     end
+    
+    #order_articles_data.sort_by!{|a| a[0].downcase}
```
paroga: please avoid useless comments

app/documents/order_matrix.rb
```diff
@@ -51,12 +55,14 @@ def body
 
       # Make order_articles header
       header = [""]
+      
       for header_article in current_order_articles
         name = header_article.article.name.gsub(/[-\/]/, " ").gsub(".", ". ")
-        name = name.split.collect { |w| w.truncate(8) }.join(" ")
-        header << name.truncate(30)
+        name = name.split.collect { |w| w.truncate(20) }.join(" ")
+        header << name.truncate(35)+' - ('+header_article.article.unit+')'
```
paroga: `header << "#{name.truncate(35)} - (#{header_article.article.unit})"` might be easier to read

app/documents/order_matrix.rb
```diff
@@ -13,22 +13,26 @@ def title
   end
 
   def body
-    order_articles = @order.order_articles.ordered
+    order_articles = @order.order_articles.sort_by{|a| a.article.name.downcase}
```
paroga: can we do this sorting via ActiveRecord instead?

app/documents/order_matrix.rb
```diff
@@ -74,13 +80,22 @@ def body
...
+      table.rows(0).padding = 2
+      table.rows(0).height = 90
+      table.rows(0).overflow = :truncate 
+      table.rows(0).align = :left
+      table.rows(0).min_font_size = 6
+     # table.rows(0).valign = :bottom
+        
+      table.cells.border_width = 1
+      table.cells.border_color = '666666'
+      table.row_colors = ['ffffff','ececec']
```
paroga: you removed the indentation for this block

Gemfile
```diff
@@ -22,7 +22,7 @@ gem 'rails-i18n'
 
 gem 'mysql2'
 gem 'prawn'
-gem 'prawn-table'
+gem 'prawn-table', :git => 'https://github.com/straydogstudio/prawn-table.git', ref: '759a27b6'
```
paroga: this revision is from 2014. the latest upstream version is from 2016. do you see any chance that the commit from straydogstudio gets accepted in upstream? I do not like the idea of using such an outdated version. If the merge in upstream is not possible: Could you rebase the commit from straydogstudio to upstream master, publish them in a new repository and use that in the Gemfile?

Gemfile.lock
FcFlodo: No they are not necessary. The Gemfile.lock changed automatically and I was not sure if I should upload it.

wvengen:
Hi @FcFlodo, thanks for your contribution! Do you see any opportunity to finish it?

FcFlodo:
Hi @wvengen and @paroga !
Sorry for the long delay, I couldn´t spare much time in the past few weeks for this project.

I tried to get my code running with the new prawn-version, but until now I haven´t found a way to solve this.

wvengen:
Hi @FcFlodo, no problem. What about removing the header rotation for now, so that the other improvements can be included? If then an solution is found for the issue, the rotated headers (which I can imagine would be much much nicer) can be included.

I also suggest to revive the discussion upstream. If there are others interested in the feature, it may get some traction. If not, then you could try to fix it upstream and get it included. Putting it into a separate repository would not have my preference, because that basically means that we'd need to maintain it.

app/documents/order_matrix.rb
FcFlodo: I did the sorting still with sort_by, because I was afraid, that I make unwanted changes in the database with ActiveRecord sorting. Feel free to change this if this is a problem.

FcFlodo:
Finished my contribution without the rotated headers. I will ask if the changes made by straydogstudio in the prawn-table gem could be implemented.

Gemfile.lock
wvengen: If you would like to remove these changes, that would clean it up a bit (and avoid a conflict).
Or do you need a newer prawn? If you do, please rebase on current `master`.

app/documents/order_matrix.rb
wvengen: Sorting belongs in the database, this is a memory hog and doesn't scale for large orders. Please reconsider the `ordered` scope if you think a different sort order is useful everywhere; if it only makes sense here, don't use the `.ordered` scope but use active-record sorting. If you'd like case-insensitive sorting, indexes may also be needed.

Sorry for the condensed answer, if you need more explanation, please don't hesitate to ask. We'll do our best to help!

paroga:
@FcFlodo: I did a bigger rewrite of the OrderPdf classes in #532. Does it solve your issues too?

carermey:
hello.

we are using the foodsoft for our big wholesaler ordering! this extension:

>  sort articles by description for easier localization
>   rotate header for enhanced readability and longer descriptions
>   add net price for easier check with supplier invoice
>   add unit to header for quantity-clarification

would be absolute great !

If the articles would be ordert first by categorie and afterwards by produktname in alphabetical order would make the overview also much better ! 

Thank you a lot for your work on it !

Carermey

yksflip:
Hey,
Thanks again for your contribution!

PR #490, opened 2017-08-19
Improve mailing

paroga:
the latest commits are not ready to be commit, but i'd like to get some early feedback. it would be great if you could do the review on a commit basis and tell me which of them are already ok for you

### DIFF
--- a/app/mailers/mailer.rb
+++ b/app/mailers/mailer.rb
@@ -89,6 +89,12 @@ class Mailer < ActionMailer::Base
       args[k] = "#{show_user user} <#{user.email}>" if user.is_a? User
     end
 
+    reply_email_domain = FoodsoftConfig[:reply_email_domain]
+    if reply_email_domain && !args[:return_path]
+      address = Mail::Parsers::AddressListsParser.new.parse(args[:to]).addresses.first
+      args[:return_path] = "<#{FoodsoftConfig.scope}.bounce+#{address.local}=#{address.domain}@#{reply_email_domain}>"
+    end
+
     super
   end
 
--- a/app/mailers/mailer.rb
+++ b/app/mailers/mailer.rb
@@ -102,7 +102,7 @@ class Mailer < ActionMailer::Base
     message = yield
     message.deliver_now
   rescue => error
-    Rails.logger.warn "Can't deliver mail to #{message.to[0]}: #{error.message}"
+    MailDeliveryStatus.create email: message.to[0], message: error.message
   end
 
 end
--- a/app/controllers/admin/mail_delivery_status_controller.rb
+++ b/app/controllers/admin/mail_delivery_status_controller.rb
@@ -1,0 +1,30 @@
+class Admin::MailDeliveryStatusController < Admin::BaseController
+  inherit_resources
+
+  def index
+    @maildeliverystatus = MailDeliveryStatus.order(created_at: :desc)
+    @maildeliverystatus = @maildeliverystatus.where(email: params[:email]) unless params[:email].blank?
+    @maildeliverystatus = @maildeliverystatus.page(params[:page]).per(@per_page)
+  end
+
+  def show
+    @maildeliverystatus = MailDeliveryStatus.find(params[:id])
+    filename = "maildeliverystatus_#{params[:id]}.#{MIME::Types[@maildeliverystatus.attachment_mime].first.preferred_extension}"
+    send_data(@maildeliverystatus.attachment_data, :filename => filename, :type => @maildeliverystatus.attachment_mime)
+  end
+
+  def destroy_all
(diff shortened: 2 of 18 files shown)


paroga:
@wvengen ping?

lib/foodsoft_mail_receiver.rb
```diff
@@ -0,0 +1,34 @@
+require 'mail'
+require 'mini-smtp-server'
```
wvengen: I had trouble to `mini-smtp-server` as explained in https://github.com/foodcoops/sharedlists/issues/3#issuecomment-328119403.

plugins/messages/app/views/messages/new.haml
```diff
@@ -10,15 +10,51 @@
...
+        $('#order').slideDown();
+        $('#recipients').slideDown();
+      });
+
+      $('#message_send_method_recipients').on('click', function() {
+        $('#workgroup').slideUp();
+        $('#messagegroup').slideUp();
+        $('#order').slideUp();
+        $('#recipients').slideDown();
+      });
```
wvengen: Yay, this is nice functionality!

plugins/messages/lib/foodsoft_messages/mail_receiver.rb
```diff
@@ -0,0 +1,3 @@
+ActiveSupport.on_load(:after_initialize) do
+  FoodsoftMailReceiver.register MessagesMailReceiver
+end
```
wvengen: Great way.

paroga:
can you review this PR as it is? or should i split it put into smaller PRs? the PR is now nearly 6 weeks old 😞

app/mailers/mailer.rb
```diff
@@ -78,6 +78,11 @@ def not_enough_users_assigned(task, user)
   def mail(args)
     args[:subject] = "[#{FoodsoftConfig[:name]}] #{args[:subject]}"
 
+    if args[:from].is_a? User
+      args[:reply_to] = args[:from] unless args[:reply_to]
+      args[:from] = "#{show_user args[:from]} via #{I18n.t('layouts.foodsoft')} <#{FoodsoftConfig[:email_sender]}>"
```
wvengen: i18n!

wvengen: (hmm now I see it was like this already, feel free to add or keep)

paroga: I'll address it in a separate issue then the locale is out of the plugin directory...)

wvengen:
Sorry, I missed the bit of reviewing separate commits.
I'd merge all the small bits upto and including 2264351cf58fb8f1bc340f9de09636b03170be16 _Remove code duplication_ - cleans up it quite a bit! :)
I think a3716fcb61bcb310efe8b80595500ba102e0987b _SMTP plugin to core_ is a great step, one comment about dependency.
3585e33385265a06a20e8e5be98802d4bf2dd101 _Improve usability_ go for it

Regarding the other commit, the bounce handling, I would consider putting it carefully into the users screen. In show user we can explain that there is a bounce problem and provide buttons for re-enabling message sending if auto-blocked, add some icon in the user list if there is a bounce issue, allow filtering users by bounce issue (as part of user filtering which might not exist yet), and perhaps a notice box on top with a link to a the list of (recently) bounced addresses if there are issues to look at. These are just ideas of how to integrate.
If this is a lot of work or would block finishing this, I think the current is already a step forwards.

paroga:
ad bounce handling: the current main issue i have to cope with are "unmaintained" user lists: foodcoops have old/invalid users, with invalid email addresses. showing the bounce in the user screen does not help here, since the user does not login

wvengen:
Ah, I mean the admin users list (sorry for the confusion).

paroga:
what do you think about storing the user (when available) in the `MailDeliveryStatus` database row and show some information in the user list, if there are any errors for that user? some errors do not necessary have an user

wvengen:
I think that would be a good idea, I like what you have now.
I can see cases where there is no associated user, like email change or user deletion, but in these cases the offending addresses won't get new mails, so I don't think they're really relevant anymore.
Do we need an alert on top of the users screen (linking to the bounces index) if there are bounces?

paroga:
I'd skip the user screen alert for now, since it would require more logic. When should we show it? As long as there are MailDeliveryStatus rows? How can a user hide the message? IMHO it would be a nice feature, but it would require a lot more effort. This additional logic should then also contain handling for disabling sending mails to "invalid" addresses.

wvengen:
That sounds reasonable, it is fine as it is.
We might also consider adding a top right button from the users admin screen for viewing bounces (to avoid cluttering the admin menu), so that the info is accessible.

PR #494, opened 2017-08-20
Notify users also about tasks which are not on sundays

paroga:
notify_users_of_weekly_task notifys workgroups about tasks with and due
date of 7 days in the future. Call it every day to notify useres of all
tasks.


wvengen:
Hmm as I read it, it looks like a bug!
Just to be sure: this doesn't mean users will get an email every day?

PR #495, opened 2017-08-20
Add option to automatically finish an order

paroga:
This change allows us to implement automatic notifications
to the supplier in the next step.

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -117,8 +117,7 @@ class OrdersController < ApplicationController
   # Send a order to the supplier.
   def send_result_to_supplier
     order = Order.find(params[:id])
-    Mailer.order_result_supplier(@current_user, order).deliver_now
-    order.update!(last_sent_mail: Time.now)
+    order.send_to_supplier(@current_user)
     redirect_to order, notice: I18n.t('orders.send_to_supplier.notice')
   rescue => error
     redirect_to order, alert: I18n.t('errors.general_msg', :msg => error.message)
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -16,6 +16,8 @@ class Order < ActiveRecord::Base
   belongs_to :updated_by, :class_name => 'User', :foreign_key => 'updated_by_user_id'
   belongs_to :created_by, :class_name => 'User', :foreign_key => 'created_by_user_id'
 
+  enum end_action: { no_end_action: 0, auto_close: 1, auto_close_and_send: 2, auto_close_and_send_min_quantity: 3 }
+
   # Validations
   validates_presence_of :starts
   validate :starts_before_ends, :include_articles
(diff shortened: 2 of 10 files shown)


wvengen:
Looking good!
What is your use-case for this?

This idea has come up for me in the past several times, and I decided not to implement it, because there was always a manual step after the order was closed (sending the order to the supplier). So why bother closing automatically when it really depends on the next manual step? That gives people just a bit more time to order until it is _really_ time. Having automated mails would make this more useful (but many foodcoops want to add manual notes to it about pickup arrangement etc.).

Anyway, if there's use, let's include it. I would be careful to add many checkboxes for options like this, because it requires making more decisions in the normal flow. Anyway, I think it's fine now.

wvengen:
Ah, this ties in to #488.

wvengen:
That sounds really good!

paroga:
i've adopted this change

app/models/order.rb
```diff
@@ -263,6 +265,35 @@ def close_direct!(user)
...
+  def do_end_action
+    if auto_close?
+      finish! created_by
+    elsif auto_close_and_send?
+      finish! created_by
+      send_to_supplier created_by
+    elsif auto_close_and_send_min_quantity?
+      finish! created_by
+      send_to_supplier created_by if order.sum >= order.supplier.min_order_quantity
+    end
```
wvengen: In many other places we use `finish!(user)`, maybe here as well? (Also `send_to_supplier`.)

app/models/order.rb
```diff
@@ -263,6 +265,35 @@ def close_direct!(user)
     update_attributes! state: 'closed', updated_by: user
   end
 
+  def send_to_supplier(user)
+    Mailer.order_result_supplier(user, self).deliver_now
+    update!(last_sent_mail: Time.now)
+  end
+
+  def do_end_action
```
wvengen: Does this warrant an exclamation mark? (Since it calls `finish!`, which changes the order).

app/models/order.rb
```diff
@@ -263,6 +265,35 @@ def close_direct!(user)
     update_attributes! state: 'closed', updated_by: user
   end
 
+  def send_to_supplier(user)
```
wvengen: Does this warrant an exclamation mark? (Since it sends a mail and calls `update!`, which changes the order).

app/models/order.rb
```diff
@@ -263,6 +265,35 @@ def close_direct!(user)
...
+    elsif auto_close_and_send?
+      finish! created_by
+      send_to_supplier created_by
+    elsif auto_close_and_send_min_quantity?
+      finish! created_by
+      send_to_supplier created_by if order.sum >= order.supplier.min_order_quantity
+    end
+  end
+
+  def self.finish_ended
```
wvengen: same here

config/schedule.rb
```diff
@@ -11,3 +11,8 @@
 every :sunday, :at => '7:14 am' do
   rake "multicoops:run TASK=foodsoft:create_upcoming_periodic_tasks"
 end
+
+# Finish ended orders
+every 1.minute do
+  rake "multicoops:run TASK=foodsoft:finish_ended_orders"
+end
```
wvengen: Could it be that this task takes more than a minute to run? Would this interfere? Just to check.

paroga: it's very unlikely that it takes so long (but i'll address this problem with the new time zone aware scheduler anyway later). it it really takes that long finish! will fail, because the order is already finished and nothing happens

app/controllers/orders_controller.rb
```diff
@@ -117,8 +117,7 @@ def finish
   # Send a order to the supplier.
   def send_result_to_supplier
     order = Order.find(params[:id])
-    Mailer.order_result_supplier(@current_user, order).deliver_now
-    order.update!(last_sent_mail: Time.now)
+    order.send_to_supplier(@current_user)
```
wvengen: don't forget to change this!

PR #497, opened 2017-09-09
Improve Docker setup

wvengen:


Dockerfile
```diff
@@ -1,45 +1,60 @@
 FROM ruby:2.3
 
-RUN apt-get update && \
-    apt-get install --no-install-recommends -y cron && \
-    rm -rf /var/lib/apt/lists/* && \
-    apt-get clean
+RUN SUPERCRONIC_URL=https://github.com/aptible/supercronic/releases/download/v0.1.3/supercronic-linux-amd64 && \
```
paroga: You usually put version numner (0.1.3) and sha1 into Docker ENV variables, which makes them more readable.

Dockerfile
```diff
@@ -1,45 +1,60 @@
...
 
 WORKDIR /usr/src/app
-COPY . ./
-RUN echo $SOURCE_COMMIT > REVISION
 
-RUN buildDeps='libmagic-dev' && \
+# separately from rest of the code to avoid re-running bundler when Gemfile doesn't change
+# plugins is copied completely, because COPY can't pick wildcard directories
+COPY plugins plugins/
+COPY Gemfile Gemfile.lock ./
```
paroga: why? we create a image for RELEASE, so this does not apply here.

Dockerfile
```diff
@@ -1,45 +1,60 @@
...
+
+RUN BUILD_DEPS='libmagic-dev' && \
```
paroga: in most Dockerfiles it's written in camelCase

Dockerfile
```diff
@@ -1,45 +1,60 @@
...
-    apt-get purge -y --auto-remove $buildDeps && \
-    echo "Foodsoft::Application.config.secret_token = ENV['SECRET_KEY_BASE']" > config/initializers/secret_token.rb && \
-    mkdir -p log && \
-    ln -sfn /dev/stdout log/production.log && \
-    bundle exec whenever --update-crontab
-
-# Add a temporary mysql-server for assets precompilation
+    apt-get install --no-install-recommends -y $BUILD_DEPS && \
+    echo 'gem: --no-document' >> ~/.gemrc && \
+    bundle config --global build.nokogiri "--use-system-libraries" && \
```
paroga: is `--global` really necessary?

Dockerfile
```diff
@@ -1,45 +1,60 @@
...
-    bundle install --deployment --without development test && \
-    apt-get purge -y --auto-remove $buildDeps && \
-    echo "Foodsoft::Application.config.secret_token = ENV['SECRET_KEY_BASE']" > config/initializers/secret_token.rb && \
-    mkdir -p log && \
-    ln -sfn /dev/stdout log/production.log && \
-    bundle exec whenever --update-crontab
-
-# Add a temporary mysql-server for assets precompilation
+    apt-get install --no-install-recommends -y $BUILD_DEPS && \
+    echo 'gem: --no-document' >> ~/.gemrc && \
```
paroga: I agree with avoiding installation of documentation, but is there no better way to configure that? this looks very ugly to me

Dockerfile
```diff
@@ -1,45 +1,60 @@
...
+    bundle exec rake db:setup assets:precompile && \
+    rm -Rf tmp/* && \
     /etc/init.d/mysql stop && \
-    rm -rf /run/mysqld /tmp/* /var/lib/mysql /var/log/mysql* && \
+    rm -Rf /run/mysqld /tmp/* /var/lib/mysql /var/log/mysql* && \
     apt-get purge -y --auto-remove mysql-server && \
-    rm -rf /var/lib/apt/lists/*
+    rm -Rf /var/lib/apt/lists/* /var/cache/apt/*
+
+RUN bundle exec whenever >crontab
```
paroga: I'd keep the call to whenever int the previous `RUN`

script/start
```diff
@@ -0,0 +1,21 @@
...
+
+PROCFILE=`dirname "$0"`/../Procfile
+CMD=`cat "$PROCFILE" | grep "^$1:" | cut -d: -f2-`
+if [ ! "$CMD" ]; then
+  echo "Process name $1 not found in Procfile" 1>&2
+  exit 2
+fi
+
+PS="$1"
+[ "$PORT" ] || PORT=3000
```
paroga: do not set environment variibales in different places. `Dockerfile` should be the only place for that.

script/start
```diff
@@ -0,0 +1,21 @@
...
+export PS PORT
+
+exec /bin/sh -c "$CMD"
```
paroga: why do we need a new shell for that?

Dockerfile
```diff
@@ -1,45 +1,60 @@
...
+
+# Run app as unprivileged user
+USER nobody
 
 EXPOSE 3000
 
-ENTRYPOINT ["./docker-entrypoint.sh"]
-CMD ["rails", "server", "--binding", "0.0.0.0"]
+# start web process from Procfile
+CMD ["script/start", "web"]
```
paroga: i personally do not like the idea with the `Procfile`. If we really want to provide provide an entry file (which makes sense to some extend), I'd prefer to have everything in ONE shell script.

Dockerfile
wvengen: They are only really needed in this RUN step. I'd consider it litter to leave these variables also set in the final image (e.g. when requesting an interactive shell). Or is that better?

Dockerfile
wvengen: When developing the `Dockerfile` I found it saved some time.

Dockerfile
wvengen: Found this in other `Dockerfile`s :/ Will look a bit.

Dockerfile
wvengen: Logically it's a different topic, isn't this clearer when reading?
Comments within a single `RUN` don't really work, do they?

Dockerfile
wvengen: `Procfile` is something many people already know from foreman / Heroku, I though that would be a nice de facto standard to use. Did not find an existing tool to run just a single command from it, hence the small shellscript `script/start` (which is nonstandard, but simple enough).

script/start
wvengen: This is meant to be a small standalone tool resembling [foreman](https://ddollar.github.io/foreman/) but running only a single process (even though foreman uses port 5000 by default). Maybe I should explain that a bit more in the comments. Since `$PORT` is almost always used in the `Procfile`, one needs to have it defined.
I'd love to use an existing tool for this, but didn't find it.

script/start
wvengen: Because `$CMD` often contains shell variables (like `$PORT`).

Dockerfile
wvengen: Ideally I'd like to run this on Heroku by default as well.

wvengen:
Thanks for the review so quick! Hope you can help me to learn more about the why-s :)

Dockerfile
wvengen: Looked a bit in the gem/bundle docs, did not find a different way. See [gem environment](http://guides.rubygems.org/command-reference/#gem-environment) and [bundle install](https://bundler.io/v1.15/man/bundle-install.1.html). The former has no setting options, the latter doesn't mention changing gem environment settings.
I think it isn't too bad to set a configuration option like this? Will remove the file afterwards.

wvengen:
I think what's left is resolving the `Procfile` issue. What do you suggest?

script/start
paroga: can we put a `ENV PORT=3000` into the `Dockerfile` instead?
What's the purpose of `PS`?

Dockerfile
paroga: for me the "topic" is "setting up the environment", which includes the dependencies and cron jobs

Dockerfile
```diff
@@ -1,45 +1,57 @@
...
     apt-get purge -y --auto-remove $buildDeps && \
-    echo "Foodsoft::Application.config.secret_token = ENV['SECRET_KEY_BASE']" > config/initializers/secret_token.rb && \
-    mkdir -p log && \
-    ln -sfn /dev/stdout log/production.log && \
-    bundle exec whenever --update-crontab
+    rm -Rf /var/lib/apt/lists/* /var/cache/apt/* ~/.gemrc ~/.bundle
+
+# use the default config (or else the app won't start)
+# most non-system-related settings can be changed in the app anyway
+RUN mv config/app_config.yml.SAMPLE config/app_config.yml
```
paroga: i do not like to have a `app_config.yml` in the image. That's a "mandatory configuration" which must be supplied by the user, which can be done with a two line `Dockerfile` and I expect that the application won't start if the config is missing. We could add a nice error message to improve usability.

script/start
```diff
@@ -0,0 +1,27 @@
...
+# `type` is looked up in `Procfile` and the command is started. This allows e.g.
+# docker-compose to run multiple process/dyno types from a single image, without
+# needing to know the exact command (which may change over time).
+#
+if [ ! "$1" ]; then
+  echo "Usage: $0 <process type>" 1>&2
+  exit 1
+fi
+
+PROCFILE=`dirname "$0"`/../Procfile
```
paroga: Could you put the start script into the same folder as the Procfile? I do not like this kind of relative file paths and I also do not see much need for the `dirname`, which just makes the code more difficult to read.

wvengen: This is ugly, I agree. It should be just `Procfile` (or overridable with a `-f` option like [foreman](https://ddollar.github.io/foreman/)). It is assumed anyway that the script is run from the app's root, so it can still bein script/. Or do you still prefer it in the root?

script/start
wvengen: I think it would be useful to be able to run this script also outside a docker container (if you install Foodsoft in a different way). For this case, there is a default port. But I have nothing against giving an error if PORT is not set, would that be ok?

wvengen: `PS` is the worker/dyno type, it can be used in the `Procfile` if one wants.

script/start
paroga: i'd put both files into root, but it's not a real issue for me

script/start
paroga: i don't see any usage of it ATM. i'd keep the script as simple as possible, without any unused feature. IMHO it should only do the minimal stuff to get OUR Procfile running and the rest is up to any other service provider

Dockerfile
wvengen: I'll add an error message. Could this also be supplied with a docker volume?
Needs to end up in the docker docs as well.

script/start
wvengen: Ok, I'll leave out `PS` and show an error message if `PORT` is unset.

Dockerfile
paroga: if you provide an ABSOLUTE path for it on the host side it should work.

wvengen: Thanks, I'll experiment with it and mention it in the docs.

Dockerfile
```diff
@@ -24,25 +26,22 @@ RUN buildDeps='libmagic-dev' && \
...
     /etc/init.d/mysql stop && \
-    rm -Rf /run/mysqld /tmp/* /var/lib/mysql /var/log/mysql* && \
+    rm -Rf /run/mysqld /tmp/* /var/tmp/* /var/lib/mysql /var/log/mysql* && \
     apt-get purge -y --auto-remove mysql-server && \
-    rm -Rf /var/lib/apt/lists/* /var/cache/apt/*
-
-RUN bundle exec whenever >crontab
+    rm -Rf /var/lib/apt/lists/* /var/cache/apt/* && \
+    \
+    bundle exec whenever >crontab
```
paroga: i'd move the crontab to the previous `RUN`, so only the temoprary mysql stuff is in its own `RUN`

docker-entrypoint.sh
```diff
@@ -1,19 +1,7 @@
...
-  CRONTAB_ENV="${CRONTAB_ENV}BUNDLE_APP_CONFIG=$BUNDLE_APP_CONFIG\n"
-fi
-
-if test -n "$DATABASE_URL"; then
-  CRONTAB_ENV="${CRONTAB_ENV}DATABASE_URL=$DATABASE_URL\n"
-fi
-
-echo "$CRONTAB_ENV$CRONTAB" | crontab -
-
+# allow re-using the instance - https://stackoverflow.com/a/38732187/2866660
```
paroga: 👍

Dockerfile
```diff
@@ -1,45 +1,57 @@
...
+    bundle exec rake db:setup assets:precompile && \
+    rm -Rf config/app_config.yml tmp/* && \
     /etc/init.d/mysql stop && \
-    rm -rf /run/mysqld /tmp/* /var/lib/mysql /var/log/mysql* && \
+    rm -Rf /run/mysqld /tmp/* /var/tmp/* /var/lib/mysql /var/log/mysql* && \
     apt-get purge -y --auto-remove mysql-server && \
-    rm -rf /var/lib/apt/lists/*
+    rm -Rf /var/lib/apt/lists/* /var/cache/apt/* && \
+    \
+    bundle exec whenever >crontab
```
paroga: I'd move the crontab stuff into the previous `RUN`, so the second `RUN` contains only the ugly mysql hack

wvengen:
Great. I've added a little bit of documentation. Also tried to get docker-compose for development working, but didn't finish it. First the production setup. I'll (squash-)merge in a few days unless there are objections.

doc/DEPLOYMENT.md
```diff
@@ -1,8 +1,67 @@
...
+      -v `pwd`/config/app_config.yml:/usr/src/app/config/app_config.yml:ro \
+      foodsoft:dev  bundle exec rake db:setup
+
+To run the worker (recommended!), supply a different command
+(see [Procfile](../Procfile) for other types):
+
+    docker run --name foodsoft_mail -p 3000 \
+      -e SECRET_KEY_BASE -e DATABASE_URL -e REDIS_URL \
+      -v `pwd`/config/app_config.yml:/usr/src/app/config/app_config.yml:ro \
+      foodsoft:dev  ./proc-start mail
```
paroga: the text is about `worker`, but the code is about `mail`

proc-start
```diff
@@ -0,0 +1,24 @@
...
+  exit 1
+fi
+
+CMD=`cat Procfile | grep "^$1:" | cut -d: -f2-`
+if [ ! "$CMD" ]; then
+  echo "Process type $1 not found in Procfile" 1>&2
+  exit 1
+fi
+
+exec /bin/sh -c "$CMD"
```
paroga: i'm ok with the current `proc-start`, but i had the idea of replacing it with `rake`tasks in the meantime. e.g:
```ruby
task "cron" => :environment do
  system 'supercronic crontab'
end
```

plugins/messages/lib/tasks/foodsoft.rake
```diff
@@ -19,11 +19,11 @@ namespace :foodsoft do
     hande_mail(ENV['RECIPIENT'], STDIN.read)
   end
 
-  desc "Start STMP server for incoming email (options: PORT=25, HOST=0.0.0.0)"
+  desc "Start STMP server for incoming email (options: PORT=2525, HOST=0.0.0.0)"
   task :reply_email_smtp_server => :environment do
-    port = ENV['PORT'].to_i
+    port = ENV['PORT'].present? ? ENV['PORT'].to_i : 2525
```
paroga: since the environment variable is set, we start it now on port `3000` too

proc-start
wvengen: Nice idea for unification. It fails on one aspect, though: `proc-start` is an entry point that always works, regardless of the technology. Does one need to use `bundle exec` or not? Is the mail receiver actually a Ruby-program, or rather an MTA that runs a rake task on receival (which could be spawned by Ruby of course, but that is less efficient)? This is all agnostic to the sysadmin using the foodsoft image in a docker-compose configuration in the current situation.
Also, I have the intention to make Foodsoft work with things like Heroku and Dokku out of the box.

plugins/messages/lib/tasks/foodsoft.rake
wvengen: Yess, that's a downside. 
Could use `SMTP_PORT` instead (in rake task), or put in the `Procfile` e.g.

    PORT=$SMTP_PORT bundle exec rake foodsoft:reply_email_smtp_server

What about `Dockerfile`'s `EXPOSE` here? Maybe ... it is cleaner to have a number of docker images, each work each type. I do find the overhead a bit too much, so right now I'm willing to make this trade-off.
What do you think?

paroga: I'd prefere `SMTP_PORT` in the rake task. Docker EXPOSE ist not a real issue, since you can expose a port from the outside.

wvengen: Great, then I'll use `SMTP_PORT` in the rake task.

wvengen: `SMTP_PORT` would also be a good candidate for outgoing smtp settings (like [here](https://stackoverflow.com/a/45748832/2866660) and [here](https://www.openproject.org/operations/configuration/)), maybe `SMTP_SERVER_PORT`?

paroga: agree!

PR #500, opened 2017-09-22
Set the path of the _foodsoft_session cookie to the slug

paroga:
This allows the usage of different foodcoops at the same time.

### DIFF
--- a/config/app_config.yml.SAMPLE
+++ b/config/app_config.yml.SAMPLE
@@ -134,6 +134,7 @@ default: &defaults
   #protocol: http
   #host: localhost
   #port: 3000
+  #script_name: "/"
 
   # Access to sharedlists, the external article-database.
   # This allows a foodcoop to subscribe to a selection of a supplier's full assortment,
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,6 +1,25 @@
 # Be sure to restart your server when you modify this file.
 
-Foodsoft::Application.config.session_store :cookie_store, key: '_foodsoft_session'
+module ActionDispatch
+  module Session
+    class SlugCookieStore < CookieStore
+      alias_method :orig_set_cookie, :set_cookie
+
+      def set_cookie(env, session_id, cookie)
+        if script_name = FoodsoftConfig[:script_name]
+          request = ActionDispatch::Request.new env
+          path = request.original_fullpath[script_name.size..-1]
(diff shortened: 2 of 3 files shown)


config/initializers/session_store.rb
```diff
@@ -1,6 +1,22 @@
...
+module ActionDispatch
+  module Session
+    class SlugCookieStore < CookieStore
+      alias_method :orig_set_cookie, :set_cookie
+
+      def set_cookie(env, session_id, cookie)
+        request = ActionDispatch::Request.new env
+        slug = request.original_fullpath.split('/', 3).second
+        return if slug.blank?
+        cookie[:path] = '/' + slug
```
wvengen: This would prevent installation running on a subdomain (I'm doing that currently). Would it be possible to get the base url from somewhere (I think it is already set in the mailer config, or else we _do_ have a request here, I guess)?

paroga: hmm, good point. we are in a very early phase of the request handling, where we do not have all this information. so i'd suggest a new config variable. if that variable is not set, the old behaviour (without cookie path) remains

wvengen: Yes, we do have [`host` and such](https://github.com/foodcoops/foodsoft/blob/master/config/app_config.yml.SAMPLE#L134), could add `path` which defaults to `/`? (Right now I'm setting hostname to `hostname/path` which works for the mailer.)

wvengen: Yes, we do have [`host` and such](https://github.com/foodcoops/foodsoft/blob/master/config/app_config.yml.SAMPLE#L134), could add `path` which defaults to `/`? (Right now I'm setting hostname to `hostname/path` which works for the mailer.) Then `config/environments/production.rb` could include both hostname and path for hostname.

paroga:
i'll updated the PR

config/app_config.yml.SAMPLE
```diff
@@ -135,6 +135,9 @@ default: &defaults
   #host: localhost
   #port: 3000
 
+  # base path for the (session) cookie, to support multiple sessions in a multi_coop_install
+  #cookie_base_path: /
```
wvengen: Wouldn't it be easier config if there were just a single `base_path` which both cookie and mailer could use?

paroga: i changed it to `script_name`to match the rails configuration name

PR #502, opened 2017-09-24
Add Discourse SSO to allow login into Discourse via Foodsoft

paroga:


paroga:
could you review the second Discourse PR too? With this change Foodsoft acts as user database for Discourse, while the already merged commit uses Discourse as a user database for Foodsoft

wvengen:
I do doubt whether the Discourse plugin should be enabled by default. I suppose you need to be an admin of a Discourse installation to configure it, and that makes it likely that you're running your own Foodsoft as well.

paroga:
Ad OAuth2) One initial point for me to implement #429 was the OAuth2 API. ;-) But I have a "problem" in combination with Discourse. We would need to register Discourse as an OAuth2 application, which would require a fitting UI and so. Also Discourse handles its built in SSO different than "social login providers". With the built in SSO you get a tighter integration and control exactly "from outside Discourse" who can login, which is not possible otherwise.

I don't see any harm done by this plugin. It does nothing when not configured. I'd like to use a vanilla Foodsoft for the Austrian setup in the future, so it would be great if it would work out of the box.

wvengen:
Interesting, didn't know that :) And yes, I understand its tighter integration, great to have that.
I do have (yet vague) ideas about auto-registration of OAuth2 applications for plugins / integrated apps.

paroga:
Even if Doorkeeper would support [dynamic registration](http://openid.net/specs/openid-connect-registration-1_0.html), it would still make configuration a lot harder. This way I only need to set the config variables in `app_config.yml` (which can be supplied via environment variables in a Docker setup), which makes it very easy.

PR #507, opened 2017-10-02
Do not accept invalid addresses in SMTP RCPT TO

paroga:


paroga:
@wvengen can you help me with the database lock problem during the test? I'm not so deep into the testing framework 😕

paroga:
@wvengen what do you think about merging it with a disabled test for now?

wvengen:
Sounds like a good [idea](https://www.rubypigeon.com/posts/rspec-core-cheat-sheet/#skipping) (with a comment shortly explaining why it's skipped).
Thanks for adding a spec :+1:

PR #509, opened 2017-10-05
Use ExceptionNotification via generator generated code

paroga:
Ensure that FoodsoftConfig is loaded before running the initalizers.
This also enables notifaction for Resque.


Gemfile
```diff
@@ -63,10 +64,6 @@ gem 'foodsoft_discourse', path: 'plugins/discourse'
 #gem 'foodsoft_documents', path: 'plugins/documents'
 
 
-group :production do
-  gem 'exception_notification'
-end
```
wvengen: Why?

Gemfile
```diff
@@ -63,10 +64,6 @@ gem 'foodsoft_discourse', path: 'plugins/discourse'
...
-
```
wvengen: Why? The initializer can be wrapped in a `if defined?(ExceptionNotification)`.

config/initializers/01_load_app_config.rb
```diff
@@ -0,0 +1,6 @@
+# Initial load the default config and database from rails environment
+# See config/app_config.yml for further details
+# Load Config, start by selecting defaults via current environment
+require 'foodsoft_config'
+FoodsoftConfig.init
+FoodsoftConfig.select_default_foodcoop
```
wvengen: I like this.

Gemfile
paroga: i want to use `ExceptionNotifier.notify_exception` on some places, where the admin should be notified about exceptions. E.g. when receiving a mail throws an exception or when a specific action in a controller has a `rescue`, but the error might be relevant to admins too

config/initializers/exception_notification.rb
```diff
@@ -0,0 +1,53 @@
...
+
+ExceptionNotification.configure do |config|
+  # Ignore additional exception types.
+  # ActiveRecord::RecordNotFound, Mongoid::Errors::DocumentNotFound, AbstractController::ActionNotFound and ActionController::RoutingError are already added.
+  # config.ignored_exceptions += %w{ActionView::TemplateError CustomError}
+
+  # Adds a condition to decide when an exception must be ignored or not.
+  # The ignore_if method can be invoked multiple times to add extra conditions.
+  config.ignore_if do |exception, options|
+    not Rails.env.production?
```
wvengen: Maybe `Rails.env.development? || Rails.env.test?` - so that staging (or e.g. a demo or whatever environment someone introduces) does get notifications.

Gemfile
wvengen: That's a good idea! :+1:

config/initializers/exception_notification.rb
paroga: should that then not use the production environment?

wvengen: I know having a `RAILS_ENV=staging` as common Rails practice, with a different environment configuration. See [rails guides](http://guides.rubyonrails.org/configuring.html#creating-rails-environments).

paroga: ok, will change it to your suggestion

paroga: done

Issue #496, opened 2017-08-20
Switch license to AGPL


paroga:
GPL does not require hosters to publish their changes, but AGPL does.

wvengen:
That would be a good move, in my opinion.
Is it compatible with other licenses of gems and such we use?
I'm curious to hear other's opinions on this.

wvengen:
Great step! What are the licenses of third-party components we include (e.g. in vendor)? Are they compatible? (e.g. LGPL [is not](https://opensource.stackexchange.com/questions/5637/use-lgpl-code-in-an-agpl-project) without measures)

Also, I would put this through the developers mailing-list, just to make sure everbody is on board. Not everyone may be watching Github.

wvengen:
Also, I would not want the license to fence off people wanting to use Foodsoft, as long as it's within our realm of values.
What is our position if, for example, a foodcoop changes the source code, and hosts the service for their own use without making it open source (not or not yet). What would we want? Is is allowed by the AGPL? Do we want people to contact us and allow the developer team to provide (temporary) exceptions?
I think it would be good to have some clarity on these questions before proceeding.

paroga:
I want that if somebody changes something in the source code he/she MUST provide the changes. I don't want that people "improve" the code only for their own foodcoop. That's not the idea behind open source. I think we all put enough time into the code, that it's a reasonable requirement that other people must at least share their code changes. If that fences people off, then they should not be allowed to use foodcoop. An unmodified installation (which is the 99% use case) is still possible without any further restrictions. I do not see any reason for (temporary) exceptions. I want that other people have to share their code changes if they use a modified version of foodsoft. That's the base idea of the GPL. AGPL just adopts that idea for network services.
So I think our position should be exactly what is written in the AGPL.

wvengen:
I agree with the spirit of your last comment.
Nevertheless, I might feel uneasy even just running Foodsoft. Does changing the configuration file (`app_config.yml`) constitute a source change? The environment settings? Extra css? Tweaks in the views?
I think we could agree to a certain boundary, but as a user I would feel a little uneasy about this if I knew a little about licensing. We could make this a bit clearer by giving examples of things that we consider ok, and things that we don't consider ok. 

The GPL is clear in that respect: if you redistribute anything, make sure to also redistribute the source.
With the AGPL I think some extra clarification is needed.

paroga:
[...]

PR #513, opened 2017-10-10
Switch license to AGPL #496

paroga:


wvengen:
Please see my comments in #496, I think it would be good to think this through just a bit more.

wvengen:
If there are no objections (here, in issue, or on mailing-list), this will be merged the 10th of November just before 4.6 is released (together with some explaining text).

PR #514, opened 2017-10-10
Enable documents plugin by default

paroga:
This fixes the remaining issues from #353 to enable the plugin by default.

### DIFF
--- a/plugins/documents/app/controllers/documents_controller.rb
+++ b/plugins/documents/app/controllers/documents_controller.rb
@@ -54,10 +54,6 @@ class DocumentsController < ApplicationController
 
   def show
     @document = Document.find(params[:id])
-    filename = @document.name
-    unless filename.include? '.'
-      filename += '.' + MIME::Types[@document.mime].first.preferred_extension
-    end
-    send_data(@document.data, :filename => filename, :type => @document.mime)
+    send_data(@document.data, filename: @document.filename, type: @document.mime)
   end
 end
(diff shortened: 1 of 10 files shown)


plugins/documents/lib/foodsoft_documents/engine.rb
```diff
@@ -11,5 +11,10 @@ def navigation(primary, context)
         sub_nav.items.insert(i, sub_nav.items.delete_at(-1))
       end
     end
+
+    def default_foodsoft_config(cfg)
+      cfg[:documents_allowed_extension] = 'gif|jpg|png|txt'
+      cfg[:use_documents] = true
```
wvengen: If we want to enable this by default, I suggest we change the default app config.

paroga: it's just the same way as the other plugins do it, but i removed it for now. the only important part for me is, that the plugin is included in the `Gemfile`

plugins/documents/app/controllers/documents_controller.rb
```diff
@@ -54,10 +55,14 @@ def destroy
 
   def show
     @document = Document.find(params[:id])
-    filename = @document.name
-    unless filename.include? '.'
-      filename += '.' + MIME::Types[@document.mime].first.preferred_extension
+    send_data(@document.data, filename: @document.filename, type: @document.mime)
```
wvengen: Do we need the mime-type check on download as well? Because if the list changes, the old files are still available.
I leave that to your consideration, as long as you thought about it.

paroga: no, but i think that we should be able to download a file always. even when the list of allowed type changed in the meantime.

wvengen: Ok :+1:

PR #517, opened 2017-10-13
Add FinancialLinks

paroga:
For now this is only usefull for plugins, since there is no UI.

### DIFF
--- a/app/controllers/finance/financial_links_controller.rb
+++ b/app/controllers/finance/financial_links_controller.rb
@@ -1,0 +1,27 @@
+class Finance::FinancialLinksController < Finance::BaseController
+
+  def show
+    @financial_link = FinancialLink.find(params[:id])
+
+    @items = @financial_link.financial_transactions.map do |ft|
+      {
+        date: ft.created_on,
+        type: t('activerecord.models.financial_transaction'),
+        description: "#{ft.ordergroup.name}: #{ft.note}",
+        amount: ft.amount,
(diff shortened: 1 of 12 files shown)


wvengen:
I'm curious to learn for what you'd be using this.

paroga:
in the current form it's usefull for
1) link two `FinancialTransaction` together (transfer money from one ordergroup to another ordergroup)
2) sometimes members pay invoices directly during pickup at suppliers and we then "pay" them by adding money to the ordergroup
with the `FinacialLink` it's now possible to view the details (and verify the transactions at a later time)

with #352 invoices and finacial_transactions can be linked to the corresponding transaction at the bank account

app/controllers/finance/financial_links_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      @items << {
+        date: invoice.date || invoice.created_at,
+        type: t('activerecord.models.invoice'),
+        description: "#{invoice.supplier.name}: #{invoice.number}",
+        amount: invoice.amount,
+        link_to: finance_invoice_path(invoice)
+      }
+    end
+    @items.sort_by! { |item| item[:date] }
+  end
```
wvengen: I think this could be expressed clearer. One thing is use `map` and `+`. Maybe the functionality of gathering standard info from a `FinancialLink` relation belongs in `FinancialLink`?

paroga: I use the functions for generating the links, which is not available at the model.

PR #520, opened 2017-10-22
Add view for unpaid invoices

paroga:


app/views/finance/invoices/unpaid.html.haml
```diff
@@ -0,0 +1,27 @@
...
+      - invoices_sum += invoice.amount
+      - invoices_text << invoice.number
+      = link_to finance_invoice_path(invoice) do
+        = format_date invoice.date
+        = ' ' + invoice.number
+      = ' ' + number_to_currency(invoice.amount)
+      %br/
+    %p
+      - if supplier.iban.present?
+        %b= heading_helper(Supplier, :iban) + ':'
```
wvengen: Sorry if it causes no problem, but doesn't the string addition undo any `html_safe` that may be present in the heading helper?

paroga: the output is still safe html

Issue #449, opened 2016-10-29
Docker image for production


paroga:
The current Docker image seams ok for development, but it does not really work in production without some work.
What do you think about providing a Docker image for production with all releases as tags on Docker hub?
For that I'd like to (re)move the current Docker files and create new ones for production (including some scripts for nice bootstrapping). If you look at the period the Docker stuff did not work lately, I see no real demand in keeping the current stuff.
Background: We are now in the final phase of setting up Austrian Foodsoft Hosting and choose Docker as basis for our infrastructure. Since I do not want to maintain the Docker files in my own repository, it would be great if the official version version does already everything out-of-the-box. ;-)

wvengen:
Sounds like a great idea :+1:

simon04:
For later reference: @paroga adapted the Dockerfile file, https://github.com/foodcoopsat/foodsoft/commit/e0b1433f1e0290b0637385a903069eb68122d401. The build is running on Docker Hub: https://hub.docker.com/r/foodcoopsat/foodsoft/

paroga:
Closed with PR #482

PR #523, opened 2017-10-30
Fix setup for development with docker

wvengen:
This completes #449. Reorganizes development and production docs somewhat, and brings back the development setup using docker. It could be improved, but I think this is already a small step forward.

After this is merged, the wiki page [Deployment notes](https://github.com/foodcoops/foodsoft/wiki/Deployment-notes) may need updated links.


wvengen:
[This uid hack](https://chengl.com/using-docker-for-rails-development/) may still be useful, keeping that as a later improvement.

docker-compose-dev.yml
```diff
@@ -0,0 +1,41 @@
+version: '2'
+services:
+
+  web:
```
paroga: I'd like to align this with our foodcoops.net setup

docker-compose-dev.yml
```diff
@@ -0,0 +1,41 @@
...
+    extends: worker
```
paroga: why does web extend worker and not the other way around?

docker-compose-dev.yml
```diff
@@ -0,0 +1,41 @@
...
+    volumes:
+      - bundle:/home/app/bundle
+      - .:/home/app/src
+    environment:
+      - DATABASE_URL=mysql2://root:secret@mariadb/development?encoding=utf8
+      - REDIS_URL=redis://redis:6379
+      - QUEUE=foodsoft_notifier
+      - TEST_DATABASE_URL=mysql2://root:secret@mariadb/test?encoding=utf8
+
+  mailcatcher:
```
paroga: i see not configuration sending the mails to this container

docker-compose-dev.yml
wvengen: Ok with that.

docker-compose-dev.yml
wvengen: Took same approach as before, where current configuration detects docker environment variables and [connects to mailcatcher](https://github.com/foodcoops/foodsoft/blob/master/config/environments/development.rb.SAMPLE#L42). Later we can change this to a proper smtp url.

wvengen:
Ok to merge? I'll do so today without response.

PR #525, opened 2017-11-05
Spanish translation: first strings translated from English

acracia:
Ey hello:

Please review this translation file for Spanish, I will finish in these days. I would like to be able to show the demo in Spanish to my mates of the group...

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -196,7 +196,7 @@ en:
         last_login: Último login
         last_name: Apellido
         name: Nombre
-        nick: nombre de usuario
+        nick: apodo
         ordergroup: Grupo de pedido
         password: Contraseña
         password_confirmation: Repite la contraseña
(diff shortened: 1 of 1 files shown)


paroga:
Fist I'd like to thank you for your very valuable contribution to Foodsoft. It's very nice to hear of the Spanish community.

We use localapp for our translations. See https://www.localeapp.com/projects/4867/translations

It would be great if you could put your translation into localeapp and I'll import them for the upcoming 4.7 release next week.

Using a mixture of English and Spanish in `es.yml` raises problems, since we do not see which strings are still untranslated. Foodsoft has an automatic fallback to English if the Spanish string is missing.

IMHO we have two possible solutions: (1) translate ALL strings and import it into LocaleApp or (2) remove English strings from `es.yml` and import only Spanish strings.

acracia:
Oh I see... good I thought it was a very bad way of doing it

I see some of the strings have been already translated to spanish also

I could delete the english keys, sure. After having a look at the currently translated ones, for consistency too...

paroga:
AFAIK the spanish translations are unmaintained anyway at the moment. So feel free to replace the existing translations or adopt yours for consistency. Just let me know if I can help you somehow (beside translating the strings 😉)

acracia:
OK, I have pushed here only the strings in Spanish, they are at 
c96fd92 but I cannot see how to import them at localeapp, could you please have a look and then close this merge req?
Then I will have a look at the missing spanish translations and it would be great if i can demo the site to my group coming week 💃

paroga:
Unfortunately import is only possible for admins, but I imported the file for you. LocalApp should know all of your translations now.

acracia:
Ok, I have translated most of the stuff, do you think you could try to make the strings spread to the incomplete translations? I have the feeling some of the incomplete strings are already translated on some other file.
I have left some strings out because I cannot understand them without the context and practise of use.
Do you think it is possible to see a demo in Spanish somewhere? Thanks!

paroga:
I've sent you a private mail with credentials for a demo instance. Please let us know, if you need some additional help (e.g. in finding some translations).

Thanks again for that valuable contribution!

PR #526, opened 2017-11-05
Add polls plugin

paroga:
It's just a preview, to agree on the data model and the basic functionality.

If everything is ok, I'll improve usability. Maybe somebody has a nice idea to avoid the code duplication in the views. IMHO there is too much code in the views anyway, but moving it to the model or controller won't improve the situation either.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -210,6 +210,12 @@ class User < ApplicationRecord
     end
   end
 
+  def self.custom_fields
+    fields = FoodsoftConfig[:custom_fields] && FoodsoftConfig[:custom_fields][:user]
+    return [] unless fields
+    fields.map(&:deep_symbolize_keys)
+  end
+
   # XXX this is view-related; need to move out things like token_attributes
   #     then this can be removed
   def display
--- a/plugins/polls/app/controllers/polls_controller.rb
+++ b/plugins/polls/app/controllers/polls_controller.rb
@@ -1,0 +1,108 @@
+class PollsController < ApplicationController
+  before_filter -> { require_plugin_enabled FoodsoftPolls }
+
+  def index
+    @polls = Poll.page(params[:page]).per(@per_page).order(created_at: :desc)
+  end
+
+  def show
+    @poll = Poll.find(params[:id])
+  end
+
+  def new
+    @poll = Poll.new
+  end
+
+  def create
+    @poll = Poll.new(poll_params)
+    @poll.created_by = current_user
+
+    if @poll.save
+      redirect_to @poll, notice: t('.notice')
+    else
+      render action: 'edit'
+    end
+  end
+
+  def edit
+    @poll = Poll.find(params[:id])
+
+    if user_has_no_right
+      redirect_to polls_path, alert: t('.no_right')
+    end
+  end
+
+  def update
+    @poll = Poll.find(params[:id])
+
+    if user_has_no_right
+      redirect_to polls_path, alert: t('.no_right')
+    elsif @poll.update_attributes(poll_params)
+      redirect_to @poll, notice: t('.notice')
+    else
+      render action: 'edit'
+    end
+  end
+
+  def destroy
+    @poll = Poll.find(params[:id])
+
+    if user_has_no_right
+      redirect_to polls_path, alert: t('.no_right')
+    else
+      @poll.destroy
+      redirect_to polls_path, notice: t('.notice')
+    end
+  rescue => error
+    redirect_to polls_path, alert: t('.error', error: error.message)
+  end
+
+  def vote
+    @poll = Poll.find(params[:id])
+
+    if @poll.one_vote_per_ordergroup
+      ordergroup = current_user.ordergroup
+      return redirect_to polls_path, alert: t('.no_ordergroup') unless ordergroup
+      attributes = { ordergroup: ordergroup }
+    else
+      attributes = { user: current_user }
+    end
+
+    redirect_to @poll, alert: t('.no_right') unless @poll.user_can_vote?(current_user)
+
+    @poll_vote = @poll.poll_votes.where(attributes).first_or_initialize
+
+    if request.post?
+      @poll_vote.update!(note: params[:note], user: current_user)
+
+      if @poll.single_select?
+        choices = { params[:choice] => '1' }
+      else
+        choices = params[:choices] || {}
+      end
+
+      @poll_vote.poll_choices = choices.map do |choice, value|
+        poll_choice = @poll_vote.poll_choices.where(choice: choice).first_or_initialize
+        poll_choice.update!(value: value)
+        poll_choice
+      end
+
+      redirect_to @poll
+    end
+  end
+
+  private
+
+  def user_has_no_right
+    @poll.created_by != current_user && !current_user.role_admin?
+  end
+
+  def poll_params
+    params
+      .require(:poll)
+      .permit(:name, :starts_date_value, :starts_time_value, :ends_date_value,
+        :ends_time_value, :description, :one_vote_per_ordergroup, :voting_method,
+        :multi_select_count, :min_points, :max_points, choices: [],
+        required_ordergroup_custom_fields: [], required_user_custom_fields: [])
+  end
+end
--- a/plugins/polls/app/models/poll.rb
+++ b/plugins/polls/app/models/poll.rb
@@ -1,0 +1,50 @@
+class Poll < ActiveRecord::Base
+
+  # @!attribute required_ordergroup_custom_fields
+  #   A list of custom_fileds, which are required to poll.
+  #   If the required field on the ordergroup of the user
+  #   is empty the user will not be able to place a vote.
+  #   @return [Array<String>] Required field names.
+  # @!attribute required_user_custom_fields
+  #   A list of custom_fileds, which are required to poll.
+  #   If the required field on the user is empty the user
+  #   will not be able to place a vote.
+  #   @return [Array<String>] Required field names.
+
+  belongs_to :created_by, class_name: 'User', foreign_key: 'created_by_user_id'
+  has_many :poll_votes, dependent: :destroy
+
+  validates_presence_of :name, :choices
+  serialize :choices, Array
+  serialize :required_ordergroup_custom_fields, Array
+  serialize :required_user_custom_fields, Array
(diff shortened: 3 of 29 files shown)


wvengen:
Wow, you've got things on your sleeve :)
One part of me wonders if we need everything and the kitchen sink in Foodsoft, still I see that polls are integral part for certain foodcoops, especially with some things like the one-vote-per-ordergroup restriction. I'd be up for something like this. We probably need to rethink the configuration UI at some point.

plugins/polls/app/models/poll_vote.rb
```diff
@@ -0,0 +1,6 @@
+class PollVote < ActiveRecord::Base
+  belongs_to :poll
+  belongs_to :ordergroup
+  belongs_to :user
+  has_many :poll_choices, dependent: :destroy
+end
```
wvengen: I like that the models start with the same name, keeps them together.

wvengen:
It will need some documentation, especially the poll options. Maybe even in the UI.

paroga:
Yes, we need this in foodsoft. We need polls grouped by ordergroups, while the membership information is only stored in foodsoft. I agree 100% that this is no core functionality, but since it's only in a plugin i see nothing which speaks against it.

There's still a lot missing. The question is, if we can agree already on the data schema. If we do I can deploy it already, without fear of later incompatible schema changes.

plugins/polls/app/models/poll_vote.rb
paroga: what do you mean? rename `PollChoice` to `PollVoteChoice` and `poll_choices` to `poll_vote_choices`?

wvengen:
Regarding the data model, I had a quick look, and the only remarkable thing I noticed is that the choices is a serialized array. My experience is that questions could arise where a full-blown table is useful. Nevertheless, that can be done in a migration, so I see no objections there.
Unless someone else comes with a different poll plugin out of the blue with a different data model, I see no problems ;)

paroga:
I hat the choices as multiple rows in the database first, but it made the code needless more complex (all the joins become ugly very fast).

I don't think anybody will come up with a new poll plugin anytime soon. 😁

wvengen:
Sounds good for now :+1:

paroga:
@wvengen i've done some work here to, maybe we can polish it up, to get it into the official version

plugins/polls/app/controllers/polls_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+      redirect_to @poll, notice: t('.notice')
+    else
+      render action: 'edit'
+    end
+  end
+
+  def edit
+    @poll = Poll.find(params[:id])
+
+    if @poll.created_by != current_user && !current_user.role_admin?
```
wvengen: What about adding a poll role? It's probably on your todo-list.
Not sure if we've already implemented additional roles for a plugin.

plugins/polls/app/models/poll.rb
```diff
@@ -0,0 +1,37 @@
...
+  serialize :choices, Array
+  serialize :requied_ordergroup_custom_fields, Array
+  serialize :requied_user_custom_fields, Array
+  enum voting_method: { event: 0, single_select: 1, multi_select: 2, points: 3, resistance_points: 4 }
+
+  include DateTimeAttributeValidate
+  date_time_attribute :starts, :ends
+
+  def available_points
+    return [] if min_points.nil? || max_points.nil?
```
wvengen: Maybe return an empty range?

plugins/polls/app/models/poll_vote.rb
wvengen: It's good as it is, that's what I was trying to say :+1:

plugins/polls/app/views/polls/vote.html.haml
```diff
@@ -0,0 +1,54 @@
+- content_for :javascript do
+  :javascript
+    $(function() {
+    });
```
wvengen: leftover?

plugins/polls/db/migrate/20181110000000_create_polls.rb
```diff
@@ -0,0 +1,37 @@
...
+  def change
+    create_table :polls do |t|
+      t.integer :created_by_user_id, null: false
+      t.string :name, null: false
+      t.text :description
+      t.datetime :starts
+      t.datetime :ends
+      t.boolean :one_vote_per_ordergroup, default: false, null: false
+      t.text :requied_ordergroup_custom_fields
+      t.text :requied_user_custom_fields
```
wvengen: Idem, typo.

plugins/polls/lib/foodsoft_polls/engine.rb
```diff
@@ -0,0 +1,15 @@
...
+    def navigation(primary, context)
+      return unless FoodsoftPolls.enabled?
+      return if primary[:foodcoop].nil?
+      sub_nav = primary[:foodcoop].sub_navigation
+      sub_nav.items <<
+        SimpleNavigation::Item.new(primary, :polls, I18n.t('navigation.polls'), context.polls_path)
+      # move to right before tasks item
+      if i = sub_nav.items.index(sub_nav[:tasks])
+        sub_nav.items.insert(i, sub_nav.items.delete_at(-1))
+      end
```
wvengen: Perhaps a fallback for when the tasks item is not found. You can get the index of the tasks item and add it right before, and if it isn't found, add it at the end. Another plugin may disable the tasks menu item.

plugins/polls/app/controllers/polls_controller.rb
paroga: we don't have this somewhere else ATM. it would require an additional col in the groups table, which i do not want to touch via a plugin. since i see no need for a new role, i'd like to avoid it for now

plugins/polls/lib/foodsoft_polls/engine.rb
paroga: the `if` does exactly that: if `:tasks` has been found will be removed and inserted at the corresponding position

app/models/user.rb
```diff
@@ -210,6 +210,12 @@ def self.authenticate(login, password)
     end
   end
 
+  def self.custom_fields
+    fields = FoodsoftConfig[:custom_fields] && FoodsoftConfig[:custom_fields][:user]
+    return [] unless fields
+    fields.map(&:deep_symbolize_keys)
+  end
+
```
wvengen: This looks like it could benefit if becoming a concern?

plugins/polls/app/controllers/polls_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+      redirect_to polls_path, alert: t('.no_right')
+    end
+  end
+
+  def update
+    @poll = Poll.find(params[:id])
+
+    if @poll.created_by != current_user && !current_user.role_admin?
+      redirect_to polls_path, alert: t('.no_right')
+    elsif @poll.update_attributes(params[:poll])
```
wvengen: I'd strongly recommend using strong parameters here.

plugins/polls/app/controllers/polls_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    if @poll.created_by != current_user && !current_user.role_admin?
+      redirect_to polls_path, alert: t('.no_right')
+    end
+  end
+
+  def update
+    @poll = Poll.find(params[:id])
+
+    if @poll.created_by != current_user && !current_user.role_admin?
```
wvengen: I'd favour moving this check to a (privat) method in the controller (also because it is repeated in `destroy`).

plugins/polls/app/views/polls/_form.html.haml
```diff
@@ -0,0 +1,76 @@
...
+        return false;
+      });
+
+      $('a[data-add-choice]').on('touchclick', function() {
+        choice.clone().appendTo('#choices');
+        return false;
+      });
+
+      updateSlideState();
+    });
```
wvengen: Let's put Javascript in an external file (unless it is really tiny or contains Ruby variables ).

plugins/polls/app/views/polls/show.html.haml
```diff
@@ -0,0 +1,69 @@
...
+  %tbody
+    - for vote in @poll.poll_votes
+      %tr
+        %td
+          - if vote.ordergroup.nil?
+            = show_user vote.user
+          - else
+            = "#{vote.ordergroup.name} (#{show_user vote.user})"
+        - @poll.choices.size.times do |idx|
+          - if choice = vote.poll_choices.where(choice: idx).first
```
wvengen: This triggers a SQL query each time, better allow preloading this.

plugins/polls/db/migrate/20181110000000_create_polls.rb
```diff
@@ -0,0 +1,37 @@
...
+  def change
+    create_table :polls do |t|
+      t.integer :created_by_user_id, null: false
+      t.string :name, null: false
+      t.text :description
+      t.datetime :starts
+      t.datetime :ends
+      t.boolean :one_vote_per_ordergroup, default: false, null: false
+      t.text :required_ordergroup_custom_fields
+      t.text :required_user_custom_fields
```
wvengen: I think it would be important to explain the custom fields functionality.

plugins/polls/lib/foodsoft_polls/engine.rb
wvengen: Ah, now I see, super :)

plugins/polls/app/controllers/polls_controller.rb
wvengen: :+1:

plugins/polls/lib/foodsoft_polls/engine.rb
wvengen: Ah, I missed that before, super.

plugins/polls/db/migrate/20181110000000_create_polls.rb
paroga: done

plugins/polls/app/views/polls/show.html.haml
paroga: good catch! done

app/models/user.rb
paroga: i rebased it to the current `master`

PR #527, opened 2017-11-06
 Add view to sort orders by pickup day

paroga:
This provides an interface to generate PDFs which include multiple orders.

It reuses the `MultipleOrdersBy*` classes from the `current_orders` plugin for now. A next step is to merge `MultipleOrdersBy*` into the other `OrderPdf` classes and clean them up.

### DIFF
--- a/app/controllers/pickups_controller.rb
+++ b/app/controllers/pickups_controller.rb
@@ -1,0 +1,26 @@
+class PickupsController < ApplicationController
+
+  def index
+    @orders = Order.order('pickup DESC').group_by { |o| o.pickup }
+  end
+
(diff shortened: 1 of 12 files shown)


wvengen:
Good idea!

paroga:
what do you think about including it in 4.6?

wvengen:
Would be nice feature-wise, but I think it's wise to wait until 4.7, because the release is due in two days.

paroga:
@wvengen ping

wvengen:
Depends on #532, right? Otherwise, merge away! Great to have this in main Foodsoft.

PR #528, opened 2017-11-08
Add folders to the documents plugin

paroga:


wvengen:
At very quick glance, seems ok. As it's a plugin that you wrote, I'd say: merge when you think ready :+1:

PR #532, opened 2017-11-15
Rewrite OrderPdfs to support multiple orders at once

paroga:
This implements `MultipleOrdersBy*` from the `current_orders` in a clean way, so we can implement #527 in a clean way.

### DIFF
--- a/app/documents/order_by_articles.rb
+++ b/app/documents/order_by_articles.rb
@@ -1,12 +1,6 @@
 # encoding: utf-8
 class OrderByArticles < OrderPdf
 
-  # optimal value depends on the average number of ordergroups ordering an article
-  #   as well as the available memory
-  BATCH_SIZE = 50
-
-  attr_reader :order
-
   def filename
     I18n.t('documents.order_by_articles.filename', :name => order.name, :date => order.ends.to_date) + '.pdf'
   end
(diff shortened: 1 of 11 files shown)


app/models/group_order_article.rb
```diff
@@ -195,4 +195,8 @@ def total_price(order_article = self.order_article)
   def result_manually_changed?
     result != result_computed unless result.nil?
   end
+
+  def quantity_with_tolerance
+    tolerance > 0 ? "#{quantity} + #{tolerance}" : "#{quantity}"
+  end
```
wvengen: Are you sure to put this here? In html we may want to include coloring, for example. Maybe keep it in a pdf-related module?

app/documents/order_by_articles.rb
```diff
@@ -18,59 +12,30 @@ def title
...
-
-      rows = []
       dimrows = []
-      each_group_order_article_for(order_article) do |goa|
+      rows = [[
+        GroupOrder.human_attribute_name(:ordergroup),
+        GroupOrderArticle.human_attribute_name(:ordered),
+        GroupOrderArticle.human_attribute_name(:received),
+        GroupOrderArticle.human_attribute_name(:total_price)
+      ]]
```
wvengen: Yeah, much better to maintain :)

lib/order_pdf.rb
```diff
@@ -1,104 +1,130 @@
...
+
+      results = goa_records.group_by(&:first).transform_values do |value|
+        grouped_value = value.group_by(&:second)
+        group_ids.map do |group_id|
+          grouped_value[group_id].try(:first).try(:third)
+        end
+      end
+
+      yield go_records, results
+      offset += batch_size
```
wvengen: This method could really use some documentation.

lib/render_pdf.rb
```diff
@@ -0,0 +1,173 @@
...
+    body  # Add content, which is defined in subclasses
+    render  # Render pdf
+  end
+
+  # Helper method to test pdf via rails console: OrderByGroups.new(order).save_tmp
+  def save_tmp
+    File.open("#{Rails.root}/tmp/#{self.class.to_s.underscore}.pdf", 'w') {|f| f.write(to_pdf.force_encoding("UTF-8")) }
+  end
+
+  # XXX avoid underscore instead of unicode whitespace in pdf :/
```
wvengen: `@todo` is used in other places

lib/order_pdf.rb
```diff
@@ -1,104 +1,130 @@
...
+  def each_order_article(&block)
+    OrderArticle.
+      ordered.
+      includes(article: :supplier).
+      includes(group_order_articles: {group_order: :ordergroup}).
+      where(order: @orders).
+      order('suppliers.name, articles.name, groups.name').
+      preload(:article_price).
+      each(&block)
+  end
```
wvengen: Can we split the scope and each? This can help plugins to add preloading for data to add to the pdf, for example.

paroga:
screen shot

paroga:
@wvengen ping

app/documents/order_by_articles.rb
```diff
@@ -18,59 +12,30 @@ def title
...
+
+      order_article.group_order_articles.each do |goa|
```
wvengen: Would be good to use a function getting the goa's, e.g. if a plugin wants to add an additional field hat needs to be selected from the database.

PR #552, opened 2018-05-04
improve usability when ordering

mortbauer:
also show box with links to currently open orders after the order has
been saved

I'm until now only a use of the great foodsoft, however one annoying thing for me is, that one has to go all the way back through the main menu for placing a new order after the current edited order has been saved.
Since I'm completely new to ruby and rails I just copied the code more or less from the _form.html.haml file. 

This is the result you get with this change:
image

### DIFF
--- a/app/views/group_orders/show.html.haml
+++ b/app/views/group_orders/show.html.haml
@@ -6,27 +6,32 @@
 
 - title t('.title', order: @order.name)
 
-.well
-  // Order summary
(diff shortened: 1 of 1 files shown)


wvengen:
Hi, thanks for your contribution! I'm sorry it took so long to respond, somehow this passed without me noticing. If someone else can confirm this works in practice, I'd be happy to merge (else I'll see if I can test it myself some time).

PR #556, opened 2018-09-16
Add model and views for bank accounts

paroga:


paroga:
@wvengen ping?

app/models/bank_account.rb
```diff
@@ -0,0 +1,12 @@
+class BankAccount < ActiveRecord::Base
+
+  has_many :bank_transactions, dependent: :destroy
+
+  normalize_attributes :name, :iban, :description
+
+  validates :name, :presence => true, :uniqueness => true, :length => { :minimum => 2 }
+  validates :iban, :presence => true, :uniqueness => true
+  validates_format_of :iban, :with => /\A[A-Z]{2}[0-9]{2}[0-9A-Z]{,30}\z/
```
wvengen: Do we need to validate the check digit?

app/views/finance/bank_transactions/show.html.haml
```diff
@@ -0,0 +1,40 @@
...
+  %b= heading_helper(BankTransaction, :financial_link) + ':'
+  - if @bank_transaction.financial_link
+    = link_to t('ui.show'), finance_link_path(@bank_transaction.financial_link)
+  - else
+    = link_to t('.add_financial_link'), '/'
+
+- unless @bank_transaction.receipt.blank?
+  %pre= preserve @bank_transaction.receipt
+- if @bank_transaction.image
+  %p= image_tag @bank_transaction.image_url
```
wvengen: This is a bit ugly for a view. I this is present on other places, but for new views, what about doing a table or a form with static controls? [not a blocking question]

config/locales/en.yml
```diff
@@ -902,6 +903,18 @@ en:
         reload: Reload summary
         with_extra_charge: 'with extra charge:'
         without_extra_charge: 'without extra charge:'
+    bank_accounts:
+      index:
+        account_statement: Account statement
+        title: Bank Accounts
+    bank_transactions:
+      index:
+        search_placeholder: Search ...
```
wvengen: What about moving this to a general i18n section (like cancel etc.)?

app/models/bank_account.rb
paroga: we don't do it at any of the other places (using a IBAN checkin library is IMHO a independent point)

app/views/finance/bank_transactions/show.html.haml
paroga: 100% agree, but i don't want to start a new design pattern. i'd like to do this in a more general "make the views beautiful" change, if you don't have something i can copy from

config/locales/en.yml
paroga: agree

app/views/finance/bank_transactions/show.html.haml
wvengen: Ok, that's fair. We have several things here.

wvengen: Ok, that's fair. We have several approaches currently with their pros and cons.

app/models/bank_account.rb
wvengen: Agree.

paroga:
@wvengen i've pushed a new version...

PR #561, opened 2018-10-11
Bundle update

wvengen:
Updated dependencies require Ruby 2.3 or higher.

I propose to do a new release sometime not far long off (mentioning the new required Ruby version).

- [x] Fix wiki rendering.
- [x] Update mysql2 gem.


Gemfile.lock
```diff
@@ -230,35 +231,36 @@ GEM
...
-    mime-types-data (3.2016.0521)
+    mime-types-data (3.2018.0812)
+    mini_mime (1.0.1)
     mini_portile2 (2.3.0)
-    minitest (5.10.3)
+    minitest (5.11.3)
     mono_logger (1.1.0)
-    multi_json (1.12.2)
+    multi_json (1.13.1)
     mysql2 (0.4.9)
```
paroga: I'd like to see `mysql2` at `0.4.10` since it includes a buildfix for newer mysql versions.

What do you think about a more aggressive dependency update in the future?

PR #562, opened 2018-10-11
Add abbreviation to FinancialTransactionType

paroga:
See #352 @ https://github.com/foodcoops/foodsoft/pull/352/commits/4b82c235056a2b7f477a6a42123f24d9132aacf0 for the usage. Maybe there is a better name for this field...


wvengen:
Maybe `name_short`? That is like the i18n used for `heading_helper`, right?

PR #564, opened 2018-10-11
Add import route for bank_transactions

paroga:


app/models/bank_account.rb
```diff
@@ -9,4 +9,6 @@ class BankAccount < ActiveRecord::Base
   validates_format_of :iban, :with => /\A[A-Z]{2}[0-9]{2}[0-9A-Z]{,30}\z/
   validates_numericality_of :balance, :message => I18n.t('bank_account.model.invalid_balance')
 
+  def find_import_method
+  end
```
wvengen: I think this method could use a comment, and raise an exception when there is no valid method (in the manner of AR's find). Or document that it could return `nil`.

wvengen: It appears I was off: [`find_by`](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by) can return `nil` (but [find](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find) not).

Still let's document what this method does (including `@return` mentioning `nil` and the return type), also because it's meant to be hooked by plugins (I guess).
And for the moment, `@todo` with CSV would be useful as well.

paroga: i've added some comments

app/controllers/finance/bank_accounts_controller.rb
```diff
@@ -5,4 +5,18 @@ def index
     redirect_to finance_bank_account_transactions_url(@bank_accounts.first) if @bank_accounts.count == 1
   end
 
+  def import
+    @bank_account = BankAccount.find(params[:id])
+    import_method = @bank_account.find_import_method
+    if import_method
+      count = import_method.call(@bank_account)
```
wvengen: As it's used now, `@backaccount.import` would make more sense.  But perhaps you'd like to go somewhere where this makes sense. Let's see where it goes.

paroga: since the import methods will be implemented in a plugin i need a place where i can hook them in and it's an clean way for the CSV import in the future

PR #565, opened 2018-10-12
Add reference_calculator

paroga:


config/locales/en.yml
```diff
@@ -1173,6 +1173,11 @@ en:
       user:
         since: "(member for %{when})"
         title: "%{user}"
+    reference_calculator:
+      text0: Please transfer
+      text1: with the refernce
```
wvengen: typo

app/views/home/reference_calculator.haml
```diff
@@ -0,0 +1,44 @@
...
+        $('#amount').text(amount);
+        $('#reference').text(reference);
+      }
+
+      $(document).on('input', 'input[data-use-me]', update);
+      update();
+    });
+
+= form_tag finance_create_transaction_collection_path do
+  = hidden_field_tag 'prefix', "FS#{@current_user.ordergroup.id}.#{@current_user.id}"
```
wvengen: this seems pretty hardcoded, would be good to make that more flexible, preferably on some model

app/controllers/home_controller.rb
```diff
@@ -14,6 +14,11 @@ def index
   def profile
   end
 
+  def reference_calculator
+    @iban = BankAccount.first.try(:iban)
+    @types = FinancialTransactionType.with_name_short.order('name ASC')
```
wvengen: this will need an explanation somewhere on the form where short name is entered and in the documentation, that only financial transaction types with a short name will be selectable in the reference calculator.

paroga:
Example two users 1 in ordergroup 10 and user 2 in odergroup 20. we have two financial transaction types A an B

if we find a bank transaction witch matches FS10.1A100 with an amount of 100 EUR, we can create a transaction A with 100EUR at ordergroup 10. For "FS20.2A30B70" we can automatically create two financial transactions for ordergroup 20: A with 30EUR and B with 70 EUR.
When members use this schema for all bank transfers they get automatically assigned to the correct ordergroups (and linked with financial links)

This reference calculator helps the members to create the correct string, if they do not get how the system works (or don't know their prefix).

Once again #352 @ 4b82c23 shows where this is heading.

app/views/home/reference_calculator.haml
wvengen: What about using a helper class for encoding and decoding those strings? I understand the JS part is then a bit of an issue, so either we could compute it server-side, or include a method in the file with the js-computation part (only the computation part, not the display part).

Also, I would really suggest to add a check digit somewhere (perhaps optional, but for generated ones this helps detecting cut-n-paste errors or typos). But can be a feature to add later.

paroga: i really don't like to calculate it on the server side, but we can but the prefix in the global config. please check out #568 for the parsing part, so we can sync the ideas

>  add a check digit somewhere

it's important that this format can be generated by humans too. they only need to know the number of the ordergroup (which stays the same all the time). As long as the check digit is optional it does not provide a real value and we have a kind of check already, since the sum of the amounts in the reference must match the amount of the bank transaction

wvengen: Ok, fine with the no-check-digit route.

I still think putting the reference computation logic in one place would help in the future, and shouldn't be so involved, right? I don't know, but somehow I'd like to keep this a clearly defined piece, so it can be customized by foodcoops / plugin authors.

paroga: doing the stuff in a CLEAN way on the server is far from trivial and i don't see the benefits outweigh the effort. If someone will ever implement a plugin, she can update the existing code when needed. i don't want to over engineer it.
maybe you can give #568 a look before we continue here

wvengen: I agree about doing everything server-side, that's too involved.
But what about adding a module in `lib/` that
- converts fields (hash?) to a reference (string)
- converts a reference (string) to fields (hash?)
- emits the javascript needed to convert fields (js object?) to a reference (string)

That would be easy and give flexibility.

wvengen: The js could be something like (just a quick try):

```js
function(prefix, values) {
  var amount = 0;
  var reference = prefix;
  values.each(function(o) {
     var name = o.name;
     var value = Math.round(parseInt(o.value, 10));
     if (value > 0) {
       amount += value;
       reference += name + value;
     }
   });
   return { amount: amount, reference: reference };
}
```

paroga: i'll give it a try, but i like to get at least an initial review on #568 first, since there might be some overlapping points, which i'd like to implement together

app/models/financial_transaction_type.rb
```diff
@@ -9,6 +9,8 @@ class FinancialTransactionType < ActiveRecord::Base
 
   before_destroy :restrict_deleting_last_financial_transaction_type
 
+  scope :with_name_short, -> { where.not(name_short: '') }
```
wvengen: I guess `where.not(name_short: [nil, ''])` would be safer here?

app/views/layouts/application.html.haml
```diff
@@ -9,6 +9,8 @@
       %ul.dropdown-menu
         %li= link_to t('.profile'), my_profile_path
         %li= link_to t('.ordergroup'), my_ordergroup_path
+        - if BankAccount.any?
+          %li= link_to t('.reference_calculator'), home_reference_calculator_path
```
wvengen: I guess people would use bank accounts without the reference stuff you're using?
What about a configuration option for using the references to direct where the money is for? Or at the very least, if there is a `name_short` somewhere ... but that would need some caching perhaps, since this is called on every page view.

app/views/home/reference_calculator.haml
wvengen: One other thing: I'd consider moving the function to an asset, so it can be more easily overridden. Maybe set the function on `window.foodsoftParseBankTransactionReference`? Or make the function name a controller instance variable, so it can be overridden? Deface has issues overriding Javascript in places like this.

app/views/layouts/application.html.haml
wvengen: This can also be done later, when someone requests it. Fine.

PR #568, opened 2018-10-12
Add automatic generation of financial transactions and links

paroga:


app/controllers/finance/bank_accounts_controller.rb
```diff
@@ -5,6 +5,14 @@ def index
     redirect_to finance_bank_account_transactions_url(@bank_accounts.first) if @bank_accounts.count == 1
   end
 
+  def assign_unlinked_transactions
+    @bank_account = BankAccount.find(params[:id])
```
wvengen: What about `params.require(:id)`?

app/models/bank_transaction.rb
```diff
@@ -33,4 +33,55 @@ class BankTransaction < ActiveRecord::Base
...
   end
+
+  def assign_to_invoice
+    return false unless supplier
+
+    content = text
+    content += "\n" + reference if reference.present?
+    invoices = supplier.invoices.unpaid.select {|i| content.include? i.number}
+    invoices_sum = invoices.inject(0) {|sum,i| sum + i.amount }
+    return false if amount != -invoices_sum
```
wvengen: Just to check: may this introduce float comparison problems?

app/models/bank_transaction.rb
```diff
@@ -33,4 +33,55 @@ class BankTransaction < ActiveRecord::Base
...
+
+    ActiveRecord::Base.transaction do
+      link = FinancialLink.new
+      invoices.each {|i| i.update_attributes financial_link: link, paid_on: date}
+      update_attribute :financial_link, link
```
wvengen: No `update_attributes!` (with exclamation mark) here, to catch any errors early?

app/models/bank_transaction.rb
```diff
@@ -33,4 +33,55 @@ class BankTransaction < ActiveRecord::Base
...
+    end
+
+    return true
+  end
+
+  def assign_to_ordergroup
+    m = /FS(?<group>\d+)(\.(?<user>\d+))?(?<parts>([A-Za-z]+\d+(\.\d+)?)+)/.match(reference)
+    return unless m
```
wvengen: My hunch here would be to put this and further matching elsewhere (to avoid mega-models). Still this is pretty coupled to a bank transaction. I don't have a clear head on what is needed here to keep a clean architecture. ... :/

wvengen: Perhaps something that takes bank transactions, and deducts what they are for.
This needs knowledge of other models (like you're querying `Ordergroup` here).

Thinking aloud, perhaps something like this: with input a transaction with reference `FS20.5.Membership25` could result in the hash
```ruby
{
  ordergroup_id: 20,
  user_id: 5,
  financial_transactions: [
    { name_short: 'Membership' amount: 25 }
  ]
}
```

This method would do all the validation it can (except checking that `name_short` exists, and that user and ordergroup would belong together, if that may ever become a requirement).
This could also help to aid the user in interpreting the result: e.g. when the amounts don't add up, the ordergroup could choose in Foodsoft how to distribute the funds over transaction classes (or if we'd start looking at unmatched transactions from the user's verified IBAN, then he could also correct ordergroup typo's).

I hope you don't experience my thoughts as "blocking", and will see the beautify of thinking together how this can be implemented in an extensible way. We could also just merge it, and see what happens, but perhaps we can learn something from pausing for a minute.

app/models/bank_transaction.rb
```diff
@@ -33,4 +33,55 @@ class BankTransaction < ActiveRecord::Base
   def image_url
     'data:image/png;base64,' + Base64.encode64(self.image)
   end
+
+  def assign_to_invoice
+    return false unless supplier
+
+    content = text
+    content += "\n" + reference if reference.present?
+    invoices = supplier.invoices.unpaid.select {|i| content.include? i.number}
```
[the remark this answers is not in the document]
paroga: what do you have in mind exactly?

app/models/bank_transaction.rb
paroga: i want your ideas on this, otherwise i'd just merge it into master ;-)
i'll think about it a little bit more together with the comments in the other issue and post a new PR in the next days

paroga:
> What about un-linkable transactions, will there be a way to link them manually by an admin?
Keep it up :)

That's what #563 was about. ;-)

app/models/bank_transaction.rb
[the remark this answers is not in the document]
wvengen: Perhaps a check that the number of characters matched is at least four (and doesn't include spaces). Or add a validation on the invoice model (I don't know if e.g. invoice number `8` would be something appearing in practice). Do you have any experience on that?

wvengen:
> That's what #563 was about. ;-)

Ah, thanks for pointing that out. Would it be useful to tell the user how many / which transactions could not be linked? I don't have the UI clearly in my head, so perhaps it is clear immediately which transactions still need to be linked.

app/models/bank_transaction.rb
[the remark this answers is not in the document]
paroga: we have single character numbers in our foodcoop. I personally don't see a problem in the current more greedy way, since we check the amount anyway and if there's so clear way we ignore it.

paroga:
> I don't have the UI clearly in my head, so perhaps it is clear immediately which transactions still need to be linked.

You can sort by the link column, so there is an easy way to find unlinked transactions.

app/models/bank_transaction.rb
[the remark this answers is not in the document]
wvengen: Ok then!

app/controllers/finance/bank_accounts_controller.rb
paroga: we don't do this anywhere else. i'd like to stick to existing style

app/models/bank_transaction.rb
paroga: i don't see any problems since only add values

app/models/bank_transaction.rb
paroga: i moved it into a new class

spec/factories/supplier.rb
```diff
@@ -6,6 +6,7 @@
     name { Faker::Company.name.truncate(30) }
     phone { Faker::PhoneNumber.phone_number }
     address { Faker::Address.street_address }
+    iban { Faker::Bank.iban }
```
wvengen: Is iban required for a supplier? If not, then please put it in a trait if you need it. From [FactoryBot docs](https://www.rubydoc.info/gems/factory_bot/file/GETTING_STARTED.md#Defining_factories):
> If you're creating ActiveRecord objects, that means that you should only provide attributes that are required through validations and that do not have defaults. Other factories can be created through inheritance to cover common scenarios for each class.

spec/models/bank_transaction_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      expect(bank_transaction1.assign_to_ordergroup).to be nil
+    end
+
+    it 'ignores transaction with invalid ordergroup' do
+      expect(bank_transaction2.assign_to_ordergroup).to be false
+    end
+
+    it 'ignores transaction with invalid user' do
+      expect(bank_transaction3.assign_to_ordergroup).to be false
+    end
```
wvengen: Test one with a valid user as well?

PR #570, opened 2018-10-13
API-1: prepararations

wvengen:
Part one of #429 in chunks:
* [Doorkeeper](https://github.com/doorkeeper-gem/doorkeeper) OAuth2 provider for API auth
* A Doorkeeper app to manage API keys (_Apps_ button from config screen)
* An API base controller with authentication and error handling
* Split off code common to application and API controllers to a number of concerns
* CORS


app/controllers/api/v1/base_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  # @return [Ordergroup] Current user's ordergroup, or +nil+ if no valid token or user has no ordergroup.
+  def current_ordergroup
+    current_user.try(:ordergroup)
+  end
+
+  def require_ordergroup
+    authenticate
+    current_user.ordergroup or raise Api::Errors::PermissionRequired
```
paroga: what about using `current_ordergroup`? do we use the `something or anotherthing` pattern anywhere else? why not `||`?

app/controllers/concerns/auth.rb
```diff
@@ -0,0 +1,148 @@
...
+    if !current_user
+      # No user at all: redirect to login page.
+      logout
+      session[:return_to] = request.original_url
+      redirect_to_login :alert => I18n.t('application.controller.error_authn')
+    else
+      # We have an authenticated user, now check role...
+      # Roles gets the user through his memberships.
+      hasRole = case role
+      when "admin"  then current_user.role_admin?
```
paroga: can we use single quotes, if we touch the code already?

app/controllers/concerns/locale.rb
```diff
@@ -0,0 +1,51 @@
...
+  def user_settings_language
+    current_user.locale if current_user
+  end
+
+  def session_language
+    session[:locale]
+  end
+
+  def browser_language
+    request.env['HTTP_ACCEPT_LANGUAGE'] ? request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first : nil
```
paroga: is there no better way or at least a gem for this?

app/controllers/concerns/locale.rb
```diff
@@ -0,0 +1,51 @@
...
+  included do
+    before_filter :set_locale
+  end
+
+  def explicitly_requested_language
+    params[:locale]
+  end
+
+  def user_settings_language
+    current_user.locale if current_user
```
paroga: `.try()`?

config/application.rb
```diff
@@ -62,6 +62,14 @@ class Application < Rails::Application
     # Load legacy scripts from vendor
     config.assets.precompile += [ 'vendor/assets/javascripts/*.js' ]
 
+    # CORS for API
+    config.middleware.insert_before 0, "Rack::Cors" do
+      allow do
+        origins '*'
+        # this restricts Foodsoft scopes to certain characters - let's discuss it when it becomes an actual problem
+        resource %r{\A/[-a-zA-Z0-9_]+/api/v1/}, :headers => :any, :methods => :any
```
paroga: use newer syntax `headers: :any`?

config/initializers/doorkeeper.rb
```diff
@@ -0,0 +1,113 @@
...
+  # Under some circumstances you might want to have applications auto-approved,
+  # so that the user skips the authorization step.
+  # For example if dealing with a trusted application.
+  # skip_authorization do |resource_owner, client|
+  #   client.superapp? or resource_owner.admin?
+  # end
+  skip_authorization { true } # right now only admins can add apps, so this is ok
+
+  # WWW-Authenticate Realm (default "Doorkeeper").
+  realm "Foodsoft"
```
paroga: put it into config?

config/application.rb
```diff
@@ -62,6 +62,14 @@ class Application < Rails::Application
     # Load legacy scripts from vendor
     config.assets.precompile += [ 'vendor/assets/javascripts/*.js' ]
 
+    # CORS for API
+    config.middleware.insert_before 0, "Rack::Cors" do
```
paroga: single quotes?

app/controllers/concerns/auth.rb
```diff
@@ -0,0 +1,148 @@
+# Controller concern for authentication methods
```
paroga: I know it's just copy & past code, but it would be nice to do some cleanup if we touch it already - maybe in its own commit by just introducing the concerns

app/controllers/api/v1/base_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def skip_session
+    request.session_options[:skip] = true
+  end
+
+  def not_found_handler(e)
+    # remove where-clauses from error message (not suitable for end-users)
+    msg = e.message.try {|m| m.sub(/\s*\[.*?\]\s*$/, '')} || 'Not found'
```
paroga: is there no cleaner way than this ugly string manipulation?

paroga: Are there no HTTP status code strings definitions in rails? If not, can we put the hardcoded strings like `Not found` at least on top of the file as const defines?

app/controllers/concerns/locale.rb
```diff
@@ -0,0 +1,51 @@
...
+  def session_language
+    session[:locale]
+  end
+
+  def browser_language
+    request.env['HTTP_ACCEPT_LANGUAGE'] ? request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first : nil
+  end
+
+  def default_language
+    FoodsoftConfig[:default_locale] || ::I18n.default_locale
```
paroga: do we need the `::` prefix? We don't use at the other places...

config/locales/en.yml
```diff
@@ -666,6 +666,7 @@ en:
       use_wiki: Enable wiki
       webstats_tracking_code: Tracking code
     tabs:
+      applications: Apps
```
paroga: you can use `Apps` for german too. (long version would be `Anwendungen`, but might be not as clear as the short version)

app/controllers/api/v1/base_controller.rb
wvengen: Regarding `or`, I'd prefer that for flow-statements - [SO](https://stackoverflow.com/questions/2083112).

app/controllers/concerns/locale.rb
wvengen: Interesting, I'll open an issue for that. --> #574

app/controllers/concerns/locale.rb
wvengen: :) Since we're using Ruby 2.3 anyway, why not switch to `current_user&.locale`?

paroga: 👍

app/controllers/api/v1/base_controller.rb
wvengen: We don't really use this pattern, I'll express it differently.

wvengen: Ah, now I understand why I did this: consider the following two pieces of code:
```ruby
raise Api::Errors::PermissionRequired unless current_ordergroup.present?
```
vs.
```ruby
current_ordergroup.present? or raise Api::Errors::PermissionRequired
```

In the former I first read an exception, and then see it only happens in some deviating circumstance. In the latter I first read that an ordergroup is present, and if not, the error is raised.
I think the latter is easier to read, because the main code-path is read first.

An alternative and (in my eyes) clear, but longer, way would be:
```ruby
unless current_ordergroup.present?
   raise Api::Errors::PermissionRequired
end
```

What do you think?

paroga: i personally would prefer `raise Api::Errors::PermissionRequired unless current_ordergroup.present?` or the longer version.
imho this would also make clear that the return value is garbage, which was not clear for me in the first version

app/controllers/concerns/locale.rb
wvengen: My guess would be: this concern can be included at places where there might e.g. be an `I18n` submodule defined, and in that case we'd still want to reference the global `I18n` module.

config/initializers/doorkeeper.rb
wvengen: Good idea.
But then: this is an initializer, at that point the Foodsoft configuration may not be loaded (I'm not fully sure). And worse, it can change every request, so I'd rather keep it static here until the need arises.

app/controllers/api/v1/base_controller.rb
wvengen: Ah, that's indeed a pitfall. I'll adopt the version starting with `raise`.

config/initializers/doorkeeper.rb
paroga: would have been a candidate for single quotes 😉

wvengen: I missed that one! Changed it.
p.s. sorry for spamming you with attribution in the commit and force-pushing it several times ;)

paroga: > sorry for spamming you

since i spam you too often, i'm not in the position of complaining 😃

wvengen:
Thanks for your quick and useful comments! I'll merge it (maybe after the bundle update PR to get the Ruby 2.3 requirement clear, if you're ok (mysql2 had a minor update)).

paroga:
sure! i really like to see it land finally 🎉

PR #571, opened 2018-10-13
API-2: basic endpoints

wvengen:
Part two of #429 in chunks, continued from #570: basic endpoints
* `/api/v1/user`
* `/api/v1/config`
* `/api/v1/navigation`

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -1,0 +1,7 @@
+class Api::V1::UsersController < Api::V1::BaseController
+
+  def show
+    render json: current_user
+  end
+
+end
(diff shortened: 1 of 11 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V1::UsersController < Api::V1::BaseController
+
+  before_action :authenticate
```
paroga: can we make it the default in `BaseController` and just skip it where we don't need it?

app/serializers/config_serializer.rb
```diff
@@ -0,0 +1,30 @@
+class ConfigSerializer < ActiveModel::Serializer
+  attributes :name, :homepage, :contact,
```
paroga: what about sorting them?

PR #573, opened 2018-10-13
API-4: ordering

wvengen:
Part four of #429 in chunks, continued from #572:
* `/api/v1/orders(/:id)`
* `/api/v1/order_articles(/:id)`
* `/api/v1/group_order_articles(/:id)` with CRUD
* `/api/v1/article_categories(/:id)`

Updating single `GroupOrderArticle`s is something (I believe) we don't really do elsewhere. I've looked into this years ago in the foodcoop-adam fork, and used something similar here.

- [x] add docs and specs for endpoints (rebase on updated #572)
- [x] see if update logic can be moved to model
- [x] specs for updating endpoints
- [x] add description to swagger fields
- [x] test group_order_article serializer `total_price` (first check if Apivore might test that)
- [x] rebase

### DIFF
--- a/app/controllers/api/v1/user/group_order_articles_controller.rb
+++ b/app/controllers/api/v1/user/group_order_articles_controller.rb
@@ -1,0 +1,108 @@
+class Api::V1::User::GroupOrderArticlesController < Api::V1::BaseController
+  include Concerns::CollectionScope
+
+  before_action ->{ doorkeeper_authorize! 'group_orders:user' }
+
+  before_action :require_ordergroup
+  before_action :require_minimum_balance, only: [:create, :update] # destroy is ok
+  before_action :require_enough_apples, only: [:create, :update] # destroy is ok
+  # @todo allow decreasing amounts when minimum balance isn't met
+
+  def index
+    render_collection search_scope
+  end
+
+  def show
+    goa = scope.find(params.require(:id))
+    render_goa_with_oa(goa)
+  end
+
+  def create
+    goa = nil
+    GroupOrderArticle.transaction do
+      oa = order_articles_scope_for_create.find(create_params.require(:order_article_id))
+      go = current_ordergroup.group_orders.find_or_create_by!(order_id: oa.order_id)
+      goa = go.group_order_articles.create!(order_article: oa)
+      goa.update_quantities((create_params[:quantity] || 0).to_i, (create_params[:tolerance] || 0).to_i)
+      oa.update_results!
+      go.update_price!
+      go.update_attributes! updated_by: current_user
+    end
+    render_goa_with_oa(goa)
+  end
+
+  def update
+    goa = nil
+    GroupOrderArticle.transaction do
+      goa = scope_for_update.includes(:group_order_article_quantities).find(params.require(:id))
+      goa.update_quantities((update_params[:quantity] || goa.quantity).to_i, (update_params[:tolerance] || goa.tolerance).to_i)
+      goa.order_article.update_results!
+      goa.group_order.update_price!
+      goa.group_order.update_attributes! updated_by: current_user
+    end
+    render_goa_with_oa(goa)
+  end
+
+  def destroy
+    goa = nil
+    GroupOrderArticle.transaction do
+      goa = scope_for_update.find(params.require(:id))
+      goa.destroy!
+      goa.order_article.update_results!
+      goa.group_order.update_price!
+      goa.group_order.update_attributes! updated_by: current_user
+    end
+    render_goa_with_oa(nil, goa.order_article)
+  end
+
+  private
+
+  def max_per_page
+    nil
+  end
+
+  def scope
+    GroupOrderArticle.
+      joins(:group_order).
+      includes(order_article: :article, group_order: :order).
+      where(group_orders: { ordergroup_id: current_ordergroup.id })
+  end
+
+  def scope_for_update
+    scope.references(order_article: { group_order: :order }).merge(Order.open)
+  end
+
+  def order_articles_scope_for_create
+    OrderArticle.joins(:order).merge(Order.open)
+  end
+
+  def create_params
+    params.require(:group_order_article).permit(:order_article_id, :quantity, :tolerance)
+  end
+
+  def update_params
+    params.require(:group_order_article).permit(:quantity, :tolerance)
+  end
+
+  def require_minimum_balance
+    minimum_balance = FoodsoftConfig[:minimum_balance] or return
+    if current_ordergroup.account_balance < minimum_balance
+      raise Api::Errors::PermissionRequired.new(t('application.controller.error_minimum_balance', min: minimum_balance))
+    end
+  end
+
+  def require_enough_apples
+    if current_ordergroup.not_enough_apples?
+      s = t('group_orders.messages.not_enough_apples', apples: current_ordergroup.apples, stop_ordering_under: FoodsoftConfig[:stop_ordering_under])
+      raise Api::Errors::PermissionRequired.new(s)
+    end
+  end
+
+  def render_goa_with_oa(goa, oa = goa.order_article)
+    data = {}
+    data[:group_order_article] = GroupOrderArticleSerializer.new(goa) if goa
+    data[:order_article] = OrderArticleSerializer.new(oa) if oa
+
+    render json: data, root: nil
+  end
+end
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -30,9 +30,9 @@ class Order < ApplicationRecord
 
   # Finders
   scope :started, -> { where('starts <= ?', Time.now) }
-  scope :closed, -> { where(state: 'closed').order('ends DESC') }
-  scope :stockit, -> { where(supplier_id: nil).order('ends DESC') }
-  scope :recent, -> { order('starts DESC').limit(10) }
+  scope :closed, -> { where(state: 'closed').order(ends: :desc) }
+  scope :stockit, -> { where(supplier_id: nil).order(ends: :desc) }
+  scope :recent, -> { order(starts: :desc).limit(10) }
   scope :stock_group_order, -> { group_orders.where(ordergroup_id: nil).first }
   scope :with_invoice, -> { where.not(invoice: nil) }
 
@@ -42,9 +42,9 @@ class Order < ApplicationRecord
   # So orders can
   # 1. ...only transition in one direction (e.g. an order that has been `finished` currently cannot be reopened)
   # 2. ...be set to `closed` when having the `finished` state. (`received` is optional)
-  scope :open, -> { where(state: 'open').order('ends DESC') }
-  scope :finished, -> { where(state: %w[finished received closed]).order('ends DESC') }
-  scope :finished_not_closed, -> { where(state: %w[finished received]).order('ends DESC') }
+  scope :open, -> { where(state: 'open').order(ends: :desc) }
+  scope :finished, -> { where(state: %w[finished received closed]).order(ends: :desc) }
+  scope :finished_not_closed, -> { where(state: %w[finished received]).order(ends: :desc) }
 
   # Allow separate inputs for date and time
   #   with workaround for https://github.com/einzige/date_time_attribute/issues/14
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -51,6 +51,14 @@ class Order < ApplicationRecord
   include DateTimeAttributeValidate
   date_time_attribute :starts, :boxfill, :ends
 
+  def self.ransackable_attributes(auth_object = nil)
+    %w(id state supplier_id starts boxfill ends pickup)
+  end
+
+  def self.ransackable_associations(auth_object = nil)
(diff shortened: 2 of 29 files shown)


app/controllers/api/v1/group_order_articles_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+    goa = scope.find(params.require(:id))
+    render_goa_with_oa(goa)
+  end
+
+  def create
+    GroupOrderArticle.transaction do
+      oa = order_articles_scope.find(create_params.require(:order_article_id))
+      go = current_ordergroup.group_orders.find_or_create_by!(order_id: oa.order_id)
+      goa = go.group_order_articles.create!(order_article: oa)
+      goa.update_quantities((create_params[:quantity] || 0).to_i, (create_params[:tolerance] || 0).to_i)
```
paroga: does `create_params` do not provide any method for getting with an default value?

app/controllers/api/v1/group_order_articles_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+      oa.update_results!
```
paroga: shouldn't the model instead of the controller be responsible for updating `order_articles` and `group_order`?

app/controllers/api/v1/group_order_articles_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  private
+
+  def max_per_page
+    nil
+  end
+
+  def scope
+    current_ordergroup.group_order_articles.
+      preload(:order_article => :article).
```
paroga: `order_article: :article`

app/controllers/api/v1/group_order_articles_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def show
+    goa = scope.find(params.require(:id))
+    render_goa_with_oa(goa)
+  end
+
+  def create
+    GroupOrderArticle.transaction do
+      oa = order_articles_scope.find(create_params.require(:order_article_id))
```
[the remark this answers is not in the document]
wvengen: To have a consistent database, it should not change between fetching data from the database and updating it. Putting the find inside the transaction ensures that. In this particular case of creating a new record, there is little chance that something like that happens and is a problem, but to be sure, I'd rather keep it inside the transaction.
This also means that the order article is only available within the block, hence the render needs to happen inside. Or I could try something like:
```ruby
def create
  goa = GroupOrderArticle.transaction do
      oa = order_articles_scope.find(create_params.require(:order_article_id))
      go = current_ordergroup.group_orders.find_or_create_by!(order_id: oa.order_id)
      goa = go.group_order_articles.create!(order_article: oa)
      goa.update_quantities((create_params[:quantity] || 0).to_i, (create_params[:tolerance] || 0).to_i)
      oa.update_results!
      go.update_price!
      goa
  end
  render_goa_with_oa(goa)
end
```
Is that better?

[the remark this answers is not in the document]
paroga: yes, you're right. it makes sense to keep the `find` in the transactions, but the `render` needs not to be there. both methods for finding `goa` are not 100% beautiful IMHO, so decide which you like more

app/controllers/api/v1/group_order_articles_controller.rb
wvengen: Not that I could find (`fetch` only does it for a single key). Strong parameters is merely about passing and cleaning parameters, not about giving defaults. I think it's fine to explicitely put it here (there could be different behaviour if one is `nil` in other places).

paroga: if you say so, i'm ok with it

app/controllers/api/v1/group_order_articles_controller.rb
wvengen: Yes.
I'm a little fearful touching and refactoring the member ordering update logic. Keeping it separate like this was a first step, without having to deal with the 'classic' member ordering system. Perhaps I'd rather wait until one of the two directions starts to fade, and clean it up for one use-case.

I'll have a look at it some time later, to see if I can wrap my head around it a bit better.

app/controllers/api/v1/group_order_articles_controller.rb
wvengen: For including and preloading, I'd favour the 'old' syntax. 
`preload(:order_article, :article)` 
`preload(order_article: :article)`
`preload(:order_article => :article)`
The latter makes it most clear to me that `:article` is a 'child' of `:order_article`.

wvengen: What do you think?

app/controllers/api/v1/group_order_articles_controller.rb
wvengen: Moved to #579. There are at least three places where group order articles are saved.
I'm somewhat in favour of taking that up separately, if you're ok.

wvengen: I looked it it again, and the moment when `OrderArticle#update_results!` and `GroupOrder#update_price!` are called depends on the action that is taken: if multiple products are updated in a group order, `GroupOrder#update_price!` only needs to be called once at the end (not each time a GroupOrderArticle is modified) - this happens in the regular member ordering screens. And when an OrderArticle is updated among multiple group orders, `#update_results!` only needs to be called once at the end too (we don't do this currently, as far as I know, but it could be possible).

Perhaps ideally I'd have some kind of system that queues `OrderArticle#update_results!` and `GroupOrder#update_price!` for the speciic order article and group order when a GroupOrderArticle is changed, and then at the end of the transaction, all of these queued updates are performed once. We don't have such a system currently. So I think the place for these calls is in the code that requests the GroupOrderArticle updates, which in this case is the controller.

wvengen:
@paroga I think this is ready now, care to review?
(note that the CI error is because of a disappeared commit after a force-push, note that the other CI succeeds)

app/controllers/api/v1/user/group_order_articles_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def create
+    goa = nil
+    GroupOrderArticle.transaction do
+      oa = order_articles_scope_for_create.find(create_params.require(:order_article_id))
+      go = current_ordergroup.group_orders.find_or_create_by!(order_id: oa.order_id)
+      goa = go.group_order_articles.create!(order_article: oa)
+      goa.update_quantities((create_params[:quantity] || 0).to_i, (create_params[:tolerance] || 0).to_i)
+      oa.update_results!
+      go.update_price!
```
wvengen, line 28: set `go.updated_by` to `current_user` like in https://github.com/foodcoops/foodsoft/commit/3d5043fd052f79b33f8cae2ea3a5cc81ee972ba5

wvengen:
I'm slowly proposing to merge this, if there are no further objections.

paroga (changes requested):
maybe i missed something, but expected some additional tests:
* do api calls return the correct content? (at least check the `id` and/or a relevant property like `name`) i only see some tests which validate the content for `GET /orders`
* do api calls perform any work? does a `DELETE` really removes the element from the database and not just returns the expected status code?

we have the chance to improve our test coverage with the new api endpoints, so we should try to write tests for as many relevant cases as possible

app/serializers/order_serializer.rb
```diff
@@ -0,0 +1,11 @@
+class OrderSerializer < ActiveModel::Serializer
+  attributes :id, :name, :starts, :ends, :boxfill, :pickup, :is_open, :is_boxfill
+
+  def is_open
+    object.open?
+  end
+
+  def is_boxfill
+    !!object.boxfill?
```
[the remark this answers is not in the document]
wvengen: no, you're right, thanks!

app/models/order.rb
```diff
@@ -30,12 +30,12 @@ class Order < ApplicationRecord
...
-  scope :finished_not_closed, -> { where(state: 'finished').order('ends DESC') }
-  scope :closed, -> { where(state: 'closed').order('ends DESC') }
-  scope :stockit, -> { where(supplier_id: nil).order('ends DESC') }
-  scope :recent, -> { order('starts DESC').limit(10) }
+  scope :open, -> { where(state: 'open').order('orders.ends DESC') }
+  scope :finished, -> { where("orders.state = 'finished' OR orders.state = 'closed'").order('orders.ends DESC') }
+  scope :finished_not_closed, -> { where(state: 'finished').order('orders.ends DESC') }
+  scope :closed, -> { where(state: 'closed').order('orders.ends DESC') }
+  scope :stockit, -> { where(supplier_id: nil).order('orders.ends DESC') }
+  scope :recent, -> { order('orders.starts DESC').limit(10) }
```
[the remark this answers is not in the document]
wvengen: I got an error in one of the queries when doing this (because `starts` can be ambiguous when joining with a model having the same attribute). In a rebase I've replaced this with `order(starts: :desc)` so that Rails does this automatically.

wvengen:
> - do api calls return the correct content? (at least check the id and/or a relevant property like name) i only see some tests which validate the content for GET /orders

The swagger_spec uses Apivore which checks that all responses are conforming to the Swagger/OpenAPI specification in doc/.

> - do api calls perform any work? does a DELETE really removes the element from the database and not just returns the expected status code?

That's a good point, which I have stepped over perhaps a bit too lightly.

> we have the chance to improve our test coverage with the new api endpoints, so we should try to write tests for as many relevant cases as possible

I agree that is a good opportunity, will write tests for modifying methods.

app/models/order.rb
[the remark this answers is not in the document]
paroga: thx, `order(starts: :desc)` is what i expected

paroga:
> The swagger_spec uses Apivore which checks that all responses are conforming to the Swagger/OpenAPI specification in doc/.

does it also also check if the *correct* items are returned or does it only check for the schema?

app/serializers/order_serializer.rb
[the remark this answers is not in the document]
wvengen: Ah, now I see why: this value isn't expected to be `null` in the API output, and it can be.
I'll change this in the model, so that it always returns a boolean value (which makes sense for a `?` method).

wvengen:
> does it also also check if the correct items are returned or does it only check for the schema?

It only checks for the schema, which is an approach we've already adopted.
I think it wouldn't be a bad idea to move to a different testing approach for the API. We'd probably implement controller specs for all the API endpoints, checking status and return values. This would kind of double with the Apivore specs, which would better be moved to become part of the controller specs (if possible, or find another solution to test OpenAPI conformance, or skip that).
I think that, while it would be good to look into, this overhaul is outside the scope of this PR.

Updating data is new here, I'd be happy to include specs for that.

paroga:
> this overhaul is outside the scope of this PR.

agree

> Updating data is new here, I'd be happy to include specs for that.

thanks

doc/swagger.v1.yml
```diff
@@ -286,6 +628,134 @@ definitions:
         description: when the transaction was entered
     required: ['id', 'user_id', 'user_name', 'amount', 'note', 'created_at']
 
+  ArticleCategory:
+    type: object
+    properties:
+      id:
+        type: integer
+      name:
+        type: string
```
paroga, line 637: do we want to add `description` to all properties?

wvengen: Good idea. Except, perhaps `id` and `name` (which are self-explanatory).

app/serializers/group_order_article_serializer.rb
```diff
@@ -0,0 +1,9 @@
+class GroupOrderArticleSerializer < ActiveModel::Serializer
+  attributes :id, :order_article_id
+  attributes :quantity, :tolerance, :result, :total_price
+
+  def total_price
+    # make sure BigDecimal is serialized as a number
```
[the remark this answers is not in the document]
wvengen: Apivore tests this (since a BigDecimal is serialized as a string) - it breaks when I remove `to_f`.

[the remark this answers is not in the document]
paroga: i'd suggest to remove the comment in that case.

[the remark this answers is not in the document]
wvengen: It explains why `to_f` is there. But since it is tested, I can live with that.

spec/api/v1/user/group_order_articles_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+      include_examples "group_order_articles destroy success"
+    end
+
+    context 'without enough apple points' do
+      before { allow_any_instance_of(Ordergroup).to receive(:not_enough_apples?).and_return(true)  }
+      include_examples "group_order_articles destroy success"
+    end
+
+  end
+end
```
paroga, line 212: maybe you could a newline at the end of the file

paroga (approved):
LGTM (beside the missing newline)
many thanks for the additional tests!

PR #577, opened 2018-10-15
API-2.5: documentation and specs

wvengen:
As surfaced in #576: add documentation and specs for basic endpoints, as well as a fix for Doorkeeper 5.
Will add specs in #572 and #573.

### DIFF
--- a/db/migrate/20181013195028_add_confidential_to_applications.rb
+++ b/db/migrate/20181013195028_add_confidential_to_applications.rb
@@ -1,0 +1,5 @@
+class AddConfidentialToApplications < ActiveRecord::Migration
+  def change
+    add_column :oauth_applications, :confidential, :boolean, null: false, default: true
+  end
+end
--- a/doc/swagger.v1.yml
+++ b/doc/swagger.v1.yml
@@ -1,0 +1,67 @@
+swagger: '2.0'
+info:
+  title: Foodsoft API v1
+  version: '1.0.0'
+  description: >
+    [Foodsoft](https://github.com/foodcoops/foodsoft) is web-based software to manage
+    a non-profit food coop (product catalog, ordering, accounting, job scheduling).
+
+
+    This is a description of Foodsoft's API v1.
+
+
+    Note that each food cooperative typically has their own instance (on a shared
+    server or their own installation), and there are just as many APIs (if the Foodsoft
+    version is recent enough).
+    This API description points to the default development url with the default
+    Foodsoft scope - that would be [http://localhost:3000/f](http://localhost:3000/f).
+externalDocs:
+  description: General Foodsoft API documentation
+  url: https://github.com/foodcoops/foodsoft/blob/master/doc/API.md
+
+# development url with default scope
+host: localhost:3000
+schemes:
+  - 'http'
+basePath: /f/api/v1
+
+produces:
+  - 'application/json'
+
+paths:
+
+definitions:
+  Error:
+    type: object
+    properties:
+      error:
+        type: string
+        description: error code
+      error_description:
+        type: string
+        description: human-readable error message (localized)
+  Error404:
+    type: object
+    properties:
+      error:
+        type: string
+        description: '<tt>not_found</tt>'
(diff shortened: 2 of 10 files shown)


doc/swagger.v1.yml
```diff
@@ -0,0 +1,158 @@
...
+  /config:
+    get:
+      summary: configuration variables
+      tags:
+        - 7. General
+      responses:
+        200:
+          description: success
+          schema:
+            type: object
```
wvengen: This is pretty general right now. Do you think it makes sense to include all config keys here? It's a little work, and the configuration list in the backend is much more extensive. Or does a rake task that copies config key descriptions here make sense at some point ... ?

doc/swagger.v1.yml
```diff
@@ -0,0 +1,158 @@
...
+  /navigation:
+    get:
+      summary: navigation
+      tags:
+        - 7. General
+      responses:
+        200:
+          description: success
+          schema:
+            type: object
```
wvengen: I guess it would be nice to specify this in the definitions, still.

doc/swagger.v1.yml
paroga: do we have a full list of all keys? Doesn't it depend on the activated plugins?

doc/swagger.v1.yml
paroga: same as with the previous point.

wvengen: Added this, the form is expected to remain the same (array of objects with name+url or name+items).

doc/swagger.v1.yml
wvengen: That's a good point, actually.

db/migrate/20181013195028_add_confidential_to_applications.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
paroga: do we really need this here?

db/migrate/20181013195028_add_confidential_to_applications.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class AddConfidentialToApplications < ActiveRecord::Migration
+  def change
+    add_column(
```
paroga: can you use the style of the other migration files here too? (one line, no brackets)

spec/support/api_helper.rb
```diff
@@ -0,0 +1,18 @@
+module ApiHelper
+  extend ActiveSupport::Concern
+
+  included do
+    let(:user) { create :user }
+    let(:access_token) { create(:oauth2_access_token, resource_owner_id: user.id).token }
+    let(:authorization) { "Bearer #{token}" }
```
paroga: is this used? `token` vs `access_token`?

spec/support/api_helper.rb
```diff
@@ -2,17 +2,17 @@ module ApiHelper
   extend ActiveSupport::Concern
 
   included do
-    let(:user) { create :user }
-    let(:access_token) { create(:oauth2_access_token, resource_owner_id: user.id).token }
-    let(:authorization) { "Bearer #{token}" }
+    let(:user) { create(:user) }
+    let(:api_access_token) { create(:oauth2_access_token, resource_owner_id: user.id).token }
```
paroga: why not do that already correct in the previous commit?

db/migrate/20181013195028_add_confidential_to_applications.rb
wvengen: This migration was generated by the doorkeeper upgrade procedure, from [this template](https://github.com/doorkeeper-gem/doorkeeper/blob/3b61b6c3e99a309c8e5e25f86ee6ac2c8131b05d/lib/generators/doorkeeper/templates/add_confidential_to_applications.rb.erb). I agree that we'd better use our own style here.

wvengen:
After addressing the issues, I think it's ok to merge this, and intend to do so after a week.

PR #584, opened 2018-11-12
store the supplier's case price 

carchrae:
the current assumption in foodsoft is that we will do all accounting based on the per article price * units.  this leads to rounding errors that can make reconciling accounts more difficult.

in our fork we store the supplier's original price as well as the member's per-unit price.  this allows us to track the rounding errors and makes it easier for us to reconcile.

there are several model + UI changes that go with this one, however, i am submitting this PR as a low level db change because it will have no impact to existing users.  i will submit the other changes but need to think about how to do that in a way that doesn't disrupt any existing users.  this change should not impact any users (aside from adding a field to two tables in the db)


db/schema.rb
```diff
@@ -67,25 +69,25 @@
   add_index "assignments", ["user_id", "task_id"], name: "index_assignments_on_user_id_and_task_id", unique: true, using: :btree
 
   create_table "bank_accounts", force: :cascade do |t|
-    t.string   "name",                                                           null: false
-    t.string   "iban"
-    t.string   "description"
-    t.decimal  "balance",                   precision: 12, scale: 2, default: 0, null: false
+    t.string   "name",                      limit: 255,                                      null: false
```
carchrae: all of these changes were added when i applied `rake db:migrate` - they are a bit concerning.

carchrae: perhaps name 'limit 255' is not as worrying, but see later ones

carchrae: maybe they are not worrying.  but i am confused by why they are different.

paroga: seams that my `db:migrate` created a different `schema.rb`. i usually use SQLite. so that's nothing you did wrong. you just see now every "wrong" database migration I did over the last years 😀 i should be no problem, but to keep history clean (one commit does only one thing) just remove the lines from the commit and everything should be ok

db/migrate/20181111160726_add_supplier_price_to_articles.rb
```diff
@@ -0,0 +1,5 @@
+class AddSupplierPriceToArticles < ActiveRecord::Migration
+  def change
+    add_column :articles, :supplier_price, :decimal, :precision => 8, :scale => 2
```
paroga: maybe you could use the newer syntax `precision: 8`?

db/schema.rb
carchrae: ah!  i think you are correct, because this was generated from running against mysql.

i hate commits that show unrelated changes, so i will regenerate schema.rb with sqllite.  if we want to clean it up and generate against mysql, that should be another PR.

paroga:
i'd prefer to see at least a PR with all changes, before committing anything. I'm ok with incremental commits, especially model changes, which can be annoying to maintain downstream, but we should have at least a full change here, so that at least somebody could pick it up later in case of some problems

carchrae:
sure, i get that and we can hold on this PR because what is the point in adding a few fields if it doesn't add any feature/function to the user.  i'll see what i can do, but it may become a rather large feature.  that said, i think it this is a common use case, so i will put some time into it.

to give you an idea, in our version, when you edit the price of a case, it will also have a ui that lets you recalculate the 'per unit' price.  and on the order accounting page it shows you the rounding error so you can try to get the total to exactly match the supplier invoice.  my goal is to have it entirely accounted, so that `supplier invoice total = member charges + rounding error`  (or some variant, but always every cost is accounted for)

wvengen:
Thanks for submitting back changes!
I think it would be useful to include a yardoc comment in the `ArticlePrice` model explaining the use of this field (like you're doing in this PR).

PR #589, opened 2018-12-18
Fix create_collection in FinancialTransactionsController

paroga:


app/controllers/finance/financial_transactions_controller.rb
```diff
@@ -55,7 +55,7 @@ def new_collection
 
   def create_collection
     raise I18n.t('finance.financial_transactions.controller.create_collection.error_note_required') if params[:note].blank?
-    type = FinancialTransactionType.find_by_id(params.permit(:type))
+    type = FinancialTransactionType.find_by_id(params[:type])
```
carchrae: i need to look the form that is sending this, but `find_by_id` and then looking up `type` seems strange (may well be fine)

however, bypassing the `permit` seems a bit dirty.  (i'm not a stalwart on this, but wonder why not allow it as a valid param instead of bypassing `permit`)

paroga: since we are in the `FinancialTransactionsController` and `FinancialTransactionType` is associated to `FinancialTransaction` it should be ok to look up by `type`. 😃 

bypassing `permit` is only relevant if we would pass the whole `params`, since it returns a hash of permitted parameter. but since we need only one, there should be no problem. `params[:type]` should be equal to `params.permit(:type)[:type]`, which seams more strange to me 😉

carchrae: perhaps it is too late, but then it is a shame that `:type` is not `:id` - or perhaps `:type_id`

small points, but things that can trip up people less familiar with the code.  i would say the bug you are fixing was even caused by this very same sloppiness.

PR #594, opened 2018-12-18
Improve Docker setup for multi_coop setup

paroga:


Dockerfile
```diff
@@ -37,15 +37,18 @@ RUN export DATABASE_URL=mysql2://localhost/temp && \
     apt-get install -y mysql-server && \
     /etc/init.d/mysql start && \
     cp config/app_config.yml.SAMPLE config/app_config.yml && \
+    cp config/database.yml.MySQL_SAMPLE config/database.yml && \
```
wvengen: We have `DATABASE_URL`, right - why don't we use that?

Dockerfile
```diff
@@ -37,15 +37,18 @@ RUN export DATABASE_URL=mysql2://localhost/temp && \
...
     rm -Rf config/app_config.yml tmp/* && \
     /etc/init.d/mysql stop && \
     rm -Rf /run/mysqld /tmp/* /var/tmp/* /var/lib/mysql /var/log/mysql* && \
     apt-get purge -y --auto-remove mysql-server && \
     rm -Rf /var/lib/apt/lists/* /var/cache/apt/*
 
-# Make relevant dirs writable for app user
+# Make relevant dirs and files writable for app user
 RUN mkdir -p tmp && \
+    chown nobody config/app_config.yml && \
```
wvengen: It's too bad we have to do that, it removes one security barrier. But I think it's ok.

Dockerfile
```diff
@@ -37,15 +37,18 @@ RUN export DATABASE_URL=mysql2://localhost/temp && \
...
+    chown nobody db/schema.rb && \
```
wvengen: Hmmm when did you run into this not being writable?

Dockerfile
paroga: this way we can use `multicoops:run` again, which becomes very handy for migrations

Dockerfile
paroga: since it's only read during startup i see no problem

Dockerfile
paroga: all the time when i do migrations. if i migrate a singe coop it's ok since the error is at the end, but when i want to migrate multiple coops it stops after the first one

Dockerfile
wvengen: That makes sense, good!

wvengen:
It looks like the Docker build is failing:
```
Mysql2::Error: Specified key was too long; max key length is 767 bytes: CREATE UNIQUE INDEX `index_article_categories_on_name` USING btree ON `article_categories` (`name`) 
```
Could this be related to the changed database setting?
Perhaps https://github.com/diaspora/diaspora/issues/4312 or https://github.com/dhanasingh/redmine_wktime/issues/122 could give a clue.

paroga:
@wvengen it seams that some of the test are too flaky :disappointed: i have the problem also sometimes on my machine. can you check that?

PR #596, opened 2018-12-20
Show private messages in foodsofts index pages

paroga:


db/migrate/20181201000100_create_message_recipients.foodsoft_messages.rb
```diff
@@ -0,0 +1,44 @@
...
+    add_index :message_recipients, [:user_id, :read_at]
+
+    Message.all.each do |m|
+      recipients = YAML.load(m.recipients_ids).map do |r|
+        {
+          message_id: m.id,
+          user_id: r,
+          email_state: :sent
+        }
+      end
```
wvengen: When referencing models from a migration, it is important to define them first. In a future version of Foodsoft this model may have been removed, even though it is still used in this migration. So inside the class, at the top, put `class Message < ActiveRecord::Base; end`. Since we only use database fields here, that will do.

plugins/messages/app/models/message.rb
```diff
@@ -4,44 +4,46 @@ class Message < ActiveRecord::Base
...
     @send_method ||= 'recipients'
   end
 
   before_create :create_salt
-  before_validation :clean_up_recipient_ids, :on => :create
+  before_validation :create_message_recipients, on: :create
+
+  def create_message_recipients
+    users = @additional_recipients
+    users += User.undeleted if send_method == 'all'
```
wvengen: This can be many objects, I'd work with `id`s here.

paroga: are you sure there is now a difference to the previous implementation? There we called `User.undeleted` too...

db/migrate/20181201000100_create_message_recipients.foodsoft_messages.rb
paroga: our migration is already broken in that direction. IMHO we should only support migration from one release to next release.
e.g.: I'd like to change our wiki engine in the future. for the migrations i'll need the old engine to parse the pages, but it don't want to keep it until the end of time as dependency. i'd jut add an error message into the migration, when we remove the dependency, directing the user to migrate to a previous version first

plugins/messages/app/models/message.rb
wvengen: That's right. It should have said `.pluck(:id)`.
I you can make it, please do so, otherwise it's not a step backward.

db/migrate/20181201000100_create_message_recipients.foodsoft_messages.rb
wvengen: Hmm that is a good point, about the wiki.
Still I'd rather keep migration compatibility where possible. I agree that we can support migration from one major version to another, but it would still be nice if it would work across multiple versions for most of the time (and otherwise certainly noted in the release notes).

paroga: There is a dependency on `message_recipients` in line 36, so a simple `ActiveRecord::Base` is not enough

wvengen: Then it would be
```ruby
Class Message < ActiveRecord::Base
  has_many :message_recipients
  has_many :recipients, through: :message_recipients, source: :user
end
```

paroga: makes sense 😄

plugins/messages/app/models/message.rb
paroga: changed it to id usage only

paroga:
can you make a last check, since the last change wasn't a minor one?

plugins/messages/app/models/message.rb
```diff
@@ -4,44 +4,46 @@ class Message < ApplicationRecord
...
+    joins(:message_recipients).where('private = ? OR message_recipients.user_id = ?', false, user.try(&:id)).distinct
   }
 
-  validates_presence_of :recipients_ids, :subject, :body
+  validates_presence_of :message_recipients, :subject, :body
   validates_length_of :subject, :in => 1..255
-  validates_inclusion_of :email_state, :in => EMAIL_STATE.values
 
   after_initialize do
+    @recipients_ids ||= []
```
wvengen: why is this necessary?

plugins/messages/app/models/message.rb
```diff
@@ -4,44 +4,46 @@ class Message < ApplicationRecord
...
-    add_recipients Order.find(order_id).users_ordered if send_method == 'order'
-    self.recipients_ids = recipients_ids.uniq.reject { |id| id.blank? } unless recipients_ids.nil?
-    self.recipients_ids = User.undeleted.collect(&:id) if send_method == 'all'
+    users.uniq.each do |user|
+      recipient = MessageRecipient.new message: self, user_id: user
+      message_recipients << recipient
+    end
   end
 
   def add_recipients(users)
```
wvengen: maybe better to call this `user_ids`?

plugins/messages/app/models/message.rb
```diff
@@ -4,44 +4,46 @@ class Message < ApplicationRecord
...
 
   before_create :create_salt
-  before_validation :clean_up_recipient_ids, :on => :create
+  before_validation :create_message_recipients, on: :create
+
+  def create_message_recipients
+    users = @recipients_ids
+    users += User.undeleted.pluck(:id) if send_method == 'all'
+    users += Group.find(group_id).users.pluck(:id) unless group_id.blank?
+    users += Order.find(order_id).users_ordered.pluck(:id) if send_method == 'order'
```
wvengen: maybe better to call this `user_ids`?

db/migrate/20181201000100_create_message_recipients.foodsoft_messages.rb
```diff
@@ -0,0 +1,49 @@
+class CreateMessageRecipients < ActiveRecord::Migration
+  class Message < ActiveRecord::Base
+    has_many :message_recipients
+    has_many :recipients, through: :message_recipients, source: :user
+  end
+
+  def up
+    create_table :message_recipients do |t|
+      t.references :message, index: true, null: false
+      t.references :user, null: false
```
wvengen: `add_reference` [mentions](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference):
> `:index` defaults to true

Maybe add `index: false`?

db/migrate/20181201000100_create_message_recipients.foodsoft_messages.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    add_index :message_recipients, [:user_id, :read_at]
+
+    Message.all.each do |m|
+      recipients = YAML.load(m.recipients_ids).map do |r|
+        {
+          message_id: m.id,
+          user_id: r,
+          email_state: :sent
```
wvengen: wouldn't we get `email_state` from the message?

db/migrate/20181201000100_create_message_recipients.foodsoft_messages.rb
```diff
@@ -0,0 +1,49 @@
...
+  def down
+    change_table :messages do |t|
+      t.text :recipients_ids
+      t.integer :email_state, default: 0, null: false
+    end
+
+    messages = Message.all.includes(:message_recipients).map do |m|
+      recipients = m.message_recipients.map(&:user_id)
+      m.recipients_ids = recipients.to_yaml
+      m.email_state = 1
```
wvengen: maybe good to add a comment what the meaning of `1` is here (in case the enum would migrate later)

db/migrate/20181201000100_create_message_recipients.foodsoft_messages.rb
```diff
@@ -0,0 +1,49 @@
+class CreateMessageRecipients < ActiveRecord::Migration
+  class Message < ActiveRecord::Base
+    has_many :message_recipients
+    has_many :recipients, through: :message_recipients, source: :user
+  end
```
wvengen: We'd also need to add the `MessageRecipient` class. I'm not sure if `Recipient` is necessary here, since we don't reference it below (might even be able to remove it here altogether, since we're dealing with ids only in the migration).

db/migrate/20181201000100_create_message_recipients.foodsoft_messages.rb
paroga: on my machine it seams that the default is `false`. at least the index is missing if there is not `index: true`

wvengen: :+1: that must have changed then after Rails 4

PR #597, opened 2018-12-20
Show associated group when sending messages (fixes #559)

### LINKED ISSUE #559: Messages: Show associated groups.
Composers write messages in awareness of the associated group, but this information is missing in the notifications. Sometimes messages are difficult to understand without their context. It would be nice to show this information to the readers, for example in the message footer.

paroga:


paroga:
i don't see a much better place. IMHO it's like the footer in mailing list: e.g. "this mail was sent to xy@example.com because your subscribed to something"

PR #598, opened 2018-12-21
Allow deletion of financial transactions and small usability improvem…

paroga:
…ents

### DIFF
--- a/app/models/financial_transaction.rb
+++ b/app/models/financial_transaction.rb
@@ -5,11 +5,14 @@ class FinancialTransaction < ApplicationRecord
   belongs_to :user
   belongs_to :financial_link
   belongs_to :financial_transaction_type
+  belongs_to :reverts, class_name: 'FinancialTransaction', foreign_key: 'reverts_id'
+  has_one :reverted_by, class_name: 'FinancialTransaction', foreign_key: 'reverts_id'
 
   validates_presence_of :amount, :note, :user_id, :ordergroup_id
   validates_numericality_of :amount, greater_then: -100_000,
     less_than: 100_000
 
+  scope :visible, -> { joins('LEFT JOIN financial_transactions r ON financial_transactions.id = r.reverts_id').where('r.id IS NULL').where(reverts: nil) }
   scope :without_financial_link, -> { where(financial_link: nil) }
 
   localize_input_of :amount
@@ -21,6 +24,21 @@ class FinancialTransaction < ApplicationRecord
   # Use this save method instead of simple save and after callback
   def add_transaction!
     ordergroup.add_financial_transaction! amount, note, user, financial_transaction_type
+  end
+
+  def revert!(user)
+    transaction do
+      rt = dup
+      rt.amount = -rt.amount
+      rt.reverts = self
+      rt.user = user
+      rt.save!
+      ordergroup.update_balance!
+    end
+  end
+
+  def hidden?
+    reverts.present? || reverted_by.present?
   end
 
   protected
--- a/app/models/ordergroup.rb
+++ b/app/models/ordergroup.rb
@@ -84,8 +84,7 @@ class Ordergroup < Group
     transaction do
       t = FinancialTransaction.new(ordergroup: self, amount: amount, note: note, user: user, financial_transaction_type: transaction_type, financial_link: link)
       t.save!
-      self.account_balance = financial_transactions.sum('amount')
-      save!
+      update_balance!
       # Notify only when order group had a positive balance before the last transaction:
       if t.amount < 0 && self.account_balance < 0 && self.account_balance - t.amount >= 0
         Resque.enqueue(UserNotifier, FoodsoftConfig.scope, 'negative_balance', self.id, t.id)
@@ -101,6 +100,10 @@ class Ordergroup < Group
 
     @readonly = false # Dirty hack, avoid getting RecordReadOnly exception when called in task after_save callback. A rails bug?
     update_attribute(:stats, {:jobs_size => jobs, :orders_sum => orders_sum})
+  end
+
+  def update_balance!
+    update_attribute :account_balance, financial_transactions.sum('amount')
   end
 
   def avg_jobs_per_euro
--- a/app/views/finance/financial_transactions/_transactions.html.haml
+++ b/app/views/finance/financial_transactions/_transactions.html.haml
@@ -1,4 +1,5 @@
 - with_ordergroup = local_assigns[:with_ordergroup]
+- with_hidden = local_assigns[:with_hidden]
 - with_csv = local_assigns[:with_csv]
 .pull-right
   - if with_csv
@@ -22,9 +23,11 @@
       - FinancialTransactionClass.sorted.each do |c|
         %th
           = sort_link_helper c.display, "amount"
+      - if with_hidden
+        %th
   %tbody
     - @financial_transactions.each do |t|
-      %tr
+      %tr{class: "#{'deleted_row' if t.hidden?}"}
         %td
           - if t.financial_link
             = link_to format_time(t.created_on), finance_link_path(t.financial_link)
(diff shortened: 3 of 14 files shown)


app/views/finance/financial_transactions/index.html.haml
```diff
@@ -8,7 +8,18 @@
...
+    - unless FoodsoftConfig[:charge_members_manually]
+      %p
+        %b= heading_helper(Ordergroup, :available_funds) + ':'
+        = number_to_currency(@ordergroup.get_available_funds())
+    - if FinancialTransactionClass.has_multiple_classes
+      - FinancialTransactionClass.sorted.each do |c|
+        %p
+          %b= c.display + ':'
+          = number_to_currency(@ordergroup["sum_of_class_#{c.id}"])
+
```
carchrae: is this the usability improvement?  it would be nice if it has some explanation on what it is improving.

paroga: it just adds the sums for every financial transaction class

paroga:
i had the same concerns and that's the reason why there is a "show deleted transaction" function. this allows the admin to have the full log of what happened, but the members only get a "clear" transaction history. many people came with this request to me and i always explained why i don't like to make the transactions deletable, but they mostly insisted on it. having a reverse transaction results i e.g. two useless transactions, which confuse the members when reading their financial transaction history...

carchrae:
sorry for being picky about this; it is more an argument on the principle of a statement being a permanent ledger of all adjustments (including errors and repairs) to the account.

paroga:
in our setup we do not use the integrated balancing. all our consumed stuff is charged manually. so people ALWAYS enter transaction manually and there happen a lot of typos, since the people entering them often do not care a lot. i had the same argument, that my bank will never change existing transactions, but they replied that Foodsoft is not a bank :-)
I 100% agree with your points, but i often get the feature reuest, that people want to be able to edit wrong financial transaction. Editing them (without proper history tracking) didn't sound as a good idea to me, so I came up with a way to hide/ignore/delete existing transactions and allow people to enter new ones. would you be more ok with this, if it's behind a config option an/or renaming it to e.g. hide/ignore? calculating the balance is behind some dedicated methods anyway, so i don't see a big problems with bugs. do you have a better idea how to implement this highly requested feature?

carchrae:
> Foodsoft is not a bank :-)

haha.  yes, it is just something that manages an account balance and payments/charges.  nothing at all the same there!  :roll_eyes:   i think the standard/trust in food-coops is very low/high for finances, and it bothers me.  the main issue there isn't so much people being evil (though that could happen), but making mistakes (or not noticing) with other people's money.  when our coop moved to foodsoft it was a million times better than the previous accounting (which was an excel sheet on one persons computer)


>  would you be more ok with this, if it's behind a config option an/or renaming it to e.g. hide/ignore? calculating the balance is behind some dedicated methods anyway, so i don't see a big problems with bugs. do you have a better idea how to implement this highly requested feature?

here is one idea (and something i have meant to implement too - in my screen it is only the text that indicates what the revert relates to).  when you create a 'revert' transaction - which is just a transaction for `0-amount` of the original transaction, add a reference between the transactions.  it should be a 1-1 relation, original transaction<->revert transaction.  (i guess revert `belongs_to` original transaction but i don't know if it is rails abuse to use `belongs_to` to point to the same model)

if people want to hide reverted transactions, then you can filter the list of transactions to those where `transaction.reverted_by: nil or transaction.reverts: nil` - or something like that.

the above is the main gist anyway - basically a transaction can be linked (1-1) to a revert transaction.  it is like `deleted` but it doesn't remove it from the transaction chain, leaving a full audit if needed.  you can then use that link (or lack of link) to hide it for people who don't want to see the mistakes.

i think it is very similar to what you have proposed, but does it via an extra transaction.  i suppose my main argument is that this keeps the property that the sum of all transactions equals the balance and the transaction model simpler.  that said, i can see how filtering reverted transactions is going to end up looking pretty similar.  

let me think a bit more - thank you for explaining your use case and users requests.  i'll also try and make my 'revert settled order' into a PR.  i am trying hard to get my fork back onto origin/master.

wvengen:
Thanks for the PR, and the good discussion around it. I like @carchrae's approach with an extra transaction. Then perhaps users can indeed toggle whether to show all transactions or hide the corrected+correction transactions (like this PR, but then not in terms of deletion - the sum of these by-default-hidden transactions is then always zero).

paroga:
i've changed the data model

app/models/financial_transaction.rb
```diff
@@ -5,11 +5,14 @@ class FinancialTransaction < ApplicationRecord
   belongs_to :user
   belongs_to :financial_link
   belongs_to :financial_transaction_type
+  belongs_to :reverts, class_name: 'FinancialTransaction', foreign_key: 'reverts'
+  has_one :reverted_by, class_name: 'FinancialTransaction', foreign_key: 'reverts'
```
wvengen: What about naming this foreign key `reverts_id` in the database?

app/models/ordergroup.rb
```diff
@@ -36,6 +36,7 @@ def self.include_transaction_class_sum
     select(columns.join(', '))
       .joins('LEFT JOIN financial_transactions ON groups.id = financial_transactions.ordergroup_id')
       .joins('LEFT JOIN financial_transaction_types ON financial_transaction_types.id = financial_transactions.financial_transaction_type_id')
+      .where(financial_transactions: {reverts: nil})
```
wvengen: I don't quite get why this added where-clause is necessary.

app/views/finance/financial_transactions/_transactions.html.haml
```diff
@@ -40,3 +43,8 @@
           %td.numeric{style: 'width:5em'}
             - if t.financial_transaction_type.financial_transaction_class == c
               = format_currency t.amount
+        - if with_hidden
+          %td
+            - unless t.hidden?
+              = link_to t('ui.delete'), finance_ordergroup_transaction_path(t.ordergroup, t), method: :delete,
+                data: {confirm: t('ui.confirm_delete', name: t.note)}, class: 'btn btn-danger btn-mini'
```
wvengen: Great to only show the option to revert a transaction when hidden ones are shown.

Shouldn't we call this button _revert_ instead of _delete_?

config/locales/en.yml
```diff
@@ -957,6 +961,7 @@ en:
         remove_group: Remove group
       transactions_search:
         search_placeholder: Search note ...
+        show_deleted: Show deleted transactions
```
wvengen: These are all still old translations, right?

app/models/financial_transaction.rb
paroga: done

app/models/ordergroup.rb
paroga: some code from an older revision 😨

config/locales/en.yml
paroga: yup

app/views/finance/financial_transactions/_transactions.html.haml
wvengen: I'd like to convey to the user that one really can't 'delete' transactions, but one can undo their effects. Perhaps it's not that common to have a single word for that, like _revert_ or _reverse_. In Dutch it would be _terugdraaien_. Or use an undo icon, perhaps (perhaps with a tooltip, but there's also a confirmation dialog that would explain it)?

carchrae: i vote for "undo" - it is bad english, but a known computer term.  revert is technically better, but users aren't all technical.

paroga: i like the idea of an icon, which allows us to use a longer text: "Revert" (en), "Rückgängig machen" (de) "Terugdraaien" (nl)

paroga: i'v changed it to an icon, with a descriptive tooltip

app/models/ordergroup.rb
```diff
@@ -84,8 +84,7 @@ def add_financial_transaction!(amount, note, user, transaction_type, link = nil)
     transaction do
       t = FinancialTransaction.new(ordergroup: self, amount: amount, note: note, user: user, financial_transaction_type: transaction_type, financial_link: link)
       t.save!
-      self.account_balance = financial_transactions.sum('amount')
-      save!
+      update_balance!
```
wvengen: better :+1:

PR #599, opened 2018-12-21
Add CSV download for ordergroups

paroga:


paroga:
i agree that server code is not the most maintainable solution these days, but it's what we have now and i didn't want to introduce new methods without updating the old ones

PR #600, opened 2018-12-21
Add FinacialTransaction create_collection for ordergroup custom fields

paroga:
If we store a different membership fee for every ordergroup in a custom field, we can now populate the financial transactions with these values.

### DIFF
--- a/app/controllers/finance/financial_transactions_controller.rb
+++ b/app/controllers/finance/financial_transactions_controller.rb
@@ -58,6 +58,14 @@ class Finance::FinancialTransactionsController < ApplicationController
   end
 
   def new_collection
+    @ordergroups = {}
+    Ordergroup.undeleted.order(:name).map do |ordergroup|
+      obj = { name: ordergroup.name }
+      Ordergroup.custom_fields.each do |field|
+        obj[field[:name]] = ordergroup.settings.custom_fields[field[:name]]
+      end
+      @ordergroups[ordergroup.id] = obj
+    end
   end
 
   def create_collection
--- a/app/views/finance/financial_transactions/new_collection.html.haml
+++ b/app/views/finance/financial_transactions/new_collection.html.haml
@@ -3,6 +3,7 @@
 - content_for :javascript do
   :javascript
     var ordergroup = $($.parseHTML("#{escape_javascript(render('ordergroup'))}"));
+    var ordergroups = #{raw @ordergroups.to_json};
 
     $(function() {
       $(document).on('touchclick', 'a[data-remove-transaction]', function() {
@@ -16,16 +17,27 @@
       });
 
       $(document).on('touchclick', 'a[data-add-all-ordergroups]', function() {
-        var value = prompt("#{escape_javascript(heading_helper(FinancialTransaction, :amount))}:");
-        if (value === null)
-          return false;
+        var customField = $(this).data('custom-field');
+        var value;
+        if (!customField) {
+          value = prompt("#{escape_javascript(heading_helper(FinancialTransaction, :amount))}:");
+          if (value === null) {
+            return false;
+          }
+        }
         $('#ordergroups > tbody > tr').remove();
-        var options = ordergroup.find('td > select > option').each(function() {
+        for (var id in ordergroups) {
+          if (!ordergroups.hasOwnProperty(id)) {
+            continue;
+          }
+          if (customField) {
+            value = ordergroups[id][customField];
+          }
           var row = ordergroup.clone();
           row.find('td > input').val(value);
-          row.find('td > select').val(this.value);
+          row.find('td > select').val(id);
           row.appendTo('#ordergroups');
-        });
+        }
         return false;
       });
     });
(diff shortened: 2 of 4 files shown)


app/models/ordergroup.rb
```diff
@@ -39,6 +39,12 @@ def self.include_transaction_class_sum
       .group('groups.id')
   end
 
+  def self.custom_fields
+    fields = FoodsoftConfig[:custom_fields] && FoodsoftConfig[:custom_fields][:ordergroup]
```
carchrae: it seems like this would be a common check required - why not move the logic inside `FoodsoftConfig`

part of me wonders why not go with `FoodsoftConfig[:ordergroup]` and have it looked up by missing_method.  but this is a late comment.  

i would still recommend a specific getter for custom fields so you don't have to do the nil check.  (or alternatively, populate `FoodsoftConfig[:custom_fields]={}` so you can avoid these checks and could have the more terse `return [] unless FoodsoftConfig[:custom_fields][:ordergroup]`

carchrae: indeed, if you are going with the `ApplicationRecord` pattern, then `self.custom_fields` seems like it would be better put in there (adding custom_fields support to all models)

app/views/finance/financial_transactions/new_collection.html.haml
```diff
@@ -16,16 +17,23 @@
       });
 
       $(document).on('touchclick', 'a[data-add-all-ordergroups]', function() {
-        var value = prompt("#{escape_javascript(heading_helper(FinancialTransaction, :amount))}:");
-        if (value === null)
-          return false;
+        var customField = $(this).data('custom-field');
```
carchrae: so `customField` is the name of the field to use.  

how does `custom-field` get set?  (in which code)

app/controllers/finance/financial_transactions_controller.rb
```diff
@@ -51,6 +51,14 @@ def create
...
 
   def new_collection
+    @ordergroups = {}
+    Ordergroup.undeleted.order(:name).map do |ordergroup|
+      obj = { name: ordergroup.name }
+      Ordergroup.custom_fields.each do |field|
+        obj[field[:name]] = ordergroup.settings.custom_fields[field[:name]]
+      end
+      @ordergroups[ordergroup.id] = obj
+    end
```
carchrae: this code isn't really very easy to read.  i would suggest at least some comments on what is going on, or perhaps putting all the `custom_fields` stuff in a new method, or even in the model (or a helper).

and perhaps nit picking, but the variable names here are hard to follow.  eg, @ordergroups is not a collection of `ordergroups` but rather the `@ordergroups_custom_fields` (yes a mouthful), and `obj` is  not a great name, and a better name would make it clearer what `@ordergroups` contains.

app/views/finance/financial_transactions/new_collection.html.haml
```diff
@@ -16,16 +17,23 @@
...
+        var value;
+        if (!customField) {
+          value = prompt("#{escape_javascript(heading_helper(FinancialTransaction, :amount))}:");
+          if (value === null)
+            return false;
+        }
         $('#ordergroups > tbody > tr').remove();
-        var options = ordergroup.find('td > select > option').each(function() {
+        for (var id in ordergroups) {
```
carchrae: consider using `Array.forEach` (if >es5 is the target) or if backward compatible, the jquery util http://api.jquery.com/jquery.each/

otherwise, at least put `var id` at the top of the function, then `for (id in ordergroups)` - js hoisting and all that.

carchrae: and yes, i realise the rest of the js code in this project isn't super, but still, continuous improvement

app/controllers/finance/financial_transactions_controller.rb
paroga: i'm not a big fan of comments. if they are needed the code is not well written. i'l rework it according to your suggestions.

carchrae: for sure - clear code is the best.  but sometimes, a small hint in a comment is helpful.  i agree that comments aren't the best because the worst part is they can become out of date and misleading!

app/models/ordergroup.rb
wvengen: I agree that `FoodsoftSettings` could be refactored to better support nested settings. This would include, I guess. And then we could convert these hashes to properly nested settings as well. That's some work, so I'm ok not doing that here.

On another note, what about expanding the `CustomFields` to something like (not tested, but you'll get the gist):
```ruby
module CustomFields
  extend ActiveSupport::Concern
  include RailsSettings::Extend

  class_methods do
    def has_custom_fields(key)
      attr_accessor :custom_fields

      after_initialize do
        settings.defaults['custom_fields'] = { } unless settings.custom_fields
      end

      after_save do
        self.settings.custom_fields = custom_fields if custom_fields
      end

      def custom_fields
        @custom_fields ||= begin
          fields = FoodsoftConfig[:custom_fields] && FoodsoftConfig[:custom_fields][key]
          fields ||= []
          fields.map(&:deep_symbolize_keys)
        end
      end
    end
  end
end
```

paroga:
i've updated the commit message with a better description

wvengen:
@carchrae would you like to take a look at this PR again?

PR #604, opened 2018-12-30
Add set_balance to FinancialTransactions#new_collection (fixes #438)

### LINKED ISSUE #438: Set account balance to 0 for all
Example:
- User A has an account balance of -7
- User B has -10

Both users pay afterwards, the new account value is 0.
The administrator now has to enter the exact values they paid, as a new transaction.
For our use case, it would be great to have a button which inserts an appropriate transaction for every user so everyone's account value is at 0 again.
I know this might seem weird, but would be a convenient feature in some cases, e.g. when the administrator exactly knows everyone has paid the right sum.

paroga:
A new checkbox will allow user to set the balance to a given ABSOLUTE value
in addition to changing it by a RELATIVE value. This can be used if the
balance is tracked outside of foodsoft and should be syncroniced or for
setting the balance to zero for multiple ordergroups.

### DIFF
--- a/app/controllers/finance/financial_transactions_controller.rb
+++ b/app/controllers/finance/financial_transactions_controller.rb
@@ -59,12 +59,20 @@ class Finance::FinancialTransactionsController < ApplicationController
     params[:financial_transactions].each do |trans|
       # ignore empty amount fields ...
       unless trans[:amount].blank?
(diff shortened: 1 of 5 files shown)


carchrae:
hah.  i see that @wvengen had similar thoughts in replying to this request in the first place.

PR #606, opened 2018-12-31
Avoid ActionView::MissingTemplate occurred in sessions#new

paroga:
Some browser request /:foodcoop/login with the HTTP-Accept-Header set
to "image/webp,image/*;q=0.8", which leads to an internal server error
due to a not existing template. Call respond_to to allow only html and
respond with the correct "406 Not Acceptable" HTTP status code.


paroga:
AFAIK the API calls use a different route

PR #609, opened 2019-01-14
allow override of default limit (<200) on articles on shared suppliers

carchrae:
on my version, i will sync up to 5000 articles.  this allows the restriction to be lifted by a config option.

(some performance enhancements to make sync faster are coming, but not part of this PR)

### DIFF
--- a/app/models/shared_supplier.rb
+++ b/app/models/shared_supplier.rb
@@ -18,7 +18,7 @@ class SharedSupplier < ActiveRecord::Base
   # return list of synchronisation methods available for this supplier
   def shared_sync_methods
     methods = []
-    methods += %w(all_available all_unavailable) if shared_articles.count < 200
+    methods += %w(all_available all_unavailable)  if shared_articles.count < FoodsoftConfig[:shared_supplier_article_limit]
     methods += %w(import) # perhaps, in the future: if shared_articles.count > 20
     methods
   end
--- a/app/models/shared_supplier.rb
+++ b/app/models/shared_supplier.rb
@@ -18,7 +18,7 @@ class SharedSupplier < ActiveRecord::Base
   # return list of synchronisation methods available for this supplier
   def shared_sync_methods
     methods = []
-    methods += %w(all_available all_unavailable)  if shared_articles.count < FoodsoftConfig[:shared_supplier_article_limit]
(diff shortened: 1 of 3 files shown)


app/models/shared_supplier.rb
```diff
@@ -18,7 +18,7 @@ def autofill_attributes
   # return list of synchronisation methods available for this supplier
   def shared_sync_methods
     methods = []
-    methods += %w(all_available all_unavailable) if shared_articles.count < 200
+    methods += %w(all_available all_unavailable)  if shared_articles.count < (FoodsoftConfig[:shared_supplier_article_limit] || 200)
```
paroga: please set it [`get_default_config`](https://github.com/foodcoops/foodsoft/blob/a66a44bdcf124bc108c35a1c5a191ef475d2a6c5/lib/foodsoft_config.rb#L247) instead

config/app_config.yml.SAMPLE
```diff
@@ -162,6 +162,8 @@ default: &defaults
   # Instead of defining all details here, you can also point to an entry in database.yml.
   #shared_lists: shared_lists
 
+  # default to allow sync only on less than 200 articles
+  #shared_supplier_article_limit: 2000,
```
paroga: please keep the default value in the config file

carchrae: ok, changed to 200

app/models/shared_supplier.rb
carchrae: done

app/models/shared_supplier.rb
```diff
@@ -18,7 +18,7 @@ def autofill_attributes
   # return list of synchronisation methods available for this supplier
   def shared_sync_methods
     methods = []
-    methods += %w(all_available all_unavailable) if shared_articles.count < 200
+    methods += %w(all_available all_unavailable)  if shared_articles.count < FoodsoftConfig[:shared_supplier_article_limit]
```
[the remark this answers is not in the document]
carchrae: sure - that is more specific

paroga:
thx!
BTW: would you mind reviewing my currently pending changes? i'd like to have at least a second pair of eyes looking over it before merging, but @wvengen seems to be no available at the moment.

carchrae:
sure, I'll have a look later this evening

PR #610, opened 2019-01-14
big speedup in syncing via caching/reducing queries

carchrae:
using a little bit of caching cuts down the sync time drastically.  

i'm not sure if this is all of my previous improvements, but it is enough to stop it timing out on large sync operations.

### DIFF
--- a/app/models/shared_supplier.rb
+++ b/app/models/shared_supplier.rb
@@ -1,5 +1,5 @@
 class SharedSupplier < ActiveRecord::Base
-  
+
   # connect to database from sharedLists-Application
   SharedSupplier.establish_connection(FoodsoftConfig[:shared_lists])
   # set correct table_name in external DB
@@ -7,6 +7,37 @@ class SharedSupplier < ActiveRecord::Base
 
   has_many :suppliers
   has_many :shared_articles, :foreign_key => :supplier_id
+
+  def find_article_by_number(order_number)
+    #note inconsistent naming, order_number = number in supplier db
+    cached_articles_by_key(['number'])[cache_key([order_number])]
+  end
+
+  def find_article_by_name_origin_manufacture(name, origin, manufacturer)
+    cached_articles_by_key(['name', 'origin', 'manufacturer'])[cache_key([name, origin, manufacturer])]
+  end
+
+  def find_article_by_name_manufacture(name, manufacturer)
+    cached_articles_by_key(['name', 'manufacturer'])[cache_key([name, manufacturer])]
+  end
+
+  def cache_key(keys)
+    keys.join('::')
+  end
+
+  def cached_articles
+    # puts "cached #{id}" if @cached_articles
+    # puts "shared supplier not cached #{id}" unless @cached_articles
+    @cached_articles ||= shared_articles.all
+  end
+
+  def cached_articles_by_key(keys)
+    @cached_articles_by_key ||= {}
+    @cached_articles_by_key[keys.join('::')] ||= cached_articles.map do |a|
+      vals = keys.map {|k| a[k]}
+      [cache_key(vals), a]
+    end.to_h
+  end
 
   # These set of attributes are used to autofill attributes of new supplier,
   # when created by import from shared supplier feature.
--- a/app/models/shared_supplier.rb
+++ b/app/models/shared_supplier.rb
@@ -8,23 +8,14 @@ class SharedSupplier < ActiveRecord::Base
   has_many :suppliers
   has_many :shared_articles, :foreign_key => :supplier_id
 
+
   def find_article_by_number(order_number)
-    #note inconsistent naming, order_number = number in supplier db
-    cached_articles_by_key(['number'])[cache_key([order_number])]
+    # note that `shared_articles` uses number instead order_number
+    cached_articles.detect { |a| a.number == order_number }
   end
 
-  # generates a consistent lookup key in the cache for a list of key values
-  def cache_key(keys)
-    keys.join('::')
-  end
-
-  # indexes all in an in-memory hash for fast lookup (at the cost of the first load)
-  def cached_articles_by_key(keys)
-    @cached_articles_by_key ||= {}
-    @cached_articles_by_key[keys.join('::')] ||= shared_articles.all.map do |a|
-      vals = keys.map { |k| a[k] }
-      [cache_key(vals), a]
-    end.to_h
+  def cached_articles
+    @cached_articles ||= shared_articles.all
   end
 
   # These set of attributes are used to autofill attributes of new supplier,
--- a/app/models/shared_supplier.rb
+++ b/app/models/shared_supplier.rb
@@ -28,7 +28,7 @@ class SharedSupplier < ActiveRecord::Base
   # return list of synchronisation methods available for this supplier
   def shared_sync_methods
     methods = []
-    methods += %w(all_available all_unavailable)  if shared_articles.count < (FoodsoftConfig[:shared_supplier_article_limit] || 200)
+    methods += %w(all_available all_unavailable) if shared_articles.count < 200
     methods += %w(import) # perhaps, in the future: if shared_articles.count > 20
     methods
   end
--- a/app/models/shared_supplier.rb
+++ b/app/models/shared_supplier.rb
@@ -13,28 +13,16 @@ class SharedSupplier < ActiveRecord::Base
     cached_articles_by_key(['number'])[cache_key([order_number])]
   end
 
-  def find_article_by_name_origin_manufacture(name, origin, manufacturer)
-    cached_articles_by_key(['name', 'origin', 'manufacturer'])[cache_key([name, origin, manufacturer])]
-  end
-
-  def find_article_by_name_manufacture(name, manufacturer)
-    cached_articles_by_key(['name', 'manufacturer'])[cache_key([name, manufacturer])]
-  end
-
+  # generates a consistent lookup key in the cache for a list of key values
   def cache_key(keys)
     keys.join('::')
   end
 
-  def cached_articles
-    # puts "cached #{id}" if @cached_articles
-    # puts "shared supplier not cached #{id}" unless @cached_articles
-    @cached_articles ||= shared_articles.all
-  end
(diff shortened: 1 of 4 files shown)


paroga:
I'm not a big Rails expert, but I'd expect hat this is handled internally by the framework. Do you know why this is not handled by *SQL Caching* as described in [Caching with Rails: An overview](https://guides.rubyonrails.org/v4.2/caching_with_rails.html) (see section 1.5)?

carchrae:
it is true that some things like the categories collection fix probably don't add a lot and could be fixed by caching, but the loops in [sync_all](https://github.com/foodcoops/foodsoft/pull/610/files#diff-8e19dea549c239408cd96e75921f2badL33) are the killers.  

the real problem isn't so much the SQL caching, but the N x M queries being run to check if things are in sync or not.  rails makes it so easy to just iterate through a collection, you nearly forget that it is hooked up to a database.  i'm sure this external sync was fine on small tables (and indeed, why the 200 limit is probably included)

my changes load the entire table of articles in one go.  we're going to iterate through all of them anyway, so just loading them into memory in one query cuts down a lot of time.  there are some other subtle changes too that save some checks.

the argument against this fix might be that it'll use too much memory.  however, i've been running it on heroku and that has been fine.

there are other areas that are really slow in the application too and could use more optimisation.  the received page frequently takes over 30 seconds (which times out heroku)  i'm going to address that one sometime soon (my current fix was to move to a VPS server instead)

paroga:
sorry, i missed the `shared_articles.all` in `cached_articles`.

I personally do not see a big problem with loading everything into memory, but it would be great if you could use the [Rails functions](https://stackoverflow.com/questions/30927153/rails-activerecord-find-from-activerecord-result-in-o1-time) instead of reinventing the wheel ;-) (see

paroga:
i agree that e.g. `cached_articles.detect{ |a| a.name == name }` will loop over the __cached__ results instead of doing a simple lookup, but it won't hit the database. linear search is not the best choice, but should be enough in this case and makes the code more readable. if you insist on a lookup table for the items, i'd like to see the caching code moved into its own class with corresponding unit tests

carchrae:
you make a fair point about the lack of tests.  in fact, i've committed too many methods, so which i haven't used in this branch yet.  

but i don't think i follow, if this is your point about `detect`

```
cached_articles.detect{ |a| a.name == name }
```
is more readable than 
```
cached_articles[name]
```
(never mind the O(N^2) vs O(N) runtime)

paroga:
`cached_articles[name]`if of course more readable, but i see the following code in your change, which I qualify as perfectly readable ;-)
```ruby
def cached_articles_by_key(keys)
  @cached_articles_by_key ||= {}
  @cached_articles_by_key[keys.join('::')] ||= cached_articles.map do |a|
    vals = keys.map {|k| a[k]}
    [cache_key(vals), a]
  end.to_h
end
```

BTW: and if you are so super concerned about performance, doing the string concatenation `keys.join('::')`  and hash lookup `@cached_articles_by_key[]` for every `find_article_by_*` isn't the best solution either.

maybe i should write my whole code, to make it more clear:
```ruby
def find_article_by_number(order_number)
  cached_articles.detect{ |a| a.number == order_number }
end

def cached_articles
  @cached_articles ||= shared_articles.all
end
```

Sorry for being so picky, but I see two ways to go: Use `detect` (if only the database is the bottleneck) and O(N) is good enough or implement a new class which handles caching with an O(log N) lookup strategy and avoids __all__ unnecessary lookups. I didn't do the performance testing, so I don't know if the fist approach is enough.

carchrae:
i don't mind picky - appreciate the care and attention to the code.  


here are some performance numbers.  i'll update in a sec with `detect` too.  

```
sync of 513 articles.  17 of which have updates

initial PR performance
I, [2019-01-14T16:05:50.466639 #3346]  INFO -- : Completed 200 OK in 5292ms (Views: 240.5ms | ActiveRecord: 3911.0ms)
I, [2019-01-14T16:06:17.853097 #3346]  INFO -- : Completed 200 OK in 3790ms (Views: 108.4ms | ActiveRecord: 3267.9ms)
I, [2019-01-14T16:06:48.769787 #3346]  INFO -- : Completed 200 OK in 3649ms (Views: 129.8ms | ActiveRecord: 3110.5ms)

removing ArticleCategory.all change

I, [2019-01-14T16:10:13.308549 #3892]  INFO -- : Completed 200 OK in 5284ms (Views: 251.1ms | ActiveRecord: 3763.5ms)
I, [2019-01-14T16:10:35.327668 #3892]  INFO -- : Completed 200 OK in 3578ms (Views: 102.0ms | ActiveRecord: 3080.9ms)



previous (no cache)

I, [2019-01-14T16:15:37.281285 #4277]  INFO -- : Completed 200 OK in 141620ms (Views: 6648.8ms | ActiveRecord: 127521.5ms)
I, [2019-01-14T16:19:09.864097 #4277]  INFO -- : Completed 200 OK in 140588ms (Views: 5369.7ms | ActiveRecord: 128796.3ms)


simplified (last commits)

I, [2019-01-14T16:37:43.643551 #6549]  INFO -- : Completed 200 OK in 4467ms (Views: 144.8ms | ActiveRecord: 3793.5ms)
I, [2019-01-14T16:38:19.633525 #6549]  INFO -- : Completed 200 OK in 4121ms (Views: 115.5ms | ActiveRecord: 3584.7ms)
```

carchrae:
and it seems detect is a smidge slower, but it is hairs compared to the 120second original behaviour.

```
detect

I, [2019-01-14T16:43:39.560784 #7245]  INFO -- : Completed 200 OK in 6680ms (Views: 289.2ms | ActiveRecord: 4858.6ms)
I, [2019-01-14T16:44:07.734539 #7245]  INFO -- : Completed 200 OK in 4938ms (Views: 152.6ms | ActiveRecord: 3922.3ms)
I, [2019-01-14T16:44:32.980114 #7245]  INFO -- : Completed 200 OK in 4848ms (Views: 165.9ms | ActiveRecord: 3981.2ms)
I, [2019-01-14T16:45:20.206330 #7245]  INFO -- : Completed 200 OK in 4833ms (Views: 192.1ms | ActiveRecord: 3791.0ms)
```

paroga:
as expected it is a lot faster, but not as fast as your initial PR. 😄 
is this is already fast enough for you or do you need it faster?

carchrae:
some more numbers.  this is the impact of reverting the  `existing_articles` optimisation: https://github.com/foodcoops/foodsoft/pull/610/files#diff-8e19dea549c239408cd96e75921f2badR35

```
I, [2019-01-14T16:52:21.412971 #7882]  INFO -- : Completed 200 OK in 99781ms (Views: 6510.7ms | ActiveRecord: 83442.7ms)
I, [2019-01-14T16:54:29.644251 #7882]  INFO -- : Completed 200 OK in 96100ms (Views: 5681.1ms | ActiveRecord: 81924.1ms)
```

so, clearly, that bit needs to stay too.

> is this is already fast enough for you or do you need it faster

i can live with 5 seconds.  :)    

thanks for challenging me to back up the claims.

app/models/shared_supplier.rb
```diff
@@ -18,7 +28,7 @@ def autofill_attributes
   # return list of synchronisation methods available for this supplier
   def shared_sync_methods
     methods = []
-    methods += %w(all_available all_unavailable) if shared_articles.count < 200
+    methods += %w(all_available all_unavailable)  if shared_articles.count < (FoodsoftConfig[:shared_supplier_article_limit] || 200)
```
carchrae: ah nuts. will clean up

app/models/supplier.rb
```diff
@@ -32,11 +32,13 @@ class Supplier < ActiveRecord::Base
   # also returns an array with new articles, which should be added (depending on shared_sync_method)
   def sync_all
     updated_article_pairs, outlisted_articles, new_articles = [], [], []
+    existing_articles = {}
```
paroga: Using a `Set` instead of a `Hash` would make it clearer

app/models/supplier.rb
```diff
@@ -51,8 +53,9 @@ def sync_all
     end
     # Find any new articles, unless the import is manual
     unless shared_sync_method == 'import'
-      for shared_article in shared_supplier.shared_articles
-        unless articles.undeleted.find_by_order_number(shared_article.number)
+      # for shared_article in shared_supplier.shared_articles
+      for shared_article in shared_supplier.shared_articles.all
```
paroga: could we filter the existing articles already on the database level with something like `shared_supplier.shared_articles.where(available: true).where.not(id: existing_articles)`?

paroga: i care more about the readability than the performance.
thanks for your patience and i'm looking forward to your next PR!

app/models/supplier.rb
```diff
@@ -51,11 +53,11 @@ def sync_all
...
     # Find any new articles, unless the import is manual
     unless shared_sync_method == 'import'
-      for shared_article in shared_supplier.shared_articles
-        unless articles.undeleted.find_by_order_number(shared_article.number)
-          new_articles << shared_article.build_new_article(self)
-        end
-      end
+      shared_supplier
+        .shared_articles
+        .where(available: true)
```
wvengen: It looks like the `available` attribute is something you have in your sharedlists, but which isn't present in the [stock version](https://github.com/foodcoops/sharedlists/blob/master/db/schema.rb#L16), right? 
This [came up on the mailing-list](http://foodsoft.51229.x6.nabble.com/Configuring-sharedlists-for-Terra-tp2076p2096.html). I'll remove the `.where(available: true)`.

carchrae: i suppose i must have @wvengen - it has been many years, but it doesn't appear to exist in the original queries.  my apologies for the support hassle.

fwiw, i built my own shared-list app in node.js - it is a horrendous regex monster, but it usually gets 99% of the information out of a produce catalog string.

wvengen: No probs, we missed this during review. I just corrected it on master.
Interesting, feel free to add a link to the wiki (on a new "Related software" page or so).

PR #611, opened 2019-01-15
localize without errors applying to non-string values

carchrae:
i fixed this a few years ago.  just submitting it here.  https://github.com/bennibu/localize_input/pull/2

IMHO, the localise gem is kind of questionable.  if using something like this i'd opt for either something more supported/widely used, or just pull the code into foodsoft.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ gem 'haml-rails'
 gem 'kaminari'
 gem 'simple_form'
 gem 'inherited_resources'
-gem 'localize_input', git: "https://github.com/bennibu/localize_input.git"
+gem 'localize_input', git: 'https://github.com/carchrae/localize_input.git'
 gem 'daemons'
 gem 'doorkeeper'
 gem 'doorkeeper-i18n'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
-  remote: https://github.com/bennibu/localize_input.git
-  revision: 5eb188d2525a073d09e142cf8b0b04e6ace6e7b0
+  remote: https://github.com/carchrae/localize_input.git
+  revision: 597e33bb42a3b0fd58c3ea7d4f5a208c2911b43d
   specs:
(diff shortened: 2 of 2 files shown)


carchrae:
i should mention what i was actually fixing....   on my instance `localize_input` was being applied on many `BigNumber` values.  so, it was throwing a lot of errors.

paroga:
YEAH! finally somebody addressing this!
what do you think about removing it at all and replace it with something else? IMHO [`delocalize`](https://github.com/clemens/delocalize) seams to be the moste used one for this?

carchrae:
i'm all for removing it.  :fire:  :fire: :fire: 

delocalize seems like it would fit.  locale, timezones, all those things - things i don't want to code/maintain!  :)

carchrae:
i had a look at https://github.com/clemens/delocalize#controller-setup - and i think it definitely the way to go.  in fact, the whole idea that you sanitise inputs in the controller and _not_ in the model is what had deeply bothered me about the current use of `localize_input_of` - which was being called every time a model value was set, regardless of if it was an external/user input or not (ie, it currently wastes a tremendous amount of effort doing pointless transforms).  instead, doing it on controller parameters is a much saner way to handle this.

and the fun doesn't stop at localisation.  https://github.com/mdeering/attribute_normalizer works in the same fashion - on the model, causing much cpu cycles to be blown ensuring that already normalised values are normalised.  

so - to actually do this _right_ i think all this sanitising/normalising code should all move to controllers.  but that is no small undertaking, so i'm going to suggest that this PR is merged as-is because it is a small/quick change but a huge improvement so at least the current code will not be applied to values that have already been transformed from a string into a number. https://github.com/carchrae/localize_input/blob/master/lib/localize_input.rb#L14

carchrae:
bump.  bump. 

while i think doing a review of how values are localised is worthwhile, it is a big effort.  i suggest this PR be merged to at least stop all the errors noise in the log.

wvengen:
Ok, it's an improvement, and we can improve later. Thanks!

PR #614, opened 2019-01-18
don't allow changes to settled orders

carchrae:
this adds a few checks on the order accounting page so that you cannot adjust what people received after the order is closed (aka settled).  it would be bad news to adjust what people received after it has been settled.

i also added a model level check to prevent any saving of `GroupOrderArticle` on a settled order - so this will enforce the rule at the model level too.

i have a separate PR for re-opening a settled order (which also reverse the charges).

### DIFF
--- a/app/models/group_order_article.rb
+++ b/app/models/group_order_article.rb
@@ -5,13 +5,13 @@ class GroupOrderArticle < ApplicationRecord
 
   belongs_to :group_order
   belongs_to :order_article
-  has_many :group_order_article_quantities, :dependent => :destroy
+  has_many   :group_order_article_quantities, :dependent => :destroy
 
   validates_presence_of :group_order, :order_article
-  validates_uniqueness_of :order_article_id, :scope => :group_order_id # just once an article per group order
-  validate :check_order_not_closed  #don't allow changes to closed (aka settled) orders
+  validates_uniqueness_of :order_article_id, :scope => :group_order_id    # just once an article per group order
+  validate :check_order_not_closed # don't allow changes to closed (aka settled) orders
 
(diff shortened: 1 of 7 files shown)


carchrae:
sorry about the whitespace changes: https://github.com/foodcoops/foodsoft/pull/614/files?w=1

app/models/group_order_article.rb
```diff
@@ -195,4 +196,12 @@ def total_price(order_article = self.order_article)
   def result_manually_changed?
     result != result_computed unless result.nil?
   end
+
+  private
+
+  def check_order_not_closed
+    if order_article.order.closed?
+      errors.add(:order_article, 'order is closed/settled and cannot be modified')
```
paroga: i18n?

app/views/finance/balancing/_group_order_articles.html.haml
```diff
@@ -10,7 +10,7 @@
         %td= t('.total')
         %td{:colspan => "3",:style => "width:14em"}
           = link_to t('.add_group'), new_group_order_article_path(order_article_id: order_article.id),
-            remote: true, class: 'btn btn-mini'
+            remote: true, class: 'btn btn-mini' unless order_article.order.closed?
```
wvengen: Since this is all the way at the end of a long line, I'd put the conditional on a separate line:
```haml
%td{:colspan => "3",:style => "width:14em"}
  - unless order_article.order.closed?
    = link_to t('.add_group'), new_group_order_article_path(order_article_id: order_article.id), remote: true, class: 'btn btn-mini'
```

carchrae: i got busy and missed this.  i agree.  this is addressed by latest commit (hopefully it is ok, because i used the github online editor)

PR #615, opened 2019-01-18
fix missing translation

carchrae:
a missing translation on a modal.

a minor relabel got committed here too by accident - i would revert, but i think my change is clearer than the original "Notes/Protocol"

lots of small fiddly stuff in my branch like this.  i will persist in committing it.

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -905,7 +905,7 @@ de:
         with_extra_charge: 'mit Aufschlag:'
         without_extra_charge: 'ohne Aufschlag:'
       edit_note:
-        title: Bestellschein bearbeiten
+        title: Notitie bewerken
     create:
       notice: Rechnung wurde erstellt.
     financial_links:
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -904,6 +904,8 @@ de:
         reload: Zusammenfassung neu laden
         with_extra_charge: 'mit Aufschlag:'
         without_extra_charge: 'ohne Aufschlag:'
+      edit_note:
+        title: Bestellschein bearbeiten
     create:
       notice: Rechnung wurde erstellt.
     financial_links:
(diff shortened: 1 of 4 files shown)


config/locales/de.yml
```diff
@@ -904,6 +904,8 @@ de:
         reload: Zusammenfassung neu laden
         with_extra_charge: 'mit Aufschlag:'
         without_extra_charge: 'ohne Aufschlag:'
+      edit_note:
+        title: Bestellschein bearbeiten
```
paroga: Why did you change it from "Notiz" to "Bestellschein"? IMHO it still refers to "Note"...

carchrae: you have to remember, i don't know german from dutch!  i didn't know what the original was.  :)  

will fix this eve

paroga: AFAIK all untranslated texts are from benni and therefore in German langugage.

config/locales/de.yml
```diff
@@ -904,6 +904,8 @@ de:
         reload: Zusammenfassung neu laden
         with_extra_charge: 'mit Aufschlag:'
         without_extra_charge: 'ohne Aufschlag:'
+      edit_note:
+        title: Notitie bewerken
```
wvengen: This looks like the Dutch translation :) Should be `Notiz bearbeiten`,

carchrae: ha.  ok.  done

wvengen:
I guess we can merge this, unless I'm missing something. @paroga (just to be sure, as you reviewed before).

carchrae:
ha.  well, i can't merge it as i don't have access.  this was perhaps the most work on a PR for the smallest change.  

i would suggest if there are language changes that reviewers just directly commit their language on a PR.  there is also a way to suggest changes in github (it is beta) which would work well in cases like this: https://help.github.com/articles/incorporating-feedback-in-your-pull-request/#applying-a-suggested-change

PR #621, opened 2019-02-03
Use the correct locale when sending mails to suppliers

paroga:


wvengen:
But perhaps ... not sure if this should use the user's locale or the foodcoop's default locale?

paroga:
i was just too lacy to add the additional code in the first place 🙄

PR #622, opened 2019-02-04
Fix rendering of stock order in OrderByGroups

paroga:


lib/order_pdf.rb
```diff
@@ -132,7 +132,7 @@ def stock_ordergroup_name
     users = GroupOrder.
       eager_load(:updated_by).
       where(order: @orders).
-      where(ordergroup: nil).
+      where(ordergroup: 0).
```
wvengen: Hmmm I'd make this a scope on the `GroupOrder` model`.

PR #624, opened 2019-02-07
API-3.1: Add OAuth scopes

wvengen:
Part one of #572 (which was part of #429), continues from #577 (only the last commit here is relevant).

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -1,4 +1,6 @@
 class Api::V1::BaseController < ApplicationController
+  include Concerns::AuthApi
+
   protect_from_forgery with: :null_session
 
   before_action :skip_session
@@ -11,16 +13,6 @@ class Api::V1::BaseController < ApplicationController
 
   private
 
-  def authenticate
-    doorkeeper_authorize!
-    super if current_user
-  end
-
-  # @return [User] Current user, or +nil+ if no valid token.
-  def current_user
-    @current_user ||= User.undeleted.find(doorkeeper_token.resource_owner_id) if doorkeeper_token
-  end
-
   # @return [Ordergroup] Current user's ordergroup, or +nil+ if no valid token or user has no ordergroup.
   def current_ordergroup
     current_user.try(:ordergroup)
@@ -28,7 +20,9 @@ class Api::V1::BaseController < ApplicationController
 
   def require_ordergroup
     authenticate
-    raise Api::Errors::PermissionRequired unless current_user.ordergroup.present?
+    unless current_ordergroup.present?
+      raise Api::Errors::PermissionRequired.new('Forbidden, must be in an ordergroup')
+    end
   end
 
   def skip_session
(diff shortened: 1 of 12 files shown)


app/controllers/api/v1/base_controller.rb
```diff
@@ -11,24 +13,16 @@ class Api::V1::BaseController < ApplicationController
...
   # @return [Ordergroup] Current user's ordergroup, or +nil+ if no valid token or user has no ordergroup.
   def current_ordergroup
     current_user.try(:ordergroup)
   end
 
   def require_ordergroup
     authenticate
-    raise Api::Errors::PermissionRequired unless current_user.ordergroup.present?
+    unless current_ordergroup.present?
+      raise Api::Errors::PermissionRequired.new("Forbidden, must be in an ordergroup")
```
paroga: single quotes?

app/controllers/api/v1/base_controller.rb
```diff
@@ -42,13 +36,18 @@ def not_found_handler(e)
   end
 
   def not_acceptable_handler(e)
-    render status: 422, json: {error: 'not_acceptable', error_description: e.message || 'Data not acceptable' }
+    msg = e.message || 'Data not acceptable'
+    render status: 422, json: {error: 'not_acceptable', error_description: msg }
```
paroga: space before the closing bracket?

config/initializers/doorkeeper.rb
```diff
@@ -47,8 +47,13 @@
...
   # https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Scopes
-  # default_scopes  :public
-  # optional_scopes :write, :update
+
+  # default is a collection of read-only scopes
+  default_scopes 'config:user', 'user:read'
+
+  optional_scopes 'config:read', 'config:write',
+                  'user:write',
+                  'offline_access'
```
paroga: do we use `offline_access` already somewhere?

app/controllers/concerns/auth_api.rb
```diff
@@ -0,0 +1,66 @@
...
+    doorkeeper_authorize_roles!(*scopes) if scopes.present? && !performed?
+  end
+
+  private
+
+  # Make sure that at least one the given OAuth scopes is valid for the current user's permissions.
+  # @raise Api::Errors::PermissionsRequired
+  def doorkeeper_authorize_roles!(*scopes)
+    unless scopes.any? {|scope| doorkeeper_scope_permitted?(scope) }
+      raise Api::Errors::PermissionRequired.new("Forbidden, no permission")
```
paroga: singel quotes?

wvengen:
Thanks for the review. The scopes don't feel too nice for me, but I don't see how to really improve that. I agree about being able to change them, until we do a release. I'll add a note to the API docs about that.
I guess this is also a green light for #577 (which this PR builds on).

config/initializers/doorkeeper.rb
wvengen: It is used [here](https://github.com/foodcoops/foodsoft/blob/master/app/controllers/concerns/auth.rb#L138) (already on master).

PR #626, opened 2019-02-11
Allow changing the visibitly (public/privat) of messages (fixes #625)

### LINKED ISSUE #625: add option to make public messages private
Somtimes users send mails via foodsoft erroneously not tagged as private. That can be quite embarrassing, in any case they are confusing for others. There is no way to remove them. At least admins should have the ability to delete those from the message area.

paroga:


plugins/messages/app/views/messages/show.haml
```diff
@@ -23,6 +23,15 @@
...
       %td= t '.sent_on'
       %td= format_time(@message.created_at)
+    %tr
+      %td= t '.visibility'
+      %td
+        - if @message.private
+          = t '.visibility_private'
+        - else
+          = t '.visibility_public'
+        - if 1 ||@message.can_toggle_private?(current_user)
```
wvengen: oops :)

plugins/messages/app/models/message.rb
```diff
@@ -124,6 +124,12 @@ def is_readable_for?(user)
     !private || sender == user || recipients_ids.include?(user.id)
   end
 
+  def can_toggle_private?(user)
+    return true if sender == user
+    return false if private
```
wvengen: Tiny note: I'd use `private?`, but it doesn't really matter.

PR #628, opened 2019-02-23
Fix docker on development

### LINKED ISSUE #585: proposal: remove `Dockerfile-dev.yml` as it is broken (and not used by anyone?)
i spent a lot of time to try and use the development docker.  i eventually got it working, but i have some concerns about the approach in general.

current issues:

* there are some simple fixes, like this: #551 (but actually, it seems a waste to add mailcatcher as a docker, it makes a bigger image rather than running mailcatcher in the rails app container, which adds little bloat)

* resque and redis 4 are broken.  https://github.com/resque/resque/issues/1614 `Error: undefined method 'connect' for Redis:Class (NoMethodError)`  - one fix is to lock redis in the Gemfile with `gem 'redis', '~> 3.3'   # bug with resque https://github.com/resque/resque/issues/1614` - the 'better' fix is upgrade to use `resque 2` - but i didn't explore that. 

* non-standard `Dockerfile-dev` and `docker-compose-dev.yml` are a pain to work with and clutter the root of the project.  would be nicer to have these in a subdir.  

* but at a deep level, why even have a development version of the Dockerfile - it seems wrong to me.  instead, maintain one Dockerfile that is used in both production and development.  this would be much cleaner.  if you need to alter for development, then consider adding some environment variables to do that.

i do have a working environment and could make a PR that addresses all but the last comment, but i am not sure if it is the approach i would choose.

however, i do have an immediate suggestion - we should delete the development one because it is broken.  if a new developer comes to the project, the expectation with a dockerfile is that you can get up and running instantly, but in this case, it took me a lot of time to get it working.  this can scare off new contributors.  as in all OSS projects, cleaning up the detritus makes it easier for new people to join.

wvengen:
Fixes #585.

We might still consider moving the files to `.docker`, but this should fix the setup.

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -22,9 +22,9 @@ services:
       - TEST_DATABASE_URL=mysql2://root:secret@mariadb/test?encoding=utf8
 
   mailcatcher:
-    image: aboutsource/mailcatcher
+    image: tophfr/mailcatcher
     ports:
-      - "1080:1080"
+      - "1080:80"
 
   mariadb:
     image: mariadb:10.1
--- a/bin/test
+++ b/bin/test
@@ -11,7 +11,7 @@ Xvfb $DISPLAY -nolisten tcp &
 XVFB_PID=$!
 
 # Start tests
-rspec $@
+bundle exec rspec $@
 
 # Cleanup
 kill $XVFB_PID
--- a/config/environments/development.rb.SAMPLE
+++ b/config/environments/development.rb.SAMPLE
@@ -43,7 +43,7 @@ Foodsoft::Application.configure do
   # Mailcatcher config, start mailcatcher from console with 'mailcatcher'
   # Mailcatcher can be installed by gem install mailcatcher
   config.action_mailer.delivery_method = :smtp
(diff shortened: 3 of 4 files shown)


wvengen:
Planning to merge this after a week.

docker-compose-dev.yml
```diff
@@ -22,9 +22,9 @@ services:
       - TEST_DATABASE_URL=mysql2://root:secret@mariadb/test?encoding=utf8
 
   mailcatcher:
-    image: aboutsource/mailcatcher
+    image: tophfr/mailcatcher
     ports:
-      - "1080:1080"
+      - "1080:80"
```
carchrae: depending on what OS you are working with, mapping to port 80 may be restricted.  maybe this is not a problem with docker (does it run with root level permissions).

i would also generally avoid 80 since it is the default http port - and that means someone must not have anything else using it on their machine. 

this is a minor thing.

wvengen: True. But this is about the container-side [port](https://docs.docker.com/compose/compose-file/compose-file-v2/#ports), which moved from 1080 to 80 here. So it will bind to 1080 on the maching running Docker.

wvengen: Thanks for reviewing, btw :)

carchrae: ah, my bad then!  :)  

and yes, thank you for fixing this too.  i haven't tried it out yet, but i know that extra eyes are always useful.

PR #629, opened 2019-03-04
sort by name as default

carchrae:
there are many screens that show either articles or member's names in somewhat random orders.  this PR sorts several such cases by name, which i would argue, is far easier for most humans to work with.  

perhaps there will be some complaint about removing the default sort by order number on one page, however, i feel even this is not a good default as i have not seen any invoices from suppliers that are sorted like that.  i would argue that if anything but name is required, then those pages should have the ability to sort by any field shown. (it would be nice if all screens showing table data had this feature)

for now though, i hope you see the sense in this PR to improve usability.

### DIFF
--- a/app/models/group_order_article.rb
+++ b/app/models/group_order_article.rb
@@ -11,6 +11,8 @@ class GroupOrderArticle < ApplicationRecord
   validates_uniqueness_of :order_article_id, :scope => :group_order_id    # just once an article per group order
 
   scope :ordered, -> { includes(:group_order => :ordergroup).order('groups.name') }
+  default_scope { joins(:order_article => :article).order('articles.name') }
+
 
   localize_input_of :result
 
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -122,7 +122,7 @@ class OrdersController < ApplicationController
   def receive
     @order = Order.find(params[:id])
     unless request.post?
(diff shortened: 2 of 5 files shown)


carchrae:
hmm - looks like tests presumed a particular order.  i can look into it/fix, but please let me know if you think this PR would be an acceptable change.

app/models/group_order_article.rb
```diff
@@ -11,6 +11,8 @@ class GroupOrderArticle < ApplicationRecord
   validates_uniqueness_of :order_article_id, :scope => :group_order_id    # just once an article per group order
 
   scope :ordered, -> { includes(:group_order => :ordergroup).order('groups.name') }
+  default_scope { joins(:order_article => :article).order('articles.name') }
+
```
wvengen: I'm usually quite careful using default scopes, and joining wouldn't be in my book. Ran into cases where I had to used `unscoped` to make queries work.
Instead, why not adding a scope called e.g. `sorted` or `sorted_by_article_name` and use that in controllers?

Same applies to other default scopes.

carchrae: i think you're right - it had not caused me a problem until i used the pickups view - where the sorting conflicted with sorting by supplier and then by name.

it will take more changes now in the code though i think, to find all the usages, but i agree that this is the better way to do it.

rails is sometimes a little too convenient.  best to avoid those features that are quick fixes now but create a mess later.

yksflip:
Hey,

PR #630, opened 2019-03-04
add supplier's price

carchrae:
this PR adds an attribute to store the supplier price.  there are a few reasons why this is useful

1) balancing your accounts.  currently foodsoft infers the supplier price by price*unit_quantity, however due to rounding errors, this can often be off (by at most unit_quantity).  for example, if we ordered 300 apples (which we do) and there is a rounding error, then the accumulated error can be as high as $3.  it may seem picky, but having the account balance match the invoice exactly is a good check.

2) easier to (re)compute the the article price.  instead of asking the co-op members to pull out a calculator, we add a button on the edit item page which will do the math for you.  so, if you decide you want to change the units from 50 x 1LB to 10 x 5LB, it will warn you and offer to recompute the new price for 5LB units.

this PR is missing the schema update.  i will look into that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -115,7 +115,3 @@ group :test do
   gem 'apivore', require: false
   gem 'hashie', '~> 3.4.6', require: false # https://github.com/westfieldlabs/apivore/issues/114
 end
-
-gem "dotenv-rails", "~> 2.6"
-
-gem "pg", "~> 0.19.0"
--- a/Gemfile
+++ b/Gemfile
@@ -115,3 +115,7 @@ group :test do
   gem 'apivore', require: false
   gem 'hashie', '~> 3.4.6', require: false # https://github.com/westfieldlabs/apivore/issues/114
 end
+
+gem "dotenv-rails", "~> 2.6"
+
+gem "pg", "~> 0.19.0"
--- a/app/models/article_price.rb
+++ b/app/models/article_price.rb
@@ -32,3 +32,4 @@ class ArticlePrice < ApplicationRecord
   validates_numericality_of :deposit, :tax
 
 end
+
--- a/app/models/article_price.rb
+++ b/app/models/article_price.rb
@@ -27,7 +27,7 @@ class ArticlePrice < ApplicationRecord
   localize_input_of :price, :tax, :deposit
 
   validates_presence_of :price, :tax, :deposit, :unit_quantity
-  validates_numericality_of :price, :greater_than_or_equal_to => 0
+  validates_numericality_of :price, :supplier_price, :greater_than_or_equal_to => 0
   validates_numericality_of :unit_quantity, :greater_than => 0
   validates_numericality_of :deposit, :tax
 
--- a/app/models/article_price.rb
+++ b/app/models/article_price.rb
@@ -20,6 +20,9 @@ class ArticlePrice < ApplicationRecord
   # @!attribute order_articles
   #   @return [Array<OrderArticle>] Order articles this price is associated with.
   has_many :order_articles
+  # @!attribute supplier_price
+  #   @return [Number] Supplier's case price
+  #   @see Article#supplier_price
 
   localize_input_of :price, :tax, :deposit
 
--- a/app/models/concerns/price_calculation.rb
+++ b/app/models/concerns/price_calculation.rb
@@ -12,14 +12,14 @@ module PriceCalculation
     add_percent(gross_price, FoodsoftConfig[:price_markup])
   end
 
+  def supplier_price
+    read_attribute(:supplier_price) || (unit_quantity * price unless unit_quantity.nil? || price.nil?)
+  end
+
   private
 
   def add_percent(value, percent)
     (value * (percent * 0.01 + 1)).round(2)
-  end
-
-  def supplier_price
-    read_attribute(:supplier_price) || unit_quantity * price
   end
 
 end
--- a/app/models/concerns/price_calculation.rb
+++ b/app/models/concerns/price_calculation.rb
@@ -17,4 +17,9 @@ module PriceCalculation
   def add_percent(value, percent)
     (value * (percent * 0.01 + 1)).round(2)
   end
+
+  def supplier_price
+    read_attribute(:supplier_price) || unit_quantity * price
+  end
+
 end
--- a/app/models/concerns/price_calculation.rb
+++ b/app/models/concerns/price_calculation.rb
@@ -21,5 +21,4 @@ module PriceCalculation
   def add_percent(value, percent)
     (value * (percent * 0.01 + 1)).round(2)
   end
-
 end
--- a/app/views/shared/_article_fields_price.html.haml
+++ b/app/views/shared/_article_fields_price.html.haml
@@ -17,12 +17,18 @@
       = Article.human_attribute_name(:fc_price)
     .controls.control-text#article_fc_price
       = number_to_currency(f.object.fc_price) rescue nil
+.fold-line
+  = f.input :supplier_price do
+    .input-prepend
+      %span.add-on= t 'number.currency.format.unit'
+      = f.input_field :supplier_price, class: 'input-mini'
+
 
 -# do this inline, since it's being used in ajax forms only
 - field = f.object.class.model_name.to_s.underscore
 :javascript
   var form = $('#article_fc_price').closest('form');
-  $('##{field}_price, ##{field}_tax, ##{field}_deposit', form).on('change keyup', function() {
+  function updatePrice(){
     var price = parseFloat($('##{field}_price', form).val());
     var tax = parseFloat($('##{field}_tax', form).val());
     var deposit = parseFloat($('##{field}_deposit', form).val());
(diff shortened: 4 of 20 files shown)


carchrae:
i have numerous improvements to the account balancing page, but i will submit those in another PR.  many of those make use of supplier_price.

carchrae:
@wvengen  - i had wanted to run these tests locally so i could investigate and fix the code (or tests) but the rake task used in travis does not work.  what command do you recommend for running the tests locally?

carchrae:
(i am pretty keen to get the supplier_price modification in because many of my changes use it and it is really tedious/error-prone to keep two forks of the code and copy pasting only parts of the code)

app/models/concerns/price_calculation.rb
```diff
@@ -17,4 +17,13 @@ def fc_price
...
   end
+
+  def supplier_price
+    unless self[:supplier_price]
+      unit_quantity*price
+    else
+      self[:supplier_price]
+    end
+  end
+
```
wvengen: Wouldn't this be better expressed as
```ruby
read_attribute(:supplier_price) || unit_quantity * price
```

app/views/shared/_article_fields_price.html.haml
```diff
@@ -1,3 +1,7 @@
+.fold-line
+  .split-price-warning
+    .alert.alert-warning Supplier price does not equal (Price * Unit).  If this is not intended use the 'Recalculate Split Price' button to update the price.
```
wvengen: Double space in this line.
I guess you mean _price * unit_quantity_?
Let's also use i18n here.

app/views/shared/_article_fields_price.html.haml
```diff
@@ -17,18 +21,62 @@
...
+
+  function updateSplitPrice(){
+    var price = computeSplitPrice();
+    var oldPrice = parseFloat($('##{field}_price', form).val());
+    if (oldPrice == price){
+      window.alert(      [
+      'this button lets you quickly recalculate the split price, but the',
+      'price is still the same!  price = (supplier price/unit)',
+      'did you change unit or supplier price?',
+      ].join(' '));
```
wvengen: Candidate for i18n-js (see also `config/i18n-js.yml` for which strings are included in the js).

app/models/article_price.rb
```diff
@@ -21,11 +21,11 @@ class ArticlePrice < ApplicationRecord
...
   has_many :order_articles
 
-  localize_input_of :price, :tax, :deposit
+  localize_input_of :price, :tax, :deposit, :supplier_price
 
   validates_presence_of :price, :tax, :deposit, :unit_quantity
   validates_numericality_of :price, :greater_than_or_equal_to => 0
   validates_numericality_of :unit_quantity, :greater_than => 0
-  validates_numericality_of :deposit, :tax
+  validates_numericality_of :deposit, :tax, :supplier_price
```
wvengen: Would the supplier price also need a check for >= 0?

wvengen:
Regarding testing, I have issues running the integration tests locally as well, most of the time. Sometimes I run the tests on Docker, which seems to a bit better.

carchrae:
> Regarding testing, I have issues running the integration tests locally as well, most of the time. Sometimes I run the tests on Docker, which seems to a bit better.

thank you @wvengen for at least confirming my troubles are not just me.  the next time i get some time to work on foodsoft, i will try to get the same tests running locally.  i think it is great that the project has some tests but we should not have to struggle to reproduce the conditions in the travis runs.

app/views/shared/_article_fields_price.html.haml
carchrae: i've replaced with i18n.

as to the term, yes, technically correct,  however the UI doesn't show _unit_quantity_  rather it shows "Unit"

app/views/shared/_article_fields_price.html.haml
carchrae: agree and done

app/models/article_price.rb
carchrae: done

carchrae:
i don't have translations for :de or :nl however, i suspect i might do those poorly.  i could add google translate versions if that helps.

otherwise, this is ready to go i think.  test failure was actually important, so fixed.

Gemfile
```diff
@@ -115,3 +115,7 @@ group :test do
   gem 'apivore', require: false
   gem 'hashie', '~> 3.4.6', require: false # https://github.com/westfieldlabs/apivore/issues/114
 end
+
+gem "dotenv-rails", "~> 2.6"
+
+gem "pg", "~> 0.19.0"
```
carchrae: ah - maybe you don't want these in there.  i actually run foodsoft on postgres.  and use dotenv.  but these probably don't belong in this PR

carchrae:
hmm - no, i see i have committed more changes than i meant to.  let me revert some of those

wvengen:
Ok, I did some tests. This is what it looks like (when editing an order article).
image
Changing either _Supplier price_ or _Unit_ (which is unit quantity) shows a notice:
image
When you press _Recalculate Split Price_ (we shouldn't use capitals for all words; _Order article_ has a capital because it is a model name), the net price changes to 6.2 here and the notice disappears. When you press the button if it is ok, an alert shows _This button lets you quickly recalculate the price = (supplier price/unit), however the price is already correct._ (Perhaps the button could be disabled using a tooltip instead of an alert?)

In general I think it's pretty clear, but for one thing: I needed some time to understand here that the supplier price is for the full box, while the net price is for one item. I wonder if we can make that more clear here. Also, the unit _250 g_ doesn't play a role, I think that is also a little confusing. I don't have a clear answer yet, but some small UI changes can help I think. My first idea would be to group the supplier's price and the unit, so that they belong together, and have the other price fields in a different group - perhaps that makes it clear that the price is for the full unit line, and the other prices for one item. Somehow I'd like to convey that in the edit screen.

Another question is: do we always want to have a supplier's price? If it is empty now and you save the article, it will automatically be saved. Is that ok?
And if unit quantity is 1, we'd perhaps better hide the split functionality, right? For certain suppliers this isn't used at all, and I would like to keep it simple in that case (now you need to keep the net price and supplier price in sync by pressing the button or copying and pasting, or you'd get the yellow notice).

All in all, super, I think there is a good basis.

carchrae:
> My first idea would be to group the supplier's price and the unit, so that they belong together, and have the other price fields in a different group - perhaps that makes it clear that the price is for the full unit line, and the other prices for one item. Somehow I'd like to convey that in the edit screen.

great idea so :+1:  on that.  above is the supplier info, below is the ordergroup price.

> And if unit quantity is 1, we'd perhaps better hide the split functionality, right? For certain suppliers this isn't used at all, and I would like to keep it simple in that case (now you need to keep the net price and supplier price in sync by pressing the button or copying and pasting, or you'd get the yellow notice).

that is a good point.  if we did that, would it just be a UI hiding thing on this modal?  i think it might be quite complex to handle the different cases if the model code distinguishes between 1 and >1 in unit_quantities.

i agree with your comment on the labels - and also that perhaps the button to recalculate might be better located near the supplier price (like a little mini button there - or perhaps a lock/unlock ratio button - which would toggle auto-updating the prices - like when you resize an image (changing the height, will change the width if the ratio is locked))

carchrae:
i have a lot of balancing page improvements that stem from this change too.  one other aspect is the rounding errors due in rounding up to the next cent/penny.  i'm not sure if you noticed, but i always round up so the co-op makes a small profit on any error.  the reason being that we never want the co-op to keep taking a loss.  however, these rounding errors need to be considered when doing the balancing.

here is what that looks like in our current version (please ignore the "Balance Fix" line).  the rounding error here amounts to $0.24.  this order is perfectly balanced.

image

carchrae:
this is a tangent, but the balancing still needs more improvements.  i would like it to be clear to people doing this job that there are several things that need to be done

1) ensure that the co-op "supplier price" is equal to the invoice (this means that we have the correct prices)
2) ensure that the ordergroups(members) paid amount is equal to the price in 1.  this means we accounted for all items and allocations.  (eg, if things are short/missing, this won't be equal and needs to be fixed)

carchrae:
i'm using this in my branch for at least 4 years now, it is a pretty key feature, for the reasons stated in the PR description.

that said, i've no idea what state this PR is in.

i'm not clear how a "a standardised unit system" would help.  the big problem is that a supplier charges per case, but items appear on the site per unit.  if there are no rounding errors, this is fine, but that is often not the case.

lentschi:
> i'm not clear how a "a standardised unit system" would help.

The standardised unit system we're building, will allow converting between units used for the supplier and those used for group orders/billing. With this you can enter prices in the article form in meaningful ways which - I think - will prevent any rounding errors.

*However* I haven't tried out your original apple example to check if rounding errors still occur in our fork. I hope I'll be able to do so soon and keep you posted.

carchrae:
i don't think it is logically possible to get rid of the rounding error - it just needs to be represented/stored so you can be certain you are charging the right amount to your members.  

the rounding error is a consequence of splitting something and having only two decimal places (aka cents) to represent the divided amount.  the simple example, 1 euro for 3 apples.  each apple should be 1/3 euro, but we have to represent it as 34 cents (or 33 cents - but your group would slowly lose money if you round down).  if we charge members 33, the group loses 1 cent, if we charge members 34, the group earns 2 cents.  the supplier price mechanism i suggest is solely to track this information so you can be clear you balanced an order.

here is the screen of a successfully balanced order from my group.  (they get a nice picture of [what's going on he man](https://www.youtube.com/watch?v=rOU4yN_9KhE) when it is balanced as a reward).  the total rounding error was $3.90 for this order.

image

carchrae:
the conversion system looks interesting.  it is always a challenge to know what unit to offer members to purchase.  garlic might come in a 5lb box (or 25lb), but do people really want to buy 1lb of garlic.  i have a horrendous regex script for pulling units out of produce descriptions, but it has no sense of how much is a reasonable amount to offer people.  it would be interesting to offer people the ability to order whatever fraction of the box.  although, you also run into the issue of ensuring the demand adds up to a full box.

where i think your unit system could be most useful is when i need to swap an item with another item because it has become unavailable.  a common example is when you had posted a box of apples, say 100 CT apples (CT is piece/each), but now the supplier only has a case of 12 x 3LB bags of apples.  you know that both cases weigh the same amount in total, ~48LB, so you should be able to convert all of the orders from apples into 3LB bags.  (our group sometimes makes this kind of substitution when the supplier is out of stock, but we let the humans who divide up the food do the conversion)

i have a swap item page in my fork.  it is absolutely crucial for ordering produce (prices change every few days on some things, or become unavailable).  it shows alternatives based on it matching a unit - so for that reason, it would not show the apple example above (but it can handle different box sizes, see the avocados)

image

PR #638, opened 2019-03-24
don't hardcode the date format

carchrae:
title says it all.  this may be a nice default for someone, but it looked terrible for me.  anyway, that is taste, and not-hard-coding it means we can all be happy.  :smile_cat:

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,7 +4,7 @@
 module ApplicationHelper
 
   def format_time(time = Time.now)
-    I18n.l(time, :format => "%d.%m.%Y %H:%M") unless time.nil?
+    I18n.l(time, format: :short) unless time.nil?
   end
 
   def format_date(time = Time.now)


app/helpers/application_helper.rb
```diff
@@ -4,7 +4,7 @@
 module ApplicationHelper
 
   def format_time(time = Time.now)
-    I18n.l(time, :format => "%d.%m.%Y %H:%M") unless time.nil?
+    I18n.l(time, format: :short) unless time.nil?
```
wvengen: This leaves out the year, I think we need to include it in the default time format.

carchrae: a fair point.  let me look closer to see what a better default is.  alternatively, these could (perhaps should) be specified in `locales/en.yml` etc

wvengen: Yes, I'd be ok adding the year to the locale files for the short format.

kidhab: I think it can lead to confusion if the `:short` format will be extended via locales. Maybe another approach is to introduce a new date format. Please have a look at #997.

kidhab:
Done via #997

PR #639, opened 2019-03-24
allow all users to see article list (but not edit).  case insensitive…

carchrae:
i want everyone in my co-op to be able to view the catalogues for each supplier.  this allows that.  it also adds a case insensitive search on article name and number.

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 class ArticlesController < ApplicationController
-  before_filter :authenticate_article_meta, :find_supplier
+  before_filter :find_supplier
+  before_filter :authenticate_article_meta, :except => [:index]
 
   def index
     if params['sort']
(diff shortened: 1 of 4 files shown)


wvengen:
I think it's fine if everyone can see the whole article list
Do others agree?

carchrae:
the only possible objection (to object to myself) is that there is currently no way to see an index of the all suppliers.  in our co-op, i paste the link to a particular supplier into the email, but perhaps it would be good to have the supplier `index` page also have the same modification.  and likewise, add it to the navigation.

in our co-op we do this so people can see the whole catalog and make requests.

wvengen:
> There is currently no way to see an index of the all suppliers.

Hah, that's a good point :) I think it would be good to have the supplier index page accessible to all users, and indeed add it to the navigation. Maybe we'll need to revisit the navigation at some point, but let's look at that later then, when needed.

wvengen:
I've [asked](http://foodsoft.51229.x6.nabble.com/Shall-we-make-suppliers-and-articles-visible-for-all-users-by-default-td2127.html) the foodsoft-discuss mailing-list if this change is ok.

wvengen:
The mailing-list responded that we shouldn't show article prices by default, this can be sensitive.
Let's make this optional (I guess sharedlists would be something to add later).

carchrae:
ok, so if i add a` FoodsoftConfig` setting, that would work?

wvengen:
Yes!

yksflip:
Hey,

PR #666, opened 2019-05-13
Use AJAX for StockArticle.new in StockTaking.new

JuliusR:
The sidebar of `StockTaking.new` contains a link for creating stock articles. That link has to open an AJAX modal because there is no HTML view for `StockArticle.new`.


JuliusR:
This is just a quick fix to prevent users from experiencing the ugly "internal server error".

After this fix, the user can create new stock articles via AJAX. However, the new stock article won't show up immediately in the `StockTaking.new` form because there are no callbacks. The whole form is not optimal so I think this quick fix is fine for now.

The user can still obtain an "internal server error" when trying to open the AJAX links in new tabs, but this applies to many links. HTML fallbacks (if we care about that at all) require much more work in general.

wvengen:
One can subscribe to the JS event `StockArticle#create` to re-render the contents (or if that's too much work for now, reload the page as a quickfix), as done in the [stock index](https://github.com/foodcoops/foodsoft/blob/66e9629858db5c9b8b4726457aae801a5334a00e/app/views/stockit/index.html.haml#L9) and the [deliveries form](https://github.com/foodcoops/foodsoft/blob/66e9629858db5c9b8b4726457aae801a5334a00e/app/views/deliveries/_form.html.haml#L61).

app/controllers/deliveries_controller.rb
```diff
@@ -1,8 +1,6 @@
 # encoding: utf-8
 class DeliveriesController < ApplicationController
 
-  before_filter :find_supplier, :exclude => :fill_new_stock_article_form
```
JuliusR: Oops, that is an unrelated cleanup. I just found it by grep'ing "stock_article_form". Should I rebase and remove that change?

app/controllers/deliveries_controller.rb
```diff
@@ -18,7 +16,7 @@ def new
 
   def create
     @delivery = Delivery.new(params[:delivery])
-
```
JuliusR: My editor does this automatically such that the diff gets cluttered in this way. Should I turn it off and rebase?

app/controllers/stock_takings_controller.rb
```diff
@@ -10,46 +10,18 @@ def new
...
-  end
-
-  def drop_stock_change
-    stock_change = StockChange.find(params[:stock_change_id])
-    stock_change.destroy
-
-    render :update do |page|
-      page.visual_effect :DropOut, "stock_change_#{stock_change.id}"
-    end
-  end
```
wvengen: Aren't these methods used anywhere?

app/controllers/stock_takings_controller.rb
```diff
@@ -10,46 +10,18 @@ def new
     StockArticle.undeleted.each { |a| @stock_taking.stock_changes.build(:stock_article => a) }
   end
 
+  def new_on_stock_article_create # See publish/subscribe design pattern in /doc.
+    stock_article = StockArticle.find(params[:stock_article_id])
+    @stock_change = StockChange.new(:stock_article => stock_article)
+
+    render :layout => false
```
JuliusR: This line is just for consistency. It seems to work without this line. I have no idea if we should include that line or not.

app/controllers/deliveries_controller.rb
JuliusR: OK, I will rebase and omit this change because it is not related to the rest.

Let us finish some discussions here for the moment.

app/controllers/stock_takings_controller.rb
wvengen: I think it's a good idea to include it, to make sure no encompassing layout is used, just the template itself.

app/controllers/stock_takings_controller.rb
JuliusR: No, I do not know how to trigger them from the user interface. I also searched for, e.g, "grep -r stock_article_form ~/foodsoft". I had no useful matches (besides the routes, which I deleted, too).

Methods 1 and 2 should be handled in `controllers/stockit_controller.rb`.

Method 3 deletes individual StockChanges. However, the foodsoft does not allow to modify StockTakings (other than date and comment). It is possible to delete a StockTaking as a whole, though.

JuliusR:
Closed in favor of #667

PR #667, opened 2019-05-13
Fix and cleanup stock takings

JuliusR:
Rebase of #666, which had initially been a one-line fix.


wvengen:
Super, confirmed that it works on https://app.foodcoops.net/latest/demo/stock_articles
I guess it would be useful to do a Foodsoft release some time.
Are we ready for 4.7, or do we need a patch-release 4.6.1?

JuliusR:
> I guess it would be useful to do a Foodsoft release some time.
> Are we ready for 4.7, or do we need a patch-release 4.6.1?

Yes, there should definitely be a new release "some time". I do not know what is best in general. This specific patch can wait for the next normal release (I do not need it urgently, but I just stumbled over the bug and fixed it). Feel free to proceed as you like.

Issue #665, opened 2019-05-12
Unable to save an order on mobile device


MatCast:
When making an order on a mobile device the "save order" button disappears after adding a product to the order and there is no way to scroll down to click the "save order" button. This is because the `<div id="info-box">` gets filled in with the products informations added to the order and the "save order" button disappears (see pictures attached). This makes ordering with mobile only possible if requesting the desktop site.
![button](https://user-images.githubusercontent.com/32680270/57581089-b7e8d180-74b2-11e9-8d27-91ebdd46d0ae.png)
![no_button](https://user-images.githubusercontent.com/32680270/57581090-b919fe80-74b2-11e9-925f-1f36a677b0d7.png)

PR #668, opened 2019-05-15
Added css to improve smaller screen usability

MatCast:
Added some css to improve the order page from mobile and small screens.
Feel free edit the code or place it somewhere else in the css files.
Solves: https://github.com/foodcoops/foodsoft/issues/665

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -8,3 +8,10 @@
 *= require list.missing
 *= require recurring_select
 */
+/* Hide the orders article info for small screens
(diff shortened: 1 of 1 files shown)


wvengen:
Hi, thanks for your contribution! I've merged it (with some small changes). Super :)

flo-f:
Thank you very much for the contribution, @MatCast!

The current solution still displays `article-info` when I click a row or when I add/remove a product. Unfortunately, it also overlays the `total-sum` node with the order button (you can only spot the blue text link ending with `..n` in the right bottom):

image

image

Please see PR #677 for further details.

PR #670, opened 2019-06-13
Fix matching of uploaded articles by `order_number`

### LINKED ISSUE #588: Unexpected behaviour in articles upload
Hi there!

I (already for quite some time) experience an issue when using the option "Delete articles not in uploaded file." when uploading articles for a supllier from a csv file.

I provide this as test data: [Test.zip](https://github.com/foodcoops/foodsoft/files/2607642/Test.zip)

Now doing
2018-11-22 11_36_51-clipboard
after uploading test1.csv before

will result in
2018-11-22 11_37_10-foodsoft - upload articles

That's not what I would expect. In my opinion only article-ID 1 should be deleted.

Could you please check on that?

Thanks for your support! Michael

PS: Example can be found in the Plattsalat foodcoop.

JuliusR:
Fixes #588.

### DIFF
--- a/app/models/supplier.rb
+++ b/app/models/supplier.rb
@@ -103,7 +103,7 @@ class Supplier < ApplicationRecord
       all_order_numbers << article.order_number if article
     end
     if options[:outlist_absent]
-      outlisted_articles += articles.undeleted.where.not(id: all_order_numbers+[nil])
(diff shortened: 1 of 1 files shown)


wvengen:
Ouch, this was a serious bug, thanks for fixing!

PR #673, opened 2019-06-13
Allow to copy articles

### LINKED ISSUE #541: Copy this article and add as new after edit.
When entering articles from the website it would be great if you can copy one article to create the next, as many times for example I have articles in 1L and 5L bottles and I have to enter the same values again in most of the form.

JuliusR:
Fixes #541.

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -36,6 +36,11 @@ class ArticlesController < ApplicationController
     render :layout => false
   end
 
+  def copy
+    @article = @supplier.articles.find(params[:article_id]).dup
+    render :layout => false
+  end
+
   def create
     @article = Article.new(params[:article])
     if @article.valid? && @article.save
--- a/app/views/articles/_article.html.haml
+++ b/app/views/articles/_article.html.haml
@@ -11,7 +11,11 @@
       = number_to_currency(article.price)
   %td= number_to_percentage(article.tax) if article.tax != 0
   %td= number_to_currency(article.deposit) if article.deposit != 0
-  %td= link_to t('ui.edit'), edit_supplier_article_path(@supplier, article),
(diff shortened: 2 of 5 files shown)


JuliusR:
The article index already offers "edit" and "delete" buttons. This pull request adds a third button in between.

As a disadvantage, the table becomes even wider. I am not sure if this pull requests improves or worsens the usability on average. What do you think?

wvengen:
Could you make a screenshot (perhaps even of different window widths)?

JuliusR:
| Width in px | without copy | with copy |
|-|-|-|
| 375 |  375-without | 375-with |
| 800 | 800-without | 800-with |
| 1024 | 1024-without | 1024-with |
| 1280 | 1280-without | 1280-with |
| 1600 | 1600-without | 1600-with |

JuliusR:
After looking at it more carefully, I am not so much bothered about the wide tables. It seems that the real difficulty are the "line breaks" between the buttons. This increases the table row height and leads to a small vertical spacing of buttons. With my mouse, I can safely point on one of them, but what about touchscreen devices?

We can prevent the buttons from "line breaking" by putting them in separate columns, but that would always reserve lots of space for the buttons. We should also consider languages where "edit", "copy", and "delete" have more characters.

What are your thoughts?

f0k:
> It seems that the real difficulty are the "line breaks" between the buttons. 
> [...]
> We should also consider languages where "edit", "copy", and "delete" have more characters.
>
> What are your thoughts?

edit duplicate delete

wvengen:
Ah, it's buttons breaking over the line is even better now, I'd say.
Icons may be nice, especially if we do it consistenly across the app.
Fine for now, I'd say.

JuliusR:
I agree that we can merge this, and we can introduce icons later (if desired).

PR #677, opened 2019-08-26
Show article infos and order button on smaller screens

flo-f:
The solution from PR #665 still displays `article-info` when I click a row or when I add/remove a product. Unfortunately, it also overlays the `total-sum` node with the order button (you can only spot the blue text link ending with `..n` in the right bottom):

image

image


flo-f:
Actually there are still some TODOs. Sorry for creating the PR too early. I will close and recreate once my suggestion is fully ready...

PR #681, opened 2019-10-25
Bundle update, switch integration test drivers

wvengen:
- bundle update (important updates)
- get rid of warnings because of new conventions (like factory_bot)
- switch to [apparition](https://github.com/twalpole/apparition) integration test js driver (no webdriver install necessary) (the headless webkit option used became a dead end)
- silence mail logs during testing

TO DO:
- [x] bump required Ruby version, in docs, CI, docker (Capybara requirement)
- [x] update development docs (no webkit but chromium for integration tests)
- [x] make tests work in Docker development setup

There is still the occasional integration test failure, but that is still caught by rspec-rerun on CI (and we already had those).


paroga:
nice update! can you create separate changes? IMHO the `faker` update, the test driver switch and the bundle update could all be done in independent changes.
ad `puma`: i didn't test your change, but i had some troubles when enabling it on my Docker instance. I had to [change the `Procfile`](https://github.com/foodcoopsat/foodsoft/commit/a611e1d9030664f492f082f65e7973e86642769b#diff-5bd6b85c2d6fc987875f4bf82de2a15a) to get it running

wvengen:
Thanks for checking! Ouch, thanks for pointing out the `puma` change (the production container still would have worked, because it wouldn't have installed puma, but nonetheless).
I'll push a new version with separated commits shortly.

Gemfile.lock
```diff
@@ -142,36 +141,36 @@ GEM
     concurrent-ruby (1.1.5)
     connection_pool (2.2.2)
     content_for_in_controllers (0.0.2)
-    coveralls (0.8.22)
-      json (>= 1.8, < 3)
-      simplecov (~> 0.16.1)
-      term-ansicolor (~> 1.3)
-      thor (~> 0.19.4)
-      tins (~> 1.6)
+    coveralls (0.7.2)
```
paroga: this looks more link an downgrade than upgrade :joy:

spec/integration/balancing_spec.rb
```diff
@@ -167,8 +172,10 @@
     click_link I18n.t('finance.balancing.edit_results_by_articles.add_article')
     expect(page).to have_selector('form#new_order_article')
     within('#new_order_article') do
-      select new_article.name, :from => 'order_article_article_id'
+      find('#order_article_article_id').select(new_article.name)
+      sleep 0.1
```
paroga: i don't think that i need to mention that `sleep` looks wrong :wink:

Gemfile.lock
wvengen: that is weird ... thanks for catching this!

spec/integration/balancing_spec.rb
wvengen: thanks for mentioning, I'd like to look into this sometime
https://github.com/twalpole/apparition#timing-problems
https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends

Gemfile.lock
wvengen: I've tried to correct this, but there must be some combination of dependencies that doesn't work.
```
Bundler attempted to update coveralls but its version stayed the same
```

[`bundle-stats`](https://olivierlacan.com/posts/why-wont-bundle-update/) may be of help here.
[select2-rails](https://rubygems.org/gems/select2-rails) has `thor ~> 0.14` while the newer coveralls has `thor ~> 0.19.4` (from [thor 0.7.2](https://rubygems.org/gems/coveralls/versions/0.7.2) there is a thor version requirement). I guess we'll need a new select2 dependency :(

paroga: IMHO there is no need to spend any time on this. just merge it

PR #684, opened 2019-10-28
Set the common ActionMailer default_url_options only once

paroga:
There is no need to set options other than the foodcoop during every
change of the scope. This also makes sure that the correct options will
be loaded during testing to fix problems with newer rails versions.

### DIFF
--- a/spec/app_config.yml
+++ b/spec/app_config.yml
@@ -19,6 +19,8 @@ default: &defaults
   tax_default: 6.0
   email_sender: noreply@minimal.test
 
+  host: localhost
+
 
 development:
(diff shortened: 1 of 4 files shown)


spec/app_config.yml
```diff
@@ -19,6 +19,8 @@ default: &defaults
   tax_default: 6.0
   email_sender: noreply@minimal.test
 
+  host: localhost
+
```
wvengen: Why is this necessary?

paroga: newer rails versions will throw an error if the host is not set

Issue #686, opened 2019-10-29
Fix sometimes failing integration spec


wvengen:
After #681, the spec `spec/integration/balancing_spec.rb:169` "settling an order can add an article" fails more than is pleasant - even with rspec-rerun.
Let's investigate what happens here, and get it working more reliably.

See, for example, [this build](https://travis-ci.org/foodcoops/foodsoft/builds/604195376) (which keeps failing after restarting the build).

PR #693, opened 2019-10-30
Reduce integration spec error rate (#686)

wvengen:
I didn't find a way to properly fix this, and don't know why it doesn't work without the `sleep`. The [Capybara ajax notes](https://github.com/teamcapybara/capybara#asynchronous-javascript-ajax-and-friends) don't seem to help in this case.
Nevertheless, I've seen no errors in the too-often failing spec happening since.

### DIFF
--- a/spec/integration/balancing_spec.rb
+++ b/spec/integration/balancing_spec.rb
@@ -112,7 +112,7 @@ feature 'settling an order', js: true do
     end
     within("#edit_order_article_#{oa.id}") do
       find('#order_article_units_to_order').set(0)
-      sleep 0.2
+      sleep 0.25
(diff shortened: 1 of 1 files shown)


spec/integration/balancing_spec.rb
```diff
@@ -173,9 +173,8 @@
     expect(page).to have_selector('form#new_order_article')
     within('#new_order_article') do
       find('#order_article_article_id').select(new_article.name)
-      sleep 0.1
+      sleep 0.25
```
wvengen: It's just this one I focused on, and made everything `0.25` just as a bit extra. Let's add more when we see it's necessary.

PR #694, opened 2019-10-30
Fix line endings for Windows docker environment

haraldreingruber:
I was struggling setting up the development environment with docker-compose on Windows. Telling Git to not convert the line endings of shell scripts to Windows Style (CRLF) solved the issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+* text=auto
+*.sh text eol=lf
+proc-start text eol=lf
+Rakefile text eol=lf


wvengen:
If this makes it easier for Windows-users, I'd be happy to include it. Let's add a comment to the file explaining what this is for.
Also, just to check: is this needed for `Rakefile` (this is read by `rake`)?
I get the idea this good practice (though I haven't encountered it before). See:
https://rehansaeed.com/gitattributes-best-practices/
https://dev.to/deadlybyte/please-add-gitattributes-to-your-git-repository-1jld

haraldreingruber:
> Let's add a comment to the file explaining what this is for.

Done.

PR #696, opened 2019-11-01
Adopt model to allow FinancialTransaction without an ordergroup

paroga:
This will allow us to add accounting for the foodcoop itself, to support listing for spendings independent of the order (e.g. rent, electricity) and a kind of "double-entry accounting",

It's split into multiple commits for easier review (and maybe partial submit).

IMHO some of the names are not optimal, but i didn't had better ideas.

### DIFF
--- a/app/views/finance/financial_transactions/new_collection.html.haml
+++ b/app/views/finance/financial_transactions/new_collection.html.haml
@@ -68,6 +68,11 @@
   %p
     = link_to t('.new_ordergroup'), '#', 'data-add-transaction' => true, class: 'btn'
     = link_to t('.add_all_ordergroups'), '#', 'data-add-all-ordergroups' => true, class: 'btn'
+  - if FinancialTransaction.where(ordergroup: nil).any?
+    %p
+      %label
+        = check_box_tag :create_foodcoop_transaction, true, params[:create_foodcoop_transaction]
+        = t('.create_foodcoop_transaction')
   - if BankAccount.any?
     %p
       %label
--- a/app/controllers/finance/financial_transactions_controller.rb
+++ b/app/controllers/finance/financial_transactions_controller.rb
@@ -84,6 +84,7 @@ class Finance::FinancialTransactionsController < ApplicationController
 
     ActiveRecord::Base.transaction do
       financial_link = FinancialLink.new if params[:create_financial_link]
+      foodcoop_amount = 0
 
       params[:financial_transactions].each do |trans|
         # ignore empty amount fields ...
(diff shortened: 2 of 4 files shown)


paroga:
@wvengen ping

app/views/finance/financial_transactions/new_collection.html.haml
```diff
@@ -57,6 +57,10 @@
     = link_to t('.new_ordergroup'), '#', 'data-add-transaction' => true, class: 'btn'
     = link_to t('.add_all_ordergroups'), '#', 'data-add-all-ordergroups' => true, class: 'btn'
   - if BankAccount.any?
+    %p
+      %label
+        = check_box_tag :create_foodcoop_transaction, true, params[:create_foodcoop_transaction]
+        = t('.create_foodcoop_transaction')
```
wvengen, line 63: Why is this depending on BankAccount

paroga: it's misused as feature flag ATM. same applies to the financial links in the lines below. i'd like to avoid additional settings as much as possible, since people already don't get the existing ones :disappointed:

paroga:
i merged the first two commits already: 1a49bee42dc04397e5760dae53e74df0703030cc and 602f66324540146b8228491641995a46b2548a47

app/views/finance/financial_transactions/new_collection.html.haml
wvengen: Ah :) I agree the configuration has become quite complex. Can we also keep the code clear somehow?
To my mind, I'd rather put this in configuration, and leave making that better for another job.

Is there a reason not to show the option for foodcoop-transactions by default? Could we make it not stand in the way?
If it does clutter the interface (here and perhaps at other places), than maybe a feature toggle would be good anyhow.

wvengen: That said, I think "support double entry accounting" could be a useful configuration option.

paroga: i've change the check to existence of any foodcoop-related transaction for now. it's not perfect either, but it should do the job for now. i'd like to avoid addion of attional settings until everything related to finances is in `master`

wvengen: Ok, that makes sense. Would you mind creating an issue to keep track of that, so that these loose ends are somewhere public?

paroga: added #741

wvengen: Thanks!

PR #697, opened 2019-11-01
Add ExceptionNotifier to multicoops:run task

paroga:


lib/tasks/multicoops.rake
```diff
@@ -8,8 +8,12 @@ namespace :multicoops do
...
     task_to_run = ENV['TASK']
     FoodsoftConfig.each_coop do |coop|
-      rake_say "Run '#{task_to_run}' for #{coop}"
-      Rake::Task[task_to_run].execute
+      begin
+        rake_say "Run '#{task_to_run}' for #{coop}"
+        Rake::Task[task_to_run].execute
+      rescue => error
+        ExceptionNotifier.notify_exception(error, data: {foodcoop: coop})
+      end
```
wvengen: Do we need to re-raise the error, so you see something is amiss when no exception notifier is configured?

paroga: i see no need for that, since `multicoop` is usually only used in `production` environment, where it should be configured

wvengen: Some people may run without the notifier (for various reasons). Let's make it obvious when an error occurs, otherwise debugging becomes really hard. (Or maybe the notifier isn't configured properly.)

paroga: done

wvengen: great, but why not re-raise it in the `rescue` block? like [here](https://www.exceptionalcreatures.com/guides/advanced-rescue-and-raise.html#reraising-exceptions)

paroga: because then the first error will stop the whole process.
e.g. if banking import of one foodcooop fails all other will fail too, which is a problem

wvengen: super, I missed that before :)

PR #698, opened 2019-11-02
Add financial transaction class to supplier

paroga:
This allows the categorization of suppliers. For a better reporting
it is necessary to split the expenses of the invoices.
E.g. we want to be able to generate independent sums of general cost
like the rent or electricity and the cost of the bought articles.


paroga:
this has been superseded with #711

PR #700, opened 2019-11-03
Add weight to articles

paroga:
This is a preparation to calculate the transport cost according to the weight.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -24,6 +24,10 @@ module ApplicationHelper
     content_tag :span, number_to_currency(amount), class: class_name
   end
 
+  def format_weight(weight)
+    "#{weight}#{FoodsoftConfig[:weight_unit]}"
+  end
+
   # Creates ajax-controlled-links for pagination
   def pagination_links_remote(collection, options = {})
     per_page = options[:per_page] || @per_page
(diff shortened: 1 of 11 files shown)


wvengen:
Quick note: in many cases, `unit` will already contain the weight. Some idea on how these fields are meant to be used, would make sense, I think.

paroga:
unit != weight. since it's an optional field anyway i see no need for further hints in the UI. IMO this is an point for the (improvable) user documentation at https://foodcoops.readthedocs.io/

wvengen:
I do like the idea, but would be a bit hesitant to add another field to the article dialog that may not be used at for many, as the dialog is often used. So perhaps _another_ feature toggle may make sense? I'd rather have the complexity for admins once than everytime someone adds or edits an article.

RayOei:
The use case is not defined or clear to me. 
And is this still relevant with the new #1058 for instance?

lentschi:
> The use case is not defined or clear to me.

@RayOei Hard to tell as paroga is no longer actively working on foodsoft.
From what I can glean from the source code, he might have intended to display how much a `GroupOrder`'s contents will weigh in total.
If so, this is _not_ yet included in https://github.com/foodcoops/foodsoft/issues/1058 - however if we want this, it would be much easier and cleaner to implement based on it (without the extra `weight` field in the `Article`'s form). paroga's implemention wouldn't make any sense on top of [this PR](https://github.com/foodcoops/foodsoft/issues/1058) and since we all seem keen on merging that, I'm closing this PR. (If you disagree, @wvengen, please just reopen.)

PR #702, opened 2019-11-03
Add default roles for users

paroga:
Some foodcoops want to give some roles to every member by default,
without adding every users to a workgroup.


wvengen:
Nice simple change.
But let's consider the use-case here: does a foodcoop want allow people to see everything, or also to allow everyone to change articles in an open order, for example?

paroga:
yes, in some foodcoops EVERYBODY has the equal rights. making our roles more fine granted sounds like an valueable independent task

PR #705, opened 2019-11-11
Add .ruby-version file

mamhoff:
I use the `rbenv` Ruby version manager, and adding this file makes that
work.


paroga:
thanks for your contribution

PR #708, opened 2019-11-11
Remove `config.mass_assignment_sanitizer from development.rb sample

mamhoff:
This configuration option has been removed in Rails 4; I've only found
references to it from people running Rails 3.

### DIFF
--- a/config/environments/development.rb.SAMPLE
+++ b/config/environments/development.rb.SAMPLE
@@ -23,9 +23,6 @@ Foodsoft::Application.configure do
   # Print deprecation notices to the Rails logger.
   config.active_support.deprecation = :log
 
-  # Raise exception on mass assignment protection for Active Record models
-  config.active_record.mass_assignment_sanitizer = :strict
-
   # Raise an error on page load if there are pending migrations
   config.active_record.migration_error = :page_load
 


paroga:
i agree, but i like to keep it until we finally switched to strong parameters in the whole project and remove all remaining stuff from rails 3

mamhoff:
Hm, I don't think I explained myself well then: This is in the sample `development.rb` file that gets copied over. Because the method is gone since Rails 4, the result is: 

```
➜  foodsoft git:(master) ✗ be rake db:create
rake aborted!
NoMethodError: undefined method `mass_assignment_sanitizer=' for ActiveRecord::Base:Class
/home/anselm/code/foodsoft/config/environment.rb:5:in `<top (required)>'
/home/anselm/.rbenv/versions/2.4.6/bin/bundle:23:in `load'
/home/anselm/.rbenv/versions/2.4.6/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate => environment
(See full trace by running task with --trace)
```
So the app will refuse to even start. Not a great onboarding experience :crying_cat_face: 

Given that this primarily affects first-time contributors, maybe make an exception from the rule that nothing can be removed until everything is gone? :heart:

paroga:
sorry, my fault. i thought it was in `production.rb`. thanks for the PR!

PR #711, opened 2019-11-17
Add SupplierCategory

paroga:
This allows the categorization of suppliers. For a better reporting
it is necessary to split the expenses of the invoices.
E.g. we want to be able to generate independent sums of general cost
like the rent or electricity and the cost of the bought articles.

### DIFF
--- a/app/controllers/suppliers_controller.rb
+++ b/app/controllers/suppliers_controller.rb
@@ -27,6 +27,7 @@ class SuppliersController < ApplicationController
 
   def create
     @supplier = Supplier.new(supplier_params)
+    @supplier.supplier_category ||= SupplierCategory.first
     if @supplier.save
       flash[:notice] = I18n.t('suppliers.create.notice')
       redirect_to suppliers_path
@@ -70,7 +71,7 @@ class SuppliersController < ApplicationController
     params
       .require(:supplier)
       .permit(:name, :address, :phone, :phone2, :fax, :email, :url, :contact_person, :customer_number,
-              :iban, :custom_fields, :delivery_days, :order_howto, :note,
+              :iban, :custom_fields, :delivery_days, :order_howto, :note, :supplier_category_id,
               :shared_supplier_id, :min_order_quantity, :shared_sync_method)
   end
 
--- a/app/models/financial_transaction_class.rb
+++ b/app/models/financial_transaction_class.rb
@@ -1,5 +1,6 @@
 class FinancialTransactionClass < ApplicationRecord
   has_many :financial_transaction_types, dependent: :destroy
+  has_many :supplier_category, dependent: :restrict_with_exception
 
   validates :name, presence: true
   validates_uniqueness_of :name
--- a/app/models/supplier.rb
+++ b/app/models/supplier.rb
@@ -8,6 +8,7 @@ class Supplier < ApplicationRecord
   has_many :orders
   has_many :deliveries
   has_many :invoices
+  belongs_to :supplier_category
   belongs_to :shared_supplier  # for the sharedLists-App
 
   validates :name, :presence => true, :length => { :in => 4..30 }
--- a/app/models/supplier_category.rb
+++ b/app/models/supplier_category.rb
@@ -1,0 +1,19 @@
+class SupplierCategory < ActiveRecord::Base
+
+  belongs_to :financial_transaction_class
+  has_many :suppliers
+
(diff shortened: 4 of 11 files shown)


app/controllers/suppliers_controller.rb
```diff
@@ -27,6 +27,7 @@ def new
 
   def create
     @supplier = Supplier.new(supplier_params)
+    @supplier.supplier_category ||= SupplierCategory.first
```
wvengen, line 30: You don't want to leave it blank? At the beginning there is no supplier category anyway.

app/models/financial_transaction_class.rb
```diff
@@ -1,5 +1,6 @@
 class FinancialTransactionClass < ApplicationRecord
   has_many :financial_transaction_types, dependent: :destroy
+  has_many :suppliers
```
wvengen, line 3: Does it make sense to add `dependent` here?

app/models/supplier.rb
```diff
@@ -8,6 +8,7 @@ class Supplier < ApplicationRecord
   has_many :orders
   has_many :deliveries
   has_many :invoices
+  belongs_to :supplier_category
```
wvengen, line 11: idem

app/models/supplier_category.rb
```diff
@@ -0,0 +1,19 @@
+class SupplierCategory < ActiveRecord::Base
+
+  belongs_to :financial_transaction_class
+  has_many :suppliers
```
wvengen, line 4: idem

db/migrate/20181201000400_create_supplier_categories.rb
```diff
@@ -0,0 +1,28 @@
...
+  class SupplierCategory < ActiveRecord::Base
+  end
+  class Supplier < ActiveRecord::Base
+  end
+
+  def change
+    create_table :supplier_categories do |t|
+      t.string :name, null: false
+      t.string :description
+      t.references :financial_transaction_class, null: false
```
wvengen, line 13: Just to be sure: can there be zero transaction classes?
I can image supplier categories to make sense without them.

app/controllers/suppliers_controller.rb
paroga: i'd like to add a default category, which needs to exist, so there are no null values (i think i need to adopt the migrations to show that :roll_eyes: )

db/migrate/20181201000400_create_supplier_categories.rb
paroga: i think that the class is mandatory - otherwise the FinanceReport (you can find it in the [Austrian fork](https://github.com/foodcoopsat/foodsoft/) for now) can not assign the invoices to the correct sum
example: you have three supplier_categories ("food", "rent", "utils") and two finacial_transaction_classes ("food", "membership"). "rent" is used for the rent and electricity of the room, "utils" is used when new fridges or storage boxes are bought, while "food" is used for the "normal" consumption. at the end of the year we want to know how much of our membership fees (all in the "membership" class) are remaining. to do that we need to know to which class the category belongs.

wvengen: We also require articles to have a category, and by default there is just an "Other" category. I think we could take the same approach here. And then indeed the feature flag would be if there is more than one supplier category.

Thanks for explaining a bit :+1:

app/controllers/suppliers_controller.rb
paroga: it was in the migration already...

app/models/financial_transaction_class.rb
paroga: added

app/models/supplier.rb
paroga: it makes only sense for `has_many`

app/models/supplier_category.rb
paroga: there is a custom `check_for_associated_suppliers` for it

app/controllers/suppliers_controller.rb
wvengen: oops, good job

twothreenine:
This sounds like an interesting feature.
As I see it there is no UI for creating/managing supplier categories yet, right?
I think categories could also be useful to set options for a number of suppliers but not all, but that's another topic.

Issue #560, opened 2018-09-30
Sortable tables


vitus-a:
Only few columns in some tables are sortable. Why? 
Some columns/tables which would make administration easier, if they are sortable:

- Foodcoop - Members - all columns
- Foodcoop - Ordergroups - "Last activity" and "Last order".
- Administration - Users - "Email" and "Last activity"
- Administration - Ordergroups - "Last Activity"

haraldreingruber:
Hi @vitus-a,

I've taken a look at your improvement suggestion. I was able to make to following columns sortable:

User/Members lists:
- Name
- Email
- Last Activity

Ordergroup lists:
- Name

I am quite happy that I was able to solve it at least partly 😃 I've asked for help (#716) to make "Last activity" and "Last order" also sortable (for Ordergroups).

Do you think this would cover the most important columns for the first version?

vitus-a:
I think so. Sortable "Last activity" and "Last order" would be great to find inactive users. Thanks for your efforts.

haraldreingruber:
This feature is more or less finished, except a final review of the latest changes in PR #716.

Depending on the deployment, it's probably possible to try out the improvements from the branch of my fork...

haraldreingruber:
@vitus-a the improvement is merged to the main branch. Do you know how your deployment is updated? 
Could you test the change once you have the latest version?

vitus-a:
@haraldreingruber the improvement is already implemented in our deployment and seems to work fine. thanks!

haraldreingruber:
@wvengen @paroga I guess this issue can be closed then.

PR #716, opened 2020-01-17
Make columns of user and ordergroup lists sortable

haraldreingruber:
This commit implements the sort functionality for the user lists (by name, email, last_activity) and ordergroup lists (by name).
It is a first attempt addressing issue #560.

What is still missing is to make **last_activity** and **last_order** columns sortable for ordergoup lists, which is a bit more difficult as those attributes are not stored in the users/ordergroup table itself.

It's the first time I am working with Ruby and Rails, would be really cool if somebody could have a look if my current changes are okay or can be improved somehow and help me with the missing functionality.


haraldreingruber:
@wvengen or @paroga could you help me with making order groups also sortable by last_activity and last_order? Maybe we could do a 30min screen sharing session to get me started.

Thanks, Harald

haraldreingruber:
Hey @wvengen @paroga, I've finally managed to a fix my dev environment and dedicate some time to figure out how to sort models with associations.

@wvengen, I think I've been able to implement your previous suggestion. Please let me know if this covers what you wanted to see, or if you have more improvement ideas.

I've used the `table_print` gem for debugging the queries in the rails console. Do you think it might be useful to keep it? I guess you know it, or use other/better ones....

Also, I've extended the `small.en.seeds.rb` to make testing easier. Does it make sense to keep it as well? 

Looking forward to your feedback!

app/views/admin/users/_users.html.haml
```diff
@@ -6,10 +6,10 @@
     %tr
       - if FoodsoftConfig[:use_nick]
         %th= heading_helper User, :nick
```
paroga, line 8: why not for `nick`?

app/controllers/foodcoop/users_controller.rb
```diff
@@ -1,6 +1,25 @@
...
+             end
+           else
+             "first_name, last_name"
+           end
+
+    @users = case params["sort"]
+             when "ordergroup", "ordergroup_reverse" then User.left_joins(:groups).where("groups.type = 'Ordergroup'").undeleted.order(sort).distinct
+             else
+               User.undeleted.order(sort)
+             end
```
paroga, line 22: please share at least the `.undeleted.order(sort)`. i'd just remove the whole `case` here and do the join always to keep the code simpler. also i'd suggest to avoid the strings in the `where` and use `where(groups:{type:'Ordergroup'})` instead

app/controllers/admin/users_controller.rb
```diff
@@ -2,7 +2,21 @@ class Admin::UsersController < Admin::BaseController
...
+             when "last_activity" then "last_activity"
+             when "name_reverse" then "first_name DESC, last_name DESC"
+             when "email_reverse" then "email DESC"
+             when "last_activity_reverse" then "last_activity DESC"
+             end
+           else
+             "first_name, last_name"
+           end
+
+    @users = params[:show_deleted] ? User.deleted.order(sort) : User.undeleted.order(sort)
```
paroga, line 18: please share the `order(sort)` like:
```ruby
@users = params[:show_deleted] ? User.deleted : User.undeleted
@users = @users.order(sort)
```

or merge it into line 20

app/controllers/foodcoop/ordergroups_controller.rb
```diff
@@ -1,6 +1,24 @@
...
+             end
+           else
+             "name"
+           end
+
+    @ordergroups = case params["sort"]
+                   when "last_user_activity", "last_user_activity_reverse" then Ordergroup.left_joins(:users).group("groups.id").undeleted.order(sort).distinct
+                   when "last_order", "last_order_reverse" then Ordergroup.left_joins(:orders).group("groups.id").undeleted.order(sort).distinct
+                   else
+                     Ordergroup.undeleted.order(sort)
```
paroga, line 20: please share the common code (see my other comments)

app/views/foodcoop/users/_users.html.haml
```diff
@@ -6,10 +6,10 @@
     %tr
       - if FoodsoftConfig[:use_nick]
         %th= heading_helper User, :nick
```
paroga, line 8: why not for `nick`?

app/controllers/foodcoop/ordergroups_controller.rb
```diff
@@ -1,6 +1,24 @@
...
   def index
-    @ordergroups = Ordergroup.undeleted.order('name')
+    sort = if params["sort"]
+             case params["sort"]
+             when "name" then "name"
+             when "name_reverse" then "name DESC"
+             when "last_user_activity" then "max(users.last_activity)"
+             when "last_user_activity_reverse" then "max(users.last_activity) DESC"
+             when "last_order" then "max(orders.starts)"
+             when "last_order_reverse" then "max(orders.starts) DESC"
```
paroga, line 10: see my other comment about `last_order`

app/controllers/admin/ordergroups_controller.rb
```diff
@@ -2,7 +2,27 @@ class Admin::OrdergroupsController < Admin::BaseController
...
+           else
+             "name"
+           end
+
+    @ordergroups = case params["sort"]
+                   when "members_count", "members_count_reverse", "last_user_activity", "last_user_activity_reverse" then Ordergroup.left_joins(:users).group("groups.id").undeleted.order(sort).distinct
+                   when "last_order", "last_order_reverse" then Ordergroup.left_joins(:orders).group("groups.id").undeleted.order(sort).distinct
+                   else
+                     Ordergroup.undeleted.order(sort)
+                   end
```
paroga, line 25: please share the common code (see other comments)

paroga (changes requested):

app/views/foodcoop/users/_users.html.haml
haraldreingruber: Done!

app/controllers/foodcoop/ordergroups_controller.rb
haraldreingruber: Done!

app/controllers/admin/ordergroups_controller.rb
haraldreingruber: Done!

app/controllers/admin/users_controller.rb
haraldreingruber: Done!

app/controllers/foodcoop/users_controller.rb
haraldreingruber: Done!

app/views/admin/users/_users.html.haml
haraldreingruber: Done!

app/views/admin/ordergroups/_ordergroups.html.haml
```diff
@@ -17,6 +18,7 @@
           %abbr{title: ordergroup_members_title(ordergroup)}= ordergroup.users.size
         %td= link_to_gmaps ordergroup.contact_address
         %td= format_date ordergroup.last_user_activity
+        %td= format_date ordergroup.last_order.try(:starts)
```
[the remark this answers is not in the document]
haraldreingruber: Okay, Happy to look into this, in a separate PR.

app/controllers/foodcoop/ordergroups_controller.rb
haraldreingruber: You mean also the `ordergroup.last_group_order.try(:updated_on)` idea right?

Gemfile
```diff
@@ -55,6 +55,7 @@ gem 'gaffe'
 gem 'ruby-filemagic'
 gem 'mime-types'
 gem 'midi-smtp-server'
+gem "table_print"
```
paroga, line 58: adding `table_print` is IMO completely unrelated with the other changes. please create a separate PR and add it to the [`development` group in the Gemfile](https://github.com/foodcoops/foodsoft/blob/1e883f05c0c59641af9f9da2d1b039f752dc1f21/Gemfile#L74), since it's not required for a production deployment

app/controllers/admin/ordergroups_controller.rb
```diff
@@ -2,7 +2,28 @@ class Admin::OrdergroupsController < Admin::BaseController
   inherit_resources
 
   def index
-    @ordergroups = Ordergroup.undeleted.order('name ASC')
+    sort = if params["sort"]
+             case params["sort"]
```
paroga, line 6: could you address the rubocop issues?

app/controllers/foodcoop/ordergroups_controller.rb
paroga: yes. i'd suggest to add sorting for `last_order` in a second step, since it's not so straight forward as the other ones

app/controllers/foodcoop/ordergroups_controller.rb
```diff
@@ -1,6 +1,25 @@
 class Foodcoop::OrdergroupsController < ApplicationController
   def index
-    @ordergroups = Ordergroup.undeleted.order('name')
+    sort = if params["sort"]
```
paroga, line 3: i think this is more or less the same code as in `app/controllers/admin/ordergroups_controller.rb`. What do you think about moving the sorting into the model instead, so code in the controller will reduce do somthing like `@ordergroups = Ordergroup.undeleted.params_sort(params["sort"] || "name")`? (if you have a better name than `params_sort`, just use it). Moving it to the model has the additional advantage that it's much much easier to write tests for it.

app/controllers/foodcoop/ordergroups_controller.rb
```diff
@@ -1,6 +1,25 @@
...
+             when "last_user_activity_reverse" then "max(users.last_activity) DESC"
+             when "last_order" then "max(orders.starts)"
+             when "last_order_reverse" then "max(orders.starts) DESC"
+             end
+           else
+             "name"
+           end
+
+    @ordergroups = case params["sort"]
+                   when "last_user_activity", "last_user_activity_reverse" then Ordergroup.left_joins(:users).group("groups.id")
```
paroga, line 17: would that work with `.distinct` instead of `.group("groups.id")` too?

app/views/admin/ordergroups/_ordergroups.html.haml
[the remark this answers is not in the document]
paroga: i'd suggest to remove it from this PR then

paroga (changes requested):
thanks for the changes, looks already a lot better, just some findings. please let me know, if you have different opinions

Gemfile
haraldreingruber: Done. Extracted to #935.

app/controllers/foodcoop/ordergroups_controller.rb
haraldreingruber: No, `.distinct` doesn't seem to return the expected result.

app/controllers/foodcoop/ordergroups_controller.rb
haraldreingruber: done.

app/controllers/admin/ordergroups_controller.rb
haraldreingruber: done.

app/views/admin/ordergroups/_ordergroups.html.haml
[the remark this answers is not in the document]
haraldreingruber: As this column is already there in the `foodcoop/ordergroups` view, I would prefer to include the current sorting solution at least there. Would you prefer not to add the column here until tackling your suggested improvement idea?

app/controllers/foodcoop/ordergroups_controller.rb
haraldreingruber: See comment in `app/views/admin/ordergroups/_ordergroups.html.haml` above.

app/controllers/foodcoop/users_controller.rb
```diff
@@ -1,6 +1,18 @@
 class Foodcoop::UsersController < ApplicationController
   def index
-    @users = User.undeleted.natural_order
+    sort_param_map = {
```
paroga, line 3: what about merging them with the `sort_param_map` in `app/controllers/admin/users_controller.rb` and move it to the model too?

app/models/ordergroup.rb
```diff
@@ -148,6 +148,27 @@ def account_updated
...
+      "last_user_activity_reverse" => "max(users.last_activity) DESC",
+      "last_order" => "max(orders.starts)",
+      "last_order_reverse" => "max(orders.starts) DESC"
+    }
+
+    @ordergroup = case param
+                  when "members_count", "members_count_reverse", "last_user_activity", "last_user_activity_reverse" then @ordergroup.left_joins(:users).group("groups.id")
+                  when "last_order", "last_order_reverse" then @ordergroup.left_joins(:orders).group("groups.id")
+                  else
+                    Ordergroup
```
paroga, line 167: instead of `Ordergroup` you should use `self` here

app/models/ordergroup.rb
```diff
@@ -148,6 +148,27 @@ def account_updated
...
+      "name_reverse" => "name DESC",
+      "members_count" => "count(*)",
+      "members_count_reverse" => "count(*) DESC",
+      "last_user_activity" => "max(users.last_activity)",
+      "last_user_activity_reverse" => "max(users.last_activity) DESC",
+      "last_order" => "max(orders.starts)",
+      "last_order_reverse" => "max(orders.starts) DESC"
+    }
+
+    @ordergroup = case param
```
paroga, line 163: i'd avoid to use `@ordergroup` here and just use a normal variable (e.g. `ret`)

app/models/ordergroup.rb
```diff
@@ -148,6 +148,27 @@ def account_updated
...
+                  when "members_count", "members_count_reverse", "last_user_activity", "last_user_activity_reverse" then @ordergroup.left_joins(:users).group("groups.id")
+                  when "last_order", "last_order_reverse" then @ordergroup.left_joins(:orders).group("groups.id")
```
paroga, line 165: i don't see where `@ordergroup` is set before. i think you missed to adopt this code

app/models/ordergroup.rb
```diff
@@ -148,6 +148,27 @@ def account_updated
...
+                  else
+                    Ordergroup
+                  end
+    @ordergroup.order(sort_param_map[param])
+  end
+
```
paroga, line 171: what do you think about writing it the following way?

```ruby
ret = self
ret = ret.left_joins(:users).group("groups.id") if param.starts_with?("members_count", "last_user_activity")
ret = ret.left_joins(:orders).group("groups.id") if param.starts_with?("last_order")
ret.order(sort_param_map[param])
```

app/views/admin/ordergroups/_ordergroups.html.haml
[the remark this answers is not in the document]
paroga: i'd not at it, but i'm completely ok with it if you want to

app/controllers/foodcoop/ordergroups_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Foodcoop::OrdergroupsController < ApplicationController
   def index
-    @ordergroups = Ordergroup.undeleted.order('name')
+    @ordergroups = Ordergroup.undeleted.sort_by_param(params["sort"] || "name")
```
paroga, line 3: i know that it was my idea to to write it like you did, but what do you think about a) passing `params` directly and b) move the fallback sorting (`"name"`) into the model too?

paroga (changes requested):
i think i found an error and have some additional ideas how to make the code more readable. (the comments apply to both models)

do you see a chance to add some basic tests too?

app/models/ordergroup.rb
haraldreingruber: I like it. It's much more readable. I didn't know about these expression modifiers.

app/controllers/foodcoop/ordergroups_controller.rb
haraldreingruber: Sure. Things often get clearer once it's glued into code ;-)
I guess a default value for the method parameter makes most sense?

app/models/ordergroup.rb
haraldreingruber: Thanks, good catch!

app/models/ordergroup.rb
haraldreingruber: If it's okay I would name the variable `result`, as I like to avoid abbreviations. But as naming is rather subjective, I am fine with either way, especially if `ret` is commonly used in the project?

app/views/admin/ordergroups/_ordergroups.html.haml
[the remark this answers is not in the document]
haraldreingruber: Okay. Then let's keep it for now, and improve it in a later PR.

app/controllers/foodcoop/users_controller.rb
haraldreingruber: Done!

app/models/ordergroup.rb
```diff
@@ -148,6 +148,26 @@ def account_updated
...
+      "members_count" => "count(*)",
+      "members_count_reverse" => "count(*) DESC",
+      "last_user_activity" => "max(users.last_activity)",
+      "last_user_activity_reverse" => "max(users.last_activity) DESC",
+      "last_order" => "max(orders.starts)",
+      "last_order_reverse" => "max(orders.starts) DESC"
+    }
+
+    result = self
+    result = result.left_joins(:users).group("groups.id") if param && param.starts_with?("members_count", "last_user_activity")
```
paroga, line 166: you set `param` already in line 152, so there is not need for the `param &&`

app/models/ordergroup.rb
paroga: you'll find both in the code, so just go with what you like more

app/models/ordergroup.rb
```diff
@@ -148,6 +148,26 @@ def account_updated
     financial_transactions.last.try(:created_on) || created_on
   end
 
+  def self.sort_by_param(param)
+    param ||= "name"
```
paroga, line 152: i personally would prefer `sort_by_param(param = "name")`, since it's IMHO clearer

paroga (approved):
just two minor ideas, if you like the current version more, i'll merge it as it is

app/models/ordergroup.rb
haraldreingruber: That was also my first idea. But it turned out this only works when the parameter is not used, but not when NIL is passed.
Let me know if I might have missed something?

app/models/ordergroup.rb
haraldreingruber: Oh, that slipped in while I realized the default parameter didn't work. Thanks for pointing it out.

haraldreingruber:
I started working on the tests. Feel free to wait for them before merging the PR. Maybe it even exposes some edge cases we still missed...

haraldreingruber:
@paroga The unit tests are ready for review. I've also fixed another issue with sorting by ordergroup member count. 
Please let me know if there are test cases the I might have missed.

Also, I've noticed that Rails is logging a warning that using non-attribute arguments for ordering is going to be deprecated in Rails 6, because it could allow SQL injections. As far as I understand the code, in this case it should be safe, because the controller params are not directly passed to the query, only used for matching the actual ordering string.
So I guess we could fix this with wrappign it in a `Arel.sql()` call as suggested in the warning.
What do you think?

Here is one example of the full message:
> .DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): "max(users.last_activity)". Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql(). (called from sort_by_param at /app/app/models/ordergroup.rb:168)

paroga:
If it will be disallowed in Rails 6.0 anyway it would be nice to do it already in the expected way. Otherwise it will block the update. otherwise i think it's ready to get merged

haraldreingruber:
Done.
@paroga please let me know if you think the comments I've added, about using `Arel.sql()` with caution are valuable. If not, I don't mind removing them.

haraldreingruber:
@paroga Have you had a chance to review my latest changes?

haraldreingruber:
Anything I or somebody else can help with, in order to get this PR merged?

Issue #718, opened 2020-01-29
deleting an invoice does not clean up order.invoice_id


carchrae:
today, one of my users accidentally deleted an invoice that was associated with an order.  the invoice deletes just fine, but the order's `invoice_id` property is left pointing to the (now-deleted) invoice.  this means the user is then unable to create a new invoice that refers to the order (since the code thinks it is already linked to one, even though it is a deleted one).

for example, the code below will never populate the select (on the route `invoices/new`) with the order, even if though the order points to a now deleted invoice.

https://github.com/foodcoops/foodsoft/blob/b0ae35f73a498aaafc51b0ea586988fa103edfee/app/controllers/finance/invoices_controller.rb#L35

i am not sure of the best fix for this.  should deleting an invoice set `order.invoice_id` to `nil`? that would seem to be the most precise fix.  i suspect there is some rails magic that would do this, but otherwise, i suppose an active record callback could handle it.

or alternatively, change the linked code above to check if `invoice_id` points to a non-deleted invoice.  but this seems patchy/slow and i wonder if there are other bugs caused by `invoice_id` pointing to a deleted record.

wvengen:
Good one! The invoice_id should indeed be nullified.

PR #724, opened 2020-04-08
not a very nice fix to invoice_id pointing at deleted invoice

carchrae:
a quick fix for #718 

i'm not particularly happy with this, but it does the job in my use case.  this fix simply checks that when you visit the balancing page for an order, if the order has an invoice_id set but that invoice does not exist, it clears the invoice_id.  i dislike the code for all the reasons i hate logic in controllers.  i like the code because it is contained, predictable, and actually fixed my issue.

a better fix would be to do the cleanup when an invoice is deleted.  why didn't i do that?  well, it seems there are quite a few ways you could do it.  perhaps a callback on https://apidock.com/rails/ActiveRecord/Callbacks/after_destroy which would search for any references to the now-deleted invoice and clear them.

but then i also wondered if this was a deeper issue with how the active record relation was setup in the first place.  an invoice to order is a 1-N relationship, meaning one invoice could cover many orders (in practice for us, we never have this case, it is always 1-1, but that doesn't matter).  so perhaps there is a more elegant bit of sugar that could be added on this line to clean up the id when/if the invoice is deleted.  https://github.com/foodcoops/foodsoft/blob/739914ad430544c1e3126262bd614e3cecd4d599/app/models/order.rb#L14

### DIFF
--- a/app/controllers/finance/balancing_controller.rb
+++ b/app/controllers/finance/balancing_controller.rb
@@ -7,6 +7,14 @@ class Finance::BalancingController < Finance::BaseController
 
   def new
     @order = Order.find(params[:order_id])
+
+    # this is unpleasant - and not where this code should belong, but solves the case where
+    # an order points to an invoice that has been deleted
+    if @order.invoice.nil? && !@order.invoice_id.nil?
+      @order.invoice_id = nil
+      @order.save
+    end
+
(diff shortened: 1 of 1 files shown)


carchrae:
thanks @paroga - i knew there must be a proper rails way to do that, i just wasn't sure what it was.

paroga:
i've created #740 to fix this

PR #733, opened 2020-06-28
Fixes #732

### LINKED ISSUE #732: docker-compose DEV Setup: Running as User with UID/GID that differs from 1000/1000 fails
When following [the current docs](https://github.com/foodcoops/foodsoft/blob/9897c1056fbdc047243ea1feb8f6a57be79ea3a2/doc/SETUP_DEVELOPMENT_DOCKER.md) on how to run foodsoft with docker-compose, I get the error `Rails Error: Unable to access log file. Please ensure that /home/app/src/log/development.log exists and is writable`.

I found out, that this is due to the fact, that my local Linux user doesn't have the default uid/gid of 1000/1000, but 1001/1001 instead.

lentschi:
The solution isn't 100% perfect as omitting those environment variables currently shows a warning. (This could be bypassed, by adding env defaults in `docker-compose-dev.yml` instead of `Dockerfile-dev`, but this isn't supported in compose version 2. Upgrading to version 3 would need refactoring as `extends` is no longer supported.)


lentschi:
Ah - sorry, I accidentally did the PR on my master branch and now it includes a lot of changes.
Will close this one and  create another...

PR #735, opened 2020-07-12
Fixes #732

### LINKED ISSUE #732: docker-compose DEV Setup: Running as User with UID/GID that differs from 1000/1000 fails

lentschi:
The solution isn't 100% perfect as omitting those environment variables currently shows a warning. (This could be bypassed, by adding env defaults in `docker-compose-dev.yml` instead of `Dockerfile-dev`, but this isn't supported in compose version 2. Upgrading to version 3 would need refactoring as `extends` is no longer supported.)


paroga:
IMHO the correct fix would be to just change the file/directory permissions of the log directory to make it writeable. Or do I miss something?

wvengen:
@lentschi what's your use-case exactly? I'd be ok including this, if there is a pressing need. But if other permissions changes would solve this as well, I'd go for that.

lentschi:
@wvengen

> what's your use-case exactly?

Well like I described in #732 my "use case" is just following [the current docs](https://github.com/foodcoops/foodsoft/blob/9897c1056fbdc047243ea1feb8f6a57be79ea3a2/doc/SETUP_DEVELOPMENT_DOCKER.md) on how to run foodsoft with docker-compose. I then get the error as described in the issue - the cause being that my machine's local user/group ids is non-default (which would be 1000/1000).

> I'd be ok including this, if there is a pressing need.

No there's no pressing need. Apparently I'm the only dev with this issue, so I'm not angry if you just decline the PR ;-) (I just wanted to share this, in case someone else runs into it.)

@paroga 

> IMHO the correct fix would be to just change the file/directory permissions of the log directory to make it writeable. Or do I miss something?

Hm... This might fix the error, yes :thinking: (Haven't tried that yet.)
But still - Any file created from within the docker container will have a uid/gid that differs from the current host system's user which can be annoying in a lot of cases when using this setup for developing.

I just stumbled over this: https://docs.docker.com/engine/security/userns-remap/ - I haven't tried it out, but maybe this is a better fix than the one i offered as it doesn't require any changes in the repo. I'll investigate and close this PR for now.

paroga:
> But still - Any file created from within the docker container will have a uid/gid that differs from the current host system's user which can be annoying in a lot of cases when using this setup for developing.

Agree, but for me personally any `Dockerfile` with `COPY` makes it hard for development: You cannot use the "live reload" and need to rebuild the container after every change. I'd suggest to mount the source directory into the container instead, but this needs some work with the permissions too.

PR #737, opened 2020-07-19
Add BESTELLEN_EN.md

alex-quiterio:
WIP

### DIFF
--- a/doc/BESTELLEN_EN.md
+++ b/doc/BESTELLEN_EN.md
@@ -1,0 +1,96 @@
+
+# Order
+Ordering is the main part of this software and a little bit complicated.
+Here I start the attempt to convert the program logic into text and
+references to the corresponding controllers or models.
+The relevant controller is 'OrdersController'.
+
+## order "put into network
+By this we mean the selection of articles from a specific supplier for a temporary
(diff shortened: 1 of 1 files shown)


paroga:
Thanks for the PR! I'd prefer to add new texts to the [documentation](https://foodcoops.readthedocs.io/en/latest/) in the [user-docs](https://github.com/foodcoops/user-docs) repository.

wvengen:
@paroga I agree, for user/foodcoop-documentation. This document is kind of both: an intro for developers wanting to delve into the code, a very minimal specification, and reference documentation for users/foodcoops.
Somehow I think that having to keep this doc in sync with the algorithm would be good, and easier to do when it is in this repo.
That being said, I can live with both. What do you think, @paroga?

paroga:
if the target audience are developers, i agree that it should go into this repository, but i doesn't look like that to me. i'd redirect developers who need a basic overview about what foodsoft does to the user documentation. i see no value in duplicating that. i personally would favor removing the existing `BESTELLEN.md` (we should remove the remaining german texts anyway) instead

wvengen:
Good, let's move it to user-docs.
The integration spec already covers the need for making sure the algorithm keeps working correctly.

wvengen:
@alex-quiterio care to open a PR there, please?

PR #740, opened 2020-07-29
Nullify deliveries and orders depending on an invoice (#718)

paroga:


db/migrate/20181204000000_clear_invalid_invoices_from_orders.rb
```diff
@@ -0,0 +1,5 @@
+class ClearInvalidInvoicesFromOrders < ActiveRecord::Migration
+  def up
+    Order.where('invoice_id NOT IN (?)', Invoice.ids).update_all('invoice_id = NULL')
+  end
+end
```
wvengen, line 5: In migrations, one needs to define the models, because in later versions of the software the model may be renamed or otherwise not available anymore.
```ruby
class Order < ActiveRecord::Base; end
class Invoice < ActiveRecord::Base; end
```
may be enough. Or use SQL instead, which will always work with this database schema version.

db/migrate/20181204000000_clear_invalid_invoices_from_orders.rb
```diff
@@ -0,0 +1,5 @@
+class ClearInvalidInvoicesFromOrders < ActiveRecord::Migration
+  def up
+    Order.where('invoice_id NOT IN (?)', Invoice.ids).update_all('invoice_id = NULL')
```
wvengen, line 3: Small improvement:
```ruby
Order.where.not(invoice_id: Invoice.all).update_all(invoice_id: nil)
```

PR #742, opened 2020-07-29
Add FinancialLinksController#incomplete

paroga:


paroga:
not yet, you need to know the URL. there are still some things missing with the financial links, i'd like to add before making it visible to all users (maybe hide it behind settings...)

PR #746, opened 2020-07-30
Remove number_display.rb

paroga:
This is the last preperation for rails 5.0.


paroga:
sorry, but i don't get the question. do you want me to rename the functions or do you think i missed something?

wvengen:
I meant: a helper method like this (name could be improved, perhaps):
```ruby
def number_to_article_result(result)
   number_with_precision result, strip_insignificant_zeros: true
end
```

paroga:
thx for the clarification. i though already myself about a helper function, but we don't have a common place for that at the moment (the `number_to_*` functions are from rails itself) and i don't want to refactor too much in this PR. the usage of `number_to_currency` could need some cleanup too...

PR #750, opened 2020-07-31
Remove doc/bestellen.md

alex-quiterio:
As discussed in #737, we will be removing doc/bestellen from this repo and move the necessary information to [user-docs](https://github.com/foodcoops/user-docs)

https://github.com/foodcoops/foodsoft/pull/737


wvengen:
Thanks! Let's merge this when the documentation is merged into the user-docs repo.

PR #751, opened 2020-07-31
Update Ruby version to 2.6.6

paroga:


wvengen:
Let's do it!
This still includes the ActiveJob commit.

PR #752, opened 2020-08-01
Use ActiveJob instead of Resque directly [requires Rails 5.0]

paroga:
This is a newer and cleaner version of #749, which requires Rails 5.0.


app/jobs/application_job.rb
```diff
@@ -0,0 +1,10 @@
+class ApplicationJob < ActiveJob::Base
+  def serialize
+    super.merge(foodcoop: FoodsoftConfig.scope)
+  end
+
+  def deserialize(job_data)
+    FoodsoftConfig.select_multifoodcoop job_data['foodcoop']
+    super(job_data)
+  end
```
wvengen, line 9: Will this also work when running inline (without a background-job backend)?

paroga: since it runs inline there is no need to switch the database :wink:

wvengen: hah, you got me ;)

paroga:
reverted, since this PR has problems with multicoop setup

wvengen:
Ouch, thank you! I've added the label 'needs work' to reflect that :)

PR #756, opened 2020-08-07
Adopt printer plugin to use ActionCable

paroga:
This is an overview of what is needed to adopt the printer plugin to ActionCable. (Please ignore the commits related to the Rails 5 update). I integrated it with the API controllers, but we need a new concept of "userless api keys". I created a simple config option for now, since I didn't find a matching concept in doorkeeper. (Another usecase for this userless api key would be a background job which synchronizes something with foodsoft and is not related to a user). I'd like to get some early feedback before cleaning it up.


plugins/printer/app/channels/printer_channel.rb
```diff
@@ -0,0 +1,20 @@
...
+    job = PrinterJob.unfinished.find_by_id(data['id'])
+    return unless job
+    job.add_update! data['state'], data['message'] if data['state']
+    job.finish! if data['finish']
+    self.class.broadcast_unfinished
+  end
+
+  def self.broadcast_unfinished
+    ActionCable.server.broadcast(broadcasting, {
+      unfinished_jobs: PrinterJob.pending.pluck(:id)
```
wvengen, line 17: `unfinished_job_ids`?

plugins/printer/app/controllers/api/v1/printer_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Api::V1::PrinterController < Api::V1::BaseController
+  include Concerns::SendOrderPdf
+  before_action -> { require_plugin_enabled FoodsoftPrinter }
```
wvengen, line 3: Do we need to restrict this to the 'printer user'?

lib/foodsoft_config.rb
```diff
@@ -260,6 +265,7 @@ def get_default_config
         tasks_period_days: 7,
         tasks_upfront_days: 49,
         shared_supplier_article_sync_limit: 200,
+        api_key: '',
```
wvengen, line 268: This can be empty, right? The default value of `nil` also works. This section is primarily for being backward-compatible, for items that really can't be `nil`. I'd rather put sane defaults in the app_config sample.

wvengen:
I agree that we need some API keys that work without a user.
I'll have a look at the different OAuth flows to see if anything is suitable.

paroga:
> I'll have a look at the different OAuth flows to see if anything is suitable.

@wvengen did you find a nice way for the api_key?

wvengen:
In theory we could use the [client credentials grant type](https://tools.ietf.org/html/rfc6749#section-4.4) (see also [here](https://alexbilbie.com/guide-to-oauth-2-grants/#client-credentials-grant--section-44)). This is especially meant for services without the context of a user. We could add a printer authorization scope.
I didn't see this in the _Applications_ configuration tab in Foodsoft, but it looks like Doorkeeper [supports this](https://github.com/doorkeeper-gem/doorkeeper/wiki/Client-Credentials-flow) (and see a [spec](https://github.com/doorkeeper-gem/doorkeeper/blob/3f8950ec08d717f4d710f82f8947490f17e050a5/spec/requests/flows/client_credentials_spec.rb)).
This is probably a little more complex then putting a key in the configuration file, but would keep keys for external applications in a single place, which I think is a good thing.

paroga:
i use this already for the import foodsoft `FinanctialTransaction` into another foodsoft as `BankTransactions` in [the austria fork](https://github.com/foodcoopsat/foodsoft/blob/1b4c2224407c8499a2c8966f3cb90ee267d2b1c8/plugins/banks_austria/lib/foodsoft_banks_austria/bank_account_connector_foodsoft.rb#L22), but this does not work reliable.
It often throws an `OAuth2::Error` with `invalid_grant: Die bereitgestellte Autorisierung ist inkorrekt, abgelaufen, widerrufen, ist mit einem anderen Client verknüpft oder der Redirection URI stimmt nicht mit der Autorisierungs-Anfrage überein.`, but then it works again without problems. IMHO this is a big show stoppper.

> We could add a printer authorization scope.

of course

> keep keys for external applications in a single place, which I think is a good thing.

i think so too (if it would work reliable)

wvengen:
Ouch, that doesn't sound too nice about the `OAuth2::Error` (and I'm impressed you've already done something like that).
Did you find any clue with Doorkeeper?
Anyhow, this case would be slightly different, using the client credentials _grant_ (no user context), so tiny chance it might work still.

paroga:
> Did you find any clue with Doorkeeper?

Unfortunately no and i wasn't able to make it 100% reproducible, which makes the whole bug hunting very frustrating.

paroga:
#783 could explain the doorkeeper problem

RayOei:
What is the use case here?
The mentioned #783 seems to have been solved but no follow-up has been noted here. So the actual state, apart from having conflicts, is unclear.

PR #757, opened 2020-08-13
Change supplier_id of stock orders to NULL

paroga:


wvengen (approved):
Yes, that seems better. Thanks!

Issue #736, opened 2020-07-18
Enhance data integrity by adding foreign key constraints to the database


lentschi:
**Current state**
Currently there are no foreign key constraints in the db.

**Expected state**
Adding some would have the following advantages:

- Prevent unforeseen runtime errors caused by broken references in future changes
- Enhance manual crawling: Crawling through the DB is difficult due to broken references 
- Prevent bloated DB size
- Prevent data from partially not being removed, when removal was specifically requested by a user due to GDPR requests
- Tooling: Lots of DB viewers (such as DBeaver) support following foreign keys by ctrl clicking on them and other features like automated ERD generation.

**Difficulty**
Looking through the DB dump of my own foodcoop I found a lot of inconsistencies. (also see #158) Those will first have to be cleaned up by either deleting the records or setting foreign keys to `NULL`. (Which of the two depends on the way each of them is used in the application and on whether they are nullable.)

lentschi:
I had a first go at it in https://github.com/lentschi/foodsoft/commit/7551c7f12e28cfc922cdcf1b4f714527fbe77d32 but didn't create a PR yet, as there is an open matter to discuss:

Due to the inconsistent state of our foodcoop's SQL dump, I had to come up with a way to first clean the DB (see [this migration](https://github.com/lentschi/foodsoft/blob/issue_736/db/migrate/20200718074330_delete_orphaned_records_from_all_tables.rb)). But I've no idea if these SQL statements will be sufficient for _all foodcoops_ and it will be hard (near impossible) to get them right for anyone.
Still, I think this would add a lot of stability to the application, so we should strive to get them as good as it gets and at some point take risk that migrations will fail for some foodcoops. (We'll need to analzye their error logs then and provide patches)

paroga:
i support adding foreign keys. why do you `DELETE` the entities in the migration? why can't you just nullify them?

lentschi:
@paroga 

> i support adding foreign keys. why do you `DELETE` the entities in the migration? why can't you just nullify them?

I agree, it might be better to nullify *some* of them:
E.g. I'm pretty sure that `DELETE FROM order_articles WHERE article_id NOT IN (SELECT id FROM articles)` makes sense as this is just an extended join table, right. But `DELETE FROM articles WHERE article_category_id NOT IN (SELECT id FROM article_categories)` on the other hand is probably too drastic. Maybe these articles should just be assigned to some default category...?

[...]

PR #759, opened 2020-08-14
Partial fix for #736: Add foreign keys to ensure db integrity

lentschi:
Partial fix for #736: Add foreign keys to ensure db integrity

*Note*: At the time of creating this PR 464c9a3 on which it was based due to a request by @paroga still wasn't merged into the master. As long as that is the case a lot of the changes displayed have nothing to do with #736


paroga:
it's very difficult to review this, because you included already merged (ecb0846, 0768b47) and unrelated (
cebf10b) commits. you could also squash your changes into one commit, which would allow review of only a single commit. otherwise i need to do this myself on my machine.

can you change you migration files to a more rails-like syntax? (some examples are [`ClearInvalidInvoicesFromOrders`](https://github.com/foodcoops/foodsoft/blob/aa44291b69f41baf9af1b178d76104a0eee91781/db/migrate/20181204000000_clear_invalid_invoices_from_orders.rb) and [`ChangeOrdergroupDefaultInFinancialTransaction`](https://github.com/foodcoops/foodsoft/blob/aa44291b69f41baf9af1b178d76104a0eee91781/db/migrate/20181201000301_change_ordergroup_default_in_financial_transaction.rb))
please add newlines at the end of files (e.g. `db/migrate/20200718074330_delete_orphaned_records_from_all_tables.rb`)

since it's possible that users have been removed from the database in old versions (in the meantime the get marked as deleted), you need to nullify user columns.
please also check were NULL values are allowed and just nullify columns instead of removing rows (e.g. `group_orders.updated_by_user_id` allows `NULL` values)
deleting `invites` and `settings` depending on `users` seams good to me

paroga:
did you check if all rails models implement the correct relation details (`has_many :other_type, dependent: :destroy`)?

lentschi:
> it's very difficult to review this, because you included already merged (ecb0846, 0768b47) 

Merged where? I don't see these two commits (neither in foodcoops/foodsoft/master nor in paroga/foodsoft/master).
But I guess what you really wanted me to do is create a draft PR _without_ any rebase on your code, right? (I must have misread [your comment in the related issue](https://github.com/foodcoops/foodsoft/issues/736#issuecomment-671252176), sorry: "maybe you want to rebase your work on it for now")
I now created another PR (https://github.com/foodcoops/foodsoft/pull/762) replacing this one, where you can see the changes more clearly. (It's just those two migrations anyway)
Of course without the rebase on your non-merged changes, a lot of tests will still fail. (At least until https://github.com/foodcoops/foodsoft/pull/757 is merged.)

> can you change you migration files to a more rails-like syntax?

Okay done. (I used SQL syntax as it's a bit faster and I don't see any downside to it an it has been used in some other migrations too. But if you prefer it like that, I'm fine with it! :smile: )

> since it's possible that users have been removed from the database in old versions (in the meantime the get marked as deleted), you need to nullify user columns.
> please also check were NULL values are allowed and just nullify columns instead of removing rows (e.g. group_orders.updated_by_user_id allows NULL values)

Done. Though in some cases I'm note sure whether this makes sense. E.g. `Message.sender_id` may be nullified, but do we really want to keep messages around, of which we no longer know who sent them?

> did you check if all rails models implement the correct relation details (has_many :other_type, dependent: :destroy)?

No, not yet. Another reason why this is still a draft PR.

paroga:
> Merged where?

sorry, you're right. the correct hashes in the master branch are 67ab647eb0498f2b0a69ee07fba9d0c42ff486f4 and 7e3c601a251f36f7cf18200754472c940aedb178

> I used SQL syntax

I did that too, but wvengen wanted me to change in one of my PR IIRC.

> (At least until #757 is merged.)

maybe you can give it a review? (it looks nearly identical to your)

> do we really want to keep messages around, of which we no longer know who sent them?

the point is that they are accessible at the moment and i don't like the idea that it will be gone after an update (on the ohter side: 99.99% of the messages are useless anyway after a few weeks IMHO)

PR #760, opened 2020-08-14
Partial fix for #736: Set orders' supplier_id=NULL instead of 0

lentschi:
Partial fix for #736:
Set orders' supplier_id=NULL instead of 0.

*Note*: At the time of creating this PR 464c9a3 on which it was based due to a request by @paroga still wasn't merged into the master. As long as that is the case a lot of the changes displayed have nothing to do with #736


paroga:
i'm not sure what do you want to merge. why do you need to incloude my Rails 5 commits in this PR? FYI: there is already #757 for the supplier_id change

lentschi:
> i'm not sure what do you want to merge. why do you need to incloude my Rails 5 commits in this PR? FYI: there is already #757 for the supplier_id change

Didn't you ask me to create a PR rebased on 464c9a3 in https://github.com/foodcoops/foodsoft/issues/736#issuecomment-671252176 ? 
(That's what this is) Maybe I misunderstood...? :thinking: 

> FYI: there is already #757 for the supplier_id change

Ah, I hadn't seen that, guess I can close this one then. Should I close https://github.com/foodcoops/foodsoft/pull/759 too, instead rebase it on master and open another PR for that?

PR #761, opened 2020-08-14
Do not limit the number of page versions to 20

paroga:
This reverts 164653dbfc46f01b319880061dc0cdedf305115b.


wvengen (approved):
Looks like a good idea!

PR #762, opened 2020-08-16
On #736: Added cleanup migrations

lentschi:
(squashed and modified commits from issue_736)


db/migrate/20200718074330_delete_orphaned_records_from_all_tables.rb
```diff
@@ -0,0 +1,46 @@
...
+  class Message < ActiveRecord::Base; end
+  class Group < ActiveRecord::Base; end
+  class MessageRecipient < ActiveRecord::Base; end
+  class OrderArticle < ActiveRecord::Base; end
+  class OrderComment < ActiveRecord::Base; end
+  class Order < ActiveRecord::Base; end
+  class Supplier < ActiveRecord::Base; end
+  class Setting < ActiveRecord::Base; end
+
+  def up
```
paroga, line 19: we should also add an (empty) `def down`

db/migrate/20200718074330_delete_orphaned_records_from_all_tables.rb
```diff
@@ -0,0 +1,46 @@
...
+    OrderComment.where.not(order_id: Order.all).destroy_all
+    OrderComment.where.not(user_id: User.all).update_all(user_id: nil)
+    Order.where.not(supplier_id: Supplier.all).update_all(supplier_id: nil)
+    Order.where.not(updated_by_user_id: User.all).update_all(updated_by_user_id: nil)
+    Order.where.not(created_by_user_id: User.all).update_all(created_by_user_id: nil)
+
+    # We're not going to be able to add a FK constraint for settings, but still...:
+    Setting.where(thing_type: 'User').where.not(thing_id: User.all).destroy_all
+  end
+end
```
paroga, line 46: missing EOL at the end of file

lentschi:
I just fixed the unit tests by creating temporary tables for `shared_articles` and `shared_suppliers`. They aren't *real* temporary tables, but that I couldn't get to work with ActiveRecord for some reason.
I still believe it's cleaner that the current way of 'recycling' the internal `articles` and `suppliers` tables for testing though. (They have a completely different structure after all.)

The real clean solution would be to use an API for this kind of import/export (or at least some sort of shared schema via a gem), but that's outside of the scope of this PR.

RayOei:
What is the intent here as is still in `Draft` and has been for some time 😇 
Added clear link to #736 for navigation.

lentschi:
> What is the intent here as is still in Draft and has been for some time

I'm not sure, what you're asking...

- "Why is it still in Draft and not being merged?" -> Because we didn't dare to merge it back then as we didn't have the ressources to review it thoroughly and it might lead to the loss of data. (Actually I later found a few of the foreign key constraints I made here are simply wrong, but I didn't want to push this through on my own, so I hadn't invested time in correcting them.)
- "Why is it still in Draft and not Closed?" -> Because the code might still be useful. Now that we have a bigger team, we might have another go at it. (But after https://github.com/foodcoops/foodsoft/pull/1073 we'll have to rewrite a lot of it, of course.)

RayOei:
> I'm not sure, what you're asking...

From a communication perspective it is not clear what the idea of the PR is, at this point in time:
 - if it is a `draft` does it mean it is still worked on?
 - looking at the create date it is an old PR, so how does hat relate to the `draft` state?
 - what is the reasoning it is never merged?
    - is there something wrong?
    - missing? 
    - contentious?
    - other? 

Therefore: thanks for the clarification 👍 

I am not a fan - and I am probably biased by how my day jobs work - of long term dangling or (seemingly) abandoned PR's. Either they are useful and can be merged or they are not and can be closed. I understand the challenge of an OSS project but keeping the PR's a bit tidy, when possible, helps to keep track of what -_is_ still important and on the radar and what can possibly be picked up to get it over the finish line. Which can also be rewarding for the contributor. 

If they are not _potentially_ merge candidates (e.g. can be merged) and are still WIP, in some form or fashion, they should live on their branch until they get into a ready state. If there is not intent or possibility to finish a PR they become `noise`.
And if it is already known, by now, it should be rewritten, I would favor the clarity of closing _this_ PR. 
Prioritize the original issue (#736) and continue working on it when it is (re)scheduled. It is obviously still an important improvement in general.

The existing code can be pushed to the repo on its own branch. It is not lost that way and can serve as a basis or inspiration when the issue is addressed again but it is also clear _it is not there yet_. 

My 2 cent 😁

lentschi:
> if it is a draft does it mean it is still worked on?

Hard to tell - I can only say: _I'm_ not working on it _right now_ - someone else might be, but I doubt it.

And for all your other points: I think that this is something we can discuss _in general_ in https://github.com/foodcoops/foodsoft/discussions/1135 (You make good points, but if you present them here in the PR comments, no one will read them.).

As far as _this specific PR_ is concerned, I think the issue is important enough that it justifies people stumbling upon it from time to time, even though the PR's code is not being developed/reviewed right now.

But if there are other voices to close it or you yourself feel strongly about it - just go ahead and do so - no hard feelings :slightly_smiling_face:.

RayOei:
> But if there are other voices to close it or you yourself feel strongly about it - just go ahead and do so - no hard feelings 🙂.

BTW: at this point in time, I am hesitant to be too ambitious in closing stuff. Particularly if it is work from any of the main contributors and I am still lacking clear enough insights on `how the machine works`. 😁

PR #763, opened 2020-08-17
Ensure that the article_id is set for every article_price

paroga:
This addresses the TODO from dc92de771e870b65401c623fa704639ce7fe5bdc.


wvengen (approved):
Ouch, this seems like a good idea.

Issue #764, opened 2020-08-23
Add option to ignore delivered amounts in order group distribution


lentschi:
**Current behavior**

When receiving an order (using `OrdersController#receive`) and supplying less than the full amount of articles ordered, the distribution among order groups currently works as follows:
The order groups that first placed their group orders will receive (-> `GroupOrderArticle#result_computed`) full amounts. Those who ordered later will receive only what is left (if anything). So it could be described as a "first order first serve" rule.

However in our (and I think in many other) foodcoops this isn't a viable solution. (The sequence of ordering and fetching stuff is almost always different from each other.) So I suggest the following change:

**Suggested change**

Add a checkbox to the `others` config page: "Ignore received amounts when charging order groups"
When this is checked, the order groups are always assigned (and subsequently charged) the full amounts of what they ordered. 
(Amounts can still be modified at a later time through the existing interfaces or - once implemented - through https://github.com/foodcoops/foodsoft/issues/766.)

When the new checkbox remains unchecked (default), the behavior stays the same as it is now.

wvengen:
I can imagine that this works differently in some foodcoops.
The receive screen was originally added to redistribute on receiving, but I can imagine that just registering the received amount can be useful aswell.

Would the checkbox here belong in the receive screen, in the supplier details, or be a foodcoop-wide option: what do you think?

lentschi:
> Would the checkbox here belong in the receive screen, in the supplier details, or be a foodcoop-wide option: what do you think?

The way I started implementing it (see 8c2ba9efcc94b7480c), it would be a foodcoop-wide option. Are you suggesting we should allow different ways of calculation per supplier?

Maybe - since each foodcoop might have different requirements - the `calculate_result` method could be adapted to be easy to override (e.g. by defining a plugin API for hooking into calculation)...?

wvengen:
I think it would be useful to have an adaptable `calculate_result` method. I know of other foodcoops who were pondering if the tolerance would not be distributed on a first-come-first-serve-basis, but spread over all GroupOrderArticle tolerances.

Some parts could be split off from `calculate_result` into separate methods, which can be overridden in plugins. I'm not sure right now what parts to splitt off, suggestions welcome.

lentschi:
> I think it would be useful to have an adaptable calculate_result method. 

[...]

PR #765, opened 2020-08-23
#764 - Add option to ignore delivered amounts in order group distribution

lentschi:
**TODOs:**

- [x] ~Translate the new config checkbox to other languages~
- [x] ~Check Dutch translations.~

Not TODO after all :slightly_smiling_face:  - see https://github.com/foodcoops/foodsoft/issues/764#issuecomment-711067892

### DIFF
--- a/app/models/group_order_article.rb
+++ b/app/models/group_order_article.rb
@@ -130,13 +130,14 @@ class GroupOrderArticle < ApplicationRecord
 
       # Determine quantities to be ordered...
       order_quantities.each do |goaq|
-        q = [goaq.quantity, total - total_quantity].min
+        q = goaq.quantity
+        q = [q, total - total_quantity].min unless FoodsoftConfig[:ignore_received_units_on_redistribute]
         total_quantity += q
         if goaq.group_order_article_id == self.id
           logger.debug "increasing quantity by #{q}"
           quantity += q
         end
-        break if total_quantity >= total
+        break if total_quantity >= total && !FoodsoftConfig[:ignore_received_units_on_redistribute]
       end
 
       # Determine tolerance to be ordered...
--- a/app/models/group_order_article.rb
+++ b/app/models/group_order_article.rb
@@ -131,13 +131,13 @@ class GroupOrderArticle < ApplicationRecord
       # Determine quantities to be ordered...
       order_quantities.each do |goaq|
         q = goaq.quantity
-        q = [q, total - total_quantity].min unless FoodsoftConfig[:ignore_received_units_on_redistribute]
+        q = [q, total - total_quantity].min if FoodsoftConfig[:distribution_strategy] == :first_order_first_serve
         total_quantity += q
         if goaq.group_order_article_id == self.id
           logger.debug "increasing quantity by #{q}"
           quantity += q
         end
-        break if total_quantity >= total && !FoodsoftConfig[:ignore_received_units_on_redistribute]
+        break if total_quantity >= total && FoodsoftConfig[:distribution_strategy] == :first_order_first_serve
       end
 
       # Determine tolerance to be ordered...
--- a/app/models/group_order_article.rb
+++ b/app/models/group_order_article.rb
@@ -128,16 +128,18 @@ class GroupOrderArticle < ApplicationRecord
       order_quantities = GroupOrderArticleQuantity.where(group_order_article_id: order_article.group_order_article_ids).order('created_on')
       logger.debug "GroupOrderArticleQuantity records found: #{order_quantities.size}"
 
+      first_order_first_serve = (FoodsoftConfig[:distribution_strategy] == FoodsoftConfig::DistributionStrategy::FIRST_ORDER_FIRST_SERVE)
+
       # Determine quantities to be ordered...
       order_quantities.each do |goaq|
         q = goaq.quantity
-        q = [q, total - total_quantity].min if FoodsoftConfig[:distribution_strategy] == :first_order_first_serve
+        q = [q, total - total_quantity].min if first_order_first_serve
         total_quantity += q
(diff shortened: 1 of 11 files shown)


config/locales/en.yml
```diff
@@ -597,6 +597,7 @@ en:
       tasks_upfront_days: For how many days in advance you would like to schedule periodic tasks.
       tax_default: Default VAT percentage for new articles.
       tolerance_is_costly: Order as much of the member tolerance as possible (compared to only as much needed to fill the last box). Enabling this also includes the tolerance in the total price of the open member order.
+      ignore_received_units_on_redistribute: Even, if the supplier hasn't delivered all that was ordered, the order groups will be charged for the full amounts, unless manually changed.
```
wvengen, line 600: first comma can be omitted

wvengen:
I may be useful to show something on the receive screen about whether group orders are updated or not, something that doesn't catch the attention, but is there if you need to know it.

paroga:
i don't think we should merge as it is and instead change it to https://github.com/foodcoops/foodsoft/issues/764#issuecomment-690961696

wvengen:
Ah, sorry, you're right! :+1:

app/views/admin/configs/_tab_others.html.haml
```diff
@@ -1,5 +1,9 @@
 = config_input form, :use_nick, as: :boolean
 = config_input form, :tolerance_is_costly, as: :boolean
+- distribution_options = [[t('config.keys.distribution_strategy_options.first_order_first_serve'), FoodsoftConfig::DistributionStrategy::FIRST_ORDER_FIRST_SERVE],
+  [t('config.keys.distribution_strategy_options.no_automatic_distribution'), FoodsoftConfig::DistributionStrategy::NO_AUTOMATIC_DISTRIBUTION]]
+= config_input form, :distribution_strategy, as: :select, collection: distribution_options,
```
paroga, line 5: Can we use `label_method` here? Doing it like [here](https://github.com/foodcoops/foodsoft/blob/4043433539f34b880bec8826d2f29f9137412ef4/app/views/finance/balancing/_edit_transport.html.haml#L7-L8) would make the code a little bit more readable.

app/models/group_order_article.rb
```diff
@@ -128,15 +128,18 @@ def calculate_result(total = nil)
       order_quantities = GroupOrderArticleQuantity.where(group_order_article_id: order_article.group_order_article_ids).order('created_on')
       logger.debug "GroupOrderArticleQuantity records found: #{order_quantities.size}"
 
+      first_order_first_serve = (FoodsoftConfig[:distribution_strategy] == FoodsoftConfig::DistributionStrategy::FIRST_ORDER_FIRST_SERVE)
+
       # Determine quantities to be ordered...
       order_quantities.each do |goaq|
-        q = [goaq.quantity, total - total_quantity].min
+        q = goaq.quantity
```
paroga, line 135: i'm not 100% happy with this piece of code, but it should be ok for now and can be adopted when we add additional strategies later.

paroga (commented):
just two small comments, otherwise LGTM. i'll wait until we released 4.7.1 (see #777) before merging it into master

app/views/admin/configs/_tab_others.html.haml
lentschi: Sure - good point :ok_hand:

paroga:
LGTM, but could you please add some model tests for it

lentschi:
> LGTM, but could you please add some model tests for it

The only model method i changed here is `GroupOrderArticle#calculate_result` and as far as I can see it doesn't have any real unit tests at all. (Except for [this line](https://github.com/foodcoops/foodsoft/blob/ecda1c147809ab47af62d69ebb58e206fed00c13/spec/models/group_order_article_spec.rb#L11) implicitly calling it, but not really accessing its contained algorithms.)

I now *did* add [two simple tests](https://github.com/foodcoops/foodsoft/pull/765/files#diff-ab02b0b1d2d79ef69805e48807800586236f9699c8a22f7da6b059119f77b5c0R45), but they merely scratch the surface of this method.
IMO it's far too big and should be split up before going into writing too many unit tests.

spec/factories/group_order_article_quantity.rb
```diff
@@ -0,0 +1,9 @@
+require 'factory_bot'
+
+FactoryBot.define do
+
+  # requires order_article
+  factory :group_order_article_quantity do
+  end
```
wvengen, line 7: would be good to add `order_article` inside the factory if it's required

lentschi: That was a mistake in my comment - it should have read `# requires group_order_article`.
And actually I intentionally didn't add it to the new factory as it was manually set by my tests anyway. I followed the same pattern, that was used here for example: https://github.com/foodcoops/foodsoft/blob/45a8911ca656253d17ba095e698b232629db0b94/spec/factories/group_order.rb#L5 

But I just added this associations (and all ones this entails) in this PR and all the tests still pass:
https://github.com/foodcoops/foodsoft/pull/810/files
(With that you *could* use `create :group_order_article_quantity` without manually passing a `group_order_article`. I marked the PR as `Draft` as I don't really see the point of this, but if you do feel free to merge it anyway :smile: )

wvengen: Ah, great you followed existing convention! I think it would be useful to add this, great you submitted a PR ... ! :rocket:

wvengen: What I understood of FactoryBot's is that the idea is that each factory can be built by itself. If you want to give some as arguments, fine, but all minimum requirements are best to define in the factory (to avoid boilerplate in tests, unless it is needed because certain test objects need to be related).

Issue #772, opened 2020-10-10
searching in external article database fails with "A problem has occured." alert


JuliusR:
### affected versions

- master (4.7.99) 
- current release (4.7.0)

### steps to reproduce

1. setup sharedlists and import at least one external supplier
2. navigate to the supplier's article index and use "Import article" button

### expected behavior

New UI elements should show up (for searching and importing articles from the sharedlists database).

### actual behavior

The expected UI element does not show up. Instead, a native browser alert shows the following message: 
> A problem has occured.

The server log shows the following message:
> ```
> ActionView::Template::Error - undefined method `map' for #<ActionController::Parameters:0x00007f35d4590ba8>
> ```

JuliusR:
The error originates from the following line:

https://github.com/foodcoops/foodsoft/blob/78c4ebcb4b61891e13d3dbb9f07c34ad88c41f4a/app/helpers/articles_helper.rb#L19

This has probably been introduced by a rails upgrade.

PR #773, opened 2020-10-10
fix searching in external article database (#772)

JuliusR:
Can we use this simple fix for #772 ?

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -215,10 +215,10 @@ class ArticlesController < ApplicationController
   #
   def shared
     # build array of keywords, required for ransack _all suffix
-    params[:q][:name_cont_all] = params[:q][:name_cont_all].split(' ') if params[:q]
-    # Build search with meta search plugin
-    @search = @supplier.shared_supplier.shared_articles.search(params[:q])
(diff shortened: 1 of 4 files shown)


wvengen:
Thanks for finding the issue here! :100: 
Could this introduce a security issue?

JuliusR:
> Could this introduce a security issue?

This is the correct question. I believe the answer is "no" because we are using the result for the following two purposes only:

- `highlight`ing the matched name parts in the view
- passing the query to ransack (where [`to_unsafe_h` is called right away](https://github.com/activerecord-hackery/ransack/blob/c9cc20de9e0f7bab92e0579c85bed64d614d23de/lib/ransack/search.rb#L18))

I perfectly agree that safety should be obvious, so my first attempt with `to_unsafe_h` is bad practice.

JuliusR:
I updated the code to avoid `to_unsafe_h`. Could you have a look again please?

wvengen (approved):
Thank you, this cleans up the existing code too. Assuming you've tested this well, :+1:

paroga:
Please rebase your commits when pushing them to master to avoid the useless merge commits, which make the history a lot less readable. The change in the rewritten history is now 4043433539f34b880bec8826d2f29f9137412ef4.

Issue #774, opened 2020-10-10
minimal database seed silently fails


JuliusR:
### affected versions

- master (4.7.99) 

### steps to reproduce

1. setup fresh development environment (clone sources and follow `SETUP_DEVELOPMENT.md`)
2. make sure that no errors have been printed to the console
3. login to the foodsoft and fill the new-supplier form
4. submit the new-supplier form

### expected behavior

The form data are posted, and the supplier is created.

### actual behavior

The form data are posted, but the supplier is not created. Instead, the new-supplier form is rendered again. It does not show any validation errors.

JuliusR:
The problem is caused by a chain of silent failures.

The following seed fails because of a missing `financial_transaction_class`.

https://github.com/foodcoops/foodsoft/blob/78c4ebcb4b61891e13d3dbb9f07c34ad88c41f4a/db/seeds/minimal.seeds.rb#L30

As a result, `SupplierCategory.first` is `nil` such that the following line does not work properly:

https://github.com/foodcoops/foodsoft/blob/78c4ebcb4b61891e13d3dbb9f07c34ad88c41f4a/app/controllers/suppliers_controller.rb#L30

Finally, the validation fails. The form is re-rendered in order to show validation errors. However, no validation error is shown in this case (probably because the form does not have an input for the supplier category).

https://github.com/foodcoops/foodsoft/blob/78c4ebcb4b61891e13d3dbb9f07c34ad88c41f4a/app/controllers/suppliers_controller.rb#L31

PR #775, opened 2020-10-10
fix SupplierCategory seed from minimal.seeds.rb (#774)

JuliusR:
This is a minimal fix (no pun intended) for #774.

In addition, I would like to discuss the following options:

1. Should we use exclamation marks in seeds, e.g., `SupplierCategory.create!` instead of `SupplierCategory.create`? I have read somewhere that seeds should ideally be idempotent, but our seeds are not idempotent anyway (repeated seeding fails due to unique-key constraints).
1. Should we use exclamation marks for defaulting required fields such as `@supplier.supplier_category ||= SupplierCategory.first!` instead of `@supplier.supplier_category ||= SupplierCategory.first`?
1. Should we add "global" validation errors to forms?

### DIFF
--- a/db/seeds/minimal.seeds.rb
+++ b/db/seeds/minimal.seeds.rb
@@ -1,7 +1,7 @@
 # Create nessecary data to start with a fresh installation
 
 # Create working group with full rights
-administrators = Workgroup.create(
+administrators = Workgroup.create!(
     :name => "Administrators",
     :description => "System administrators.",
     :role_admin => true,
@@ -13,7 +13,7 @@ administrators = Workgroup.create(
 )
 
 # Create admin user
-User.create(
+User.create!(
     :nick => "admin",
     :first_name => "Anton",
     :last_name => "Administrator",
@@ -23,9 +23,9 @@ User.create(
 )
 
 # First entry for financial transaction types
-financial_transaction_class = FinancialTransactionClass.create(:name => "Other")
-FinancialTransactionType.create(:name => "Foodcoop", :financial_transaction_class_id => financial_transaction_class.id)
+financial_transaction_class = FinancialTransactionClass.create!(:name => "Other")
+FinancialTransactionType.create!(:name => "Foodcoop", :financial_transaction_class_id => financial_transaction_class.id)
 
 # First entry for article categories
-SupplierCategory.create(:name => "Other", :financial_transaction_class_id => financial_transaction_class.id)
-ArticleCategory.create(:name => "Other", :description => "other, misc, unknown")
+SupplierCategory.create!(:name => "Other", :financial_transaction_class_id => financial_transaction_class.id)
+ArticleCategory.create!(:name => "Other", :description => "other, misc, unknown")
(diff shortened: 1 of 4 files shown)


wvengen:
Yes, let's use `create!`, because it's important to see when something fails. Normally, this would not happen.

wvengen:
I like the idea of using `first!` too, makes sense.

Could you explain 'global' form errors a bit more?

JuliusR:
> Could you explain 'global' form errors a bit more?

Currently, validation errors are printed _right next to the corresponding input field_. In contrast, a _global form error_ can appear at the top of the form and state that "there has been an error creating the supplier". In the ideal case, it also lists all errors that cannot be displayed right next to the input field.

Note that there was no validation error at all when I found this bug. That is because there is no input field for the supplier category.

app/controllers/suppliers_controller.rb
```diff
@@ -27,7 +27,7 @@ def new
 
   def create
     @supplier = Supplier.new(supplier_params)
-    @supplier.supplier_category ||= SupplierCategory.first
+    @supplier.supplier_category ||= SupplierCategory.first!
```
JuliusR, line 30: This makes the server return a 404 for the whole page. I think this is not exactly what we aim for.

Screenshot_2020-10-11 ActiveRecord RecordNotFound at f suppliers

If we want a complete solution then we should create a default supplier category if none is found. Feels like overkill to me. What do you think?

wvengen: Hah, that makes sense. `supplier_category` [is not optional](https://github.com/foodcoops/foodsoft/blob/master/app/models/supplier.rb#L11). Since you can't delete SupplierCategories that have an associated Supplier, and the [initial migration](https://github.com/foodcoops/foodsoft/commit/d90d188dbf6b13bdfdbd92b34000b2c7e6a2b54c#diff-81ac1a9af21949a03f22bf0461a9a5a9) gives all Suppliers one, it is only possible for any last SupplierCategory to be deleted when there are no Suppliers. This is really an edge-case.

We could add a method `SupplierCategory.default` or so that returns the first or autocreated one, but I'm ok with leaving it like it is now, without the bang `!`.

JuliusR: > This is really an edge-case.

Yes, I agree.

> We could add a method SupplierCategory.default or so that returns the first or autocreated one, [...]

I started to write `SupplierCategory.find_or_create_default!`, but a supplier category in turn [needs a financial transaction class](https://github.com/foodcoops/foodsoft/blob/3d5043fd052f79b33f8cae2ea3a5cc81ee972ba5/db/seeds/small.en.seeds.rb#L10). That approach will end up in several on-demand seeds in the models. Not sure if that is a good idea.

> [...] but I'm ok with leaving it like it is now, without the bang !.

Yep, I am convinced now that the bang is wrong there.

wvengen:
Hmmm. It's a bit in-your-face, but I think it adds clarity. I'd be up for it. Maybe this is something to ask on foodsoft-discuss?

JuliusR:
> Maybe this is something to ask on foodsoft-discuss?

OK, good idea. Let us discuss that "global validation error message" separately.

Can we merge these basic fixes now?

PR #778, opened 2020-10-17
Fix setting the ArticlePrice only for one order after 3a6276a.

paroga:
Tests are missing

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -200,6 +200,10 @@ class Article < ApplicationRecord
     update_column :deleted_at, Time.now
   end
 
+  def current_article_price
+    article_prices.find_or_initialize_by(price: price, tax: tax, deposit: deposit, unit_quantity: unit_quantity)
+  end
+
   protected
 
   # Checks if the article is in use before it will deleted
--- a/spec/models/order_article_spec.rb
+++ b/spec/models/order_article_spec.rb
@@ -117,6 +117,159 @@ describe OrderArticle do
 
   end
 
+  describe 'article price' do
+    let(:admin) { create :user, groups:[create(:workgroup, role_finance: true)] }
+    let(:article) { create :article }
+    let(:order1) { create :order, article_ids: [article.id] }
+    let(:order2) { create :order, article_ids: [article.id] }
+    let(:order3) { create :order, article_ids: [article.id] }
+    let(:order4) { create :order, article_ids: [article.id] }
+    let(:oa1) { order1.order_articles.first }
+    let(:oa2) { order2.order_articles.first }
+    let(:oa3) { order3.order_articles.first }
+    let(:oa4) { order4.order_articles.first }
+    let(:go1) { create :group_order, order: order1 }
+    let(:go2) { create :group_order, order: order2 }
+    let(:go3) { create :group_order, order: order3 }
+    let(:go4) { create :group_order, order: order4 }
+    let(:goa1) { create :group_order_article, group_order: go1, order_article: oa1 }
+    let(:goa2) { create :group_order_article, group_order: go2, order_article: oa2 }
+    let(:goa3) { create :group_order_article, group_order: go3, order_article: oa3 }
+    let(:goa4) { create :group_order_article, group_order: go4, order_article: oa4 }
+
+    def price_attributes(price)
+      { price: price, unit_quantity: 1 }
+    end
+
+    it 'is correct for open orders' do
+      expect(order1.order_articles.first.price.price).to eq article.price
+      expect(order2.order_articles.first.price.price).to eq article.price
+      expect(order3.order_articles.first.price.price).to eq article.price
+
+      expect(order1.order_articles.first.price.tax).to eq article.tax
+      expect(order2.order_articles.first.price.tax).to eq article.tax
+      expect(order3.order_articles.first.price.tax).to eq article.tax
+
+      expect(order1.order_articles.first.price.deposit).to eq article.deposit
+      expect(order2.order_articles.first.price.deposit).to eq article.deposit
+      expect(order3.order_articles.first.price.deposit).to eq article.deposit
+
+      expect(order1.order_articles.first.price.unit_quantity).to eq article.unit_quantity
+      expect(order2.order_articles.first.price.unit_quantity).to eq article.unit_quantity
+      expect(order3.order_articles.first.price.unit_quantity).to eq article.unit_quantity
+    end
+
+    it 'changs only for open orders when the article is changed' do
+      old_article_price = article.price
+      new_article_price = old_article_price + 1
+      order1.finish!(admin)
+      article.update_attributes!({price: new_article_price})
+      oa1.reload
+      oa2.reload
+      expect(oa1.price.price).to eq old_article_price
+      expect(oa2.price.price).to eq new_article_price
+    end
+
+    it 'can change the name' do
+      old_article_name = article.name
+      new_article_name = "#{old_article_name} changed"
+
+      order1.finish!(admin)
+      oa1.reload
+      oa1.update_article_and_price!({}, {name: new_article_name})
+
+      article.reload
+      expect(article.name).to eq new_article_name
+    end
+
+    it 'can change the price for one order' do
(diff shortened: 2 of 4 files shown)


JuliusR:
> Tests are missing

Are you suggesting to write automated tests for this fix, or should we just test it manually before merging?

wvengen:
I suppose this would be automated tests. When touching this part of the code, I would strongly suggest to add automated tests to test all flows through the code.

paroga:
@wvengen could check my latest version?

spec/models/order_article_spec.rb
```diff
@@ -117,6 +117,147 @@ def goa_reload
 
   end
 
+  describe 'articel price' do
```
wvengen, line 120: typo ;)

spec/models/order_article_spec.rb
```diff
@@ -117,6 +117,147 @@ def goa_reload
...
+      { price: price, unit_quantity: 1 }
+    end
+
+    it 'is correct for open orders' do
+      expect(order1.order_articles.first.price.price).to eq article.price
+      expect(order2.order_articles.first.price.price).to eq article.price
+      expect(order3.order_articles.first.price.price).to eq article.price
+    end
+
+    it 'changs only for open orders when the article is changed' do
```
wvengen, line 150: typo

spec/models/order_article_spec.rb
```diff
@@ -117,6 +117,147 @@ def goa_reload
...
+      old_article_price = article.price
+      new_article_price = old_article_price + 1
+      order1.finish!(admin)
+      article.update_attribute(:price, new_article_price)
+      oa1.reload
+      oa2.reload
+      expect(oa1.price.price).to eq old_article_price
+      expect(oa2.price.price).to eq new_article_price
```
wvengen, line 158: I wonder if it would be good to test `article_prices` behaviour too, e.g. `current_article_price` object equality.

spec/models/order_article_spec.rb
```diff
@@ -117,6 +117,147 @@ def goa_reload
...
+      expect(order1.order_articles.first.price.price).to eq article.price
+      expect(order2.order_articles.first.price.price).to eq article.price
+      expect(order3.order_articles.first.price.price).to eq article.price
+    end
+
+    it 'changs only for open orders when the article is changed' do
+      old_article_price = article.price
+      new_article_price = old_article_price + 1
+      order1.finish!(admin)
+      article.update_attribute(:price, new_article_price)
```
wvengen, line 154: `update_attributes!` to catch any unexpected conditions?

spec/models/order_article_spec.rb
```diff
@@ -117,6 +117,147 @@ def goa_reload
...
+    let(:goa4) { create :group_order_article, group_order: go4, order_article: oa4 }
+
+    def price_attributes(price)
+      { price: price, unit_quantity: 1 }
+    end
+
+    it 'is correct for open orders' do
+      expect(order1.order_articles.first.price.price).to eq article.price
+      expect(order2.order_articles.first.price.price).to eq article.price
+      expect(order3.order_articles.first.price.price).to eq article.price
```
wvengen, line 147: In the long term we'd better test unit_quantity, tax and deposit too.

spec/models/order_article_spec.rb
```diff
@@ -117,6 +117,147 @@ def goa_reload
...
+
+      order3.finish!(admin)
+      oa1.reload
+      oa1.update_article_and_price!({}, {}, price_attributes(old_article_price))
+
+      article.reload
+      oa1.reload
+      oa2.reload
+      oa3.reload
+      oa4.reload
```
wvengen, line 199: note that technically `oa4` (and `order4`) is created in this line (first reference) - it's ok to keep `reload` for consistency

spec/models/order_article_spec.rb
```diff
@@ -117,6 +117,147 @@ def goa_reload
...
+      oa2.reload
+      oa3.reload
+      expect(article.price).to eq old_article_price
+      expect(oa1.price.price).to eq new_article_price
+      expect(oa2.price.price).to eq old_article_price
+      expect(oa3.price.price).to eq old_article_price
+
+      order3.finish!(admin)
+      oa1.reload
+      oa1.update_article_and_price!({}, {}, price_attributes(old_article_price))
```
wvengen, line 193: What about using `new_article_price_2` to be clearer what comes from where in the test?
Or do you explicitely want to test reverting (and returning the already existing `article_price` object)?

wvengen (commented):
Great to have a spec for this! It required some reading time, but I think it's all good :)
Some questions.

spec/models/order_article_spec.rb
paroga: i don't know what to test there. i don't care what happens exactly with `article_prices`

spec/models/order_article_spec.rb
paroga: i want to test a change to the old article price, so `old_article_price` is what i want to test

spec/models/order_article_spec.rb
wvengen: Well, in my mind the `article_prices` make up a history of the article prices (perhaps for future statistics purposes). In that sense it could be good to test that the actual `article_price` object is carried over to the `order_article` when the order closes, and that a new `article_price` object is created when the article's price is updated.
I'm ok with leaving that aside right now, it's already an improvement to have this.

app/models/article.rb
```diff
@@ -200,6 +200,10 @@ def mark_as_deleted
     update_column :deleted_at, Time.now
   end
 
+  def current_article_price
+    article_prices.find_or_initialize_by(price: price, tax: tax, deposit: deposit, unit_quantity: unit_quantity)
+  end
+
```
wvengen, line 206: Does this also get the _latest_ `ArticlePrice` with these values? Since the `article_prices` are as much as a price history of the article (as far as I understand it), it's good to point to the right object so that an `order_article` gets the correct record on order close.

paroga:
`article_prices` are no usable history, since there are no timestamps or anything like that. when we want to see the price during an order, we need to look at the order date.

wvengen:
> since there are no timestamps or anything like that

What about `created_at`? A record is created in a callback when the price changes.

paroga:
yes, but the timestamp reflects no "real world event" and is therefore IMHO useless and isn't used in the UI anywhere. it is also only created when a order finished, so just changing a price of an article would not create a new row. i would be in favor of removing it completely instead

wvengen:
It's true that it is no "real world event".
Why do we have the list of article prices then?
In my view this is a price history of the article (no so much the order_article).

paroga:
> Why do we have the list of article prices then?

To avoid duplicating all values all the time

> In my view this is a price history of the article 

But it's not, since not every price change results in a new row in the `ArticlePrice` table

wvengen:
> > Why do we have the list of article prices then?
> 
> To avoid duplicating all values all the time

Ah, sure, I had that wrong in my memory. For Article the price is are both in the object itself as in the ArticlePrice, but OrderArticle points to the Article or ArticlePrice.

> > In my view this is a price history of the article
>
> But it's not, since not every price change results in a new row in the ArticlePrice table

Every price change on Article [results](https://github.com/foodcoops/foodsoft/blob/master/app/models/article.rb#L68) in a new row. An OrderArticle's price is only updated using [`OrderArticle#update_article_and_price!`](https://github.com/foodcoops/foodsoft/blob/master/app/models/order_article.rb#L150), which [creates](https://github.com/foodcoops/foodsoft/blob/master/app/models/order_article.rb#L169) an ArticlePrice. So I think all these changes are recorded, actually.

yksflip:
@paroga is this still relevant or is there still a bug waiting for this fix?
I guess this would be obsolete with the upcoming changes from flo in the future?

RayOei:
> I guess this would be obsolete with the upcoming changes from @lentschi  in the future?

I would think so too... 
Added tags to check.

lentschi:
You're right, @RayOei, @yksflip - closing it. :+1:

PR #783, opened 2020-11-09
Run select_foodcoop as the first before_action

paroga:
When e.g. authenticate is run as before_action, it could use the wrong
database in a multicoop installation, if select_foodcoop is not run first.


wvengen (approved):
:100:

Issue #779, opened 2020-10-17
Received state for orders


lentschi:
In order to facilitate "self-service" features such as https://github.com/foodcoops/foodsoft/issues/766 it would be nice to have a distinct "received" `state` in `Order`.

**Current status**
Currently it's hard to determine what (group) orders have been received - s. https://github.com/foodcoops/foodsoft/issues/766#issuecomment-710770954.

The order state flow is `open` -> `finished` -> `closed`.

**Planned adaptation**
The order state flow should become `open` -> `finished` -> `received` -> `closed`. All current queries checking for `Order.state = 'finished'` would then instead need to check for `Order.state IN ('finished', 'received')` (Though that may be subject to change in the future: For foodcoops in which "self service" is enabled - again see https://github.com/foodcoops/foodsoft/issues/766 - balancing may only be done for 'received' orders.)

wvengen:
Sounds good!
Not all foodcoops use the receive screen, though, so it should be possible to skip that step (`finished` -> `closed`).

lentschi:
Well this is exactly what I _intended_ to implement. :blush:  The `received` state is never set unless the users post the receive page.

You *can* still go from `finished` to `closed` in a single step. (Happens when settling the respective order for example - there is no check that the order must have been received first - see my original, self set goal: "All current queries checking for `Order.state = 'finished'` would then instead need to check for `Order.state IN ('finished', 'received')`")

Or am I missing some other foodsoft feature where this might be a problem? :thinking:

wvengen:
Not that I can see. I'd be for implementing this.

It would be good to document this (e.g. in the model), including which state transitions can happen (to be sure other people realise what can be expected), and be sure to check all occurences of `finished`, also in plugins. And it would be good to mention it in the Changelog on the next release (for people with their own plugins).

lentschi:
> It would be good to document this (e.g. in the model), including which state transitions can happen (to be sure other people realise what can be expected)

Yes, you're right! I've added it [here in my PR](https://github.com/foodcoops/foodsoft/pull/789/files#diff-26bbec4be8ce33f60fa824e2bd29e6902e57541818578634bea06bce180b620eR39).
Is that place fine?

> ... and be sure to check all occurences of finished, also in plugins.

I did. Scopes such as `Order.finished` already include the `received` state, so there's not so much that needed changing.
When going through the plugins, I found one deactivated plugin' controller where it may need to be changed:
[...]

Issue #766, opened 2020-08-23
Add a page allowing to modify the received article amounts for "normal" users


lentschi:
**Current behavior**

Modifying the received article amounts (`GroupOrderArticles#result`) is currently only possible by viewing the `OrdersController#show` page (and then updating via the `GroupOrderArticlesController#update` ajax call).

However this shows group orders of **all** order groups and is thus correctly reserved to users with `role_orders`.

**Suggested addition**

We'd like to have a new page that allows altering the received amounts in very much the same way, but with the following changes:

- Available to all users (not reserved to users with `role_orders`)
- Only viewing/altering the amounts of the group order associated to the current user (through their order group) should be possible (-> Show only the "Sorted in groups" part, and that just for the single relevant group order).
- If less or more than the total amount has been received (`OrdersController#receive`), then this difference should be highlighted in a new column (displaying total delivered and ordered amounts).

This new page should replace the current `GroupOrdersController#show` page once an order has been received.

wvengen:
I like this 'self-service' idea, though it might add some complexity.
I agree the `GroupOrdersController` would be a sensible place (the member ordering screen), once an order has been received (or closed, in the past not all foodcoops used the receive screen, not sure how that is now - but received sounds logical).
This would need another configuration option that needs a good explanation.

I wonder if this belongs in a plugin or not.

lentschi:
> I wonder if this belongs in a plugin or not.

Yes, I agree: We / I could try implementing it as a plugin. (Not every foodcoop will need this feature.)

lentschi:
Okay, I'm now done with a first working version of this.

Plugin version: https://github.com/foodcoops/foodsoft/compare/master...lentschi:issue_766_plugin
Old version (directly integrated into the core): https://github.com/foodcoops/foodsoft/compare/master...lentschi:issue_766 (<- This won't be used, as we prefer a plugin as has been discussed above. I'm just referencing it here as it makes it easier to look through the changes and discuss some things)

I've created a small video to demonstrate what it would do: (Easier than trying it all out manually :wink:):
[issue766.ogv.zip](https://github.com/foodcoops/foodsoft/files/5361885/issue766.ogv.zip)

However I still did not create a PR as there are some things left clarify/to do:

**Clarify:**
[...]

PR #789, opened 2020-11-21
#779 Received state for orders

lentschi:


wvengen (approved):

wvengen:
Oops, one thing I forgot, I realised just after merging: we probably need to migrate received orders to the state!
Would you mind adding that still, @lentschi?

paroga:
can we PLEASE require tests for such fundamental changes. testing the core model should be no problem with the current testing infrastructure. this change nearly screams that i might cause a regression somewhere in the future

wvengen:
Hmm I can see this causes tension if a fundamental change happens that not all developers are aware of. Perhaps such a change would better been have discussed on the mailing-list.

Regarding testing, do you mean the affected scopes and the `Order#finished?` and `Order#received?` methods? Controller or integration tests for methods that could be affected?

paroga:
i don't see a problem with the change itself. it's fine and i support it. i just want to encourage us to write tests.
it's difficult to write test for ui/controller-stuff, but plain model changes should be easy to test. so i see no reason to merge model changes (`app/models/*.rb`) without (at least a few) tests. especially when it's such a fundamental entity like `Order`
i also think that the `Order.transaction do` part (of ` app/controllers/orders_controller.rb`) should be moved into the model,  to allow proper testing

lentschi:
> i just want to encourage us to write tests.

Sure, I can try and add some tests for the order model :blush: - I've started here: https://github.com/foodcoops/foodsoft/pull/809/files

> i also think that the Order.transaction do part (of app/controllers/orders_controller.rb) should be moved into the model, to allow proper testing

That is a bit tricky though: This part calls `OrdersController#update_order_amounts` which contains a lot of controller logic that first would need to be extracted. (Which would in turn mean a log of refactoring, but guess I can give it a try! :sweat_smile: )

Before I take a go at this - some more questions:
I just stumbled across this source comment: ["This was once wrapped in a transaction, but caused MySQL lock timeout exceeded errors. It's ok to do this article-by-article anway."](https://github.com/foodcoops/foodsoft/blob/e3c720589d7544d96b084bad269f50657edca02a/app/controllers/orders_controller.rb#L174)
Is it really? For me as a user, it would be worse, if only part of my order gets updated and part of it wouldn't.
Also, the default MySQL/MariaDB value for `innodb_lock_wait_timeout` seems to be 50 seconds. If an update query really ever takes that long, I think we should refactor the whole save process instead of removing the transaction.

@wvengen Could you maybe describe how to reproduce the error fixed by your commit d906a7342fb0f4261735eff62bd8302a19a58609? (If it is still reproducible, I may have caused a regression with my new `Order.transaction do` :fearful: I don't see anything that could use up that much time - even for huge orders - and I've never run into this with test data from my foodcoop. Could it be, that this just happened when some server was on overload?)

paroga:
> This part calls OrdersController#update_order_amounts

sorry, my bad. though that `update_order_amounts` would be already in the model. if you like to refactor the code it would be SUPER nice, but I don't see it as your responsibility. testing the scope as in #809 is all i can demand from you here.

> If an update query really ever takes that long, I think we should refactor the whole save process instead of removing the transaction.

the main problem is, that the article_prices are calculated in ruby and not in the database and i expect some missing eager loading. changing that is a bigger refactoring, which should be planed wisely and target also some limitations also.

lentschi:
> Oops, one thing I forgot, I realised just after merging: we probably need to migrate received orders to the state!
> Would you mind adding that still, @lentschi?

Yes, I added it (and @paroga has already merged it).
The reason I hadn't already added it before was that I thought the `received` state doesn't make any difference in the current master and for foodcoops switching to my potential new self service function (#766), manual db steps will be required anyway. (Switching to the new workflow with running orders would be difficult, so I think a clean cut when all orders are closed needs to be done there.)
But then I found that it is already useful in the `master`: https://github.com/foodcoops/foodsoft/pull/808/files
... so adding them was a good idea after all :slightly_smiling_face:

PR #791, opened 2020-12-20
Extend api

Nalaxon:
Based on a discussion within foodcoops.at forum, we want to extend the api to build a small (hopefully growing) mobile app.

Here I want to develop the needed api for that.

https://forum.foodcoops.at/t/fehlende-apis-fuer-mobiles-kontoblatt/4283

Please give me your feedback.

### DIFF
--- a/doc/swagger.v1.yml
+++ b/doc/swagger.v1.yml
@@ -324,37 +324,18 @@ definitions:
         type: 'string'
         default: null
         description: name of user who entered the transaction (may be <tt>null</tt> or empty string for deleted users or system users)
-      order_id:
+      ordergroup_id:
         type: 'integer'
         default: null
         description: id of order group, user is assigned to
-      credit:
-        type: number
+      available_funds:
+        type: 'number'
         default: 0
         description: actual available credit of user/ordergroup
-      balance:
-        type: 'integer'
+      account_balance:
+        type: 'number'
         default: 0
         description: credit + open claim
-      classes:
-        type: array
-        default: null
-        description: array of transaction classes and their credit, null if no transaction classes are defined
-        items:
-          $ref: '#/definitions/FinancialClass'
-          
-  FinancialClass:
-    type: object
-    properties:
-      id:
-        type: integer
-        description: id of transaction class
-      name:
-        type: string
-        description: name of transaction class
-      credit:
-        type: number
-        description: actual credit of transaction class
 
   Navigation:
     type: array
--- a/doc/swagger.v1.yml
+++ b/doc/swagger.v1.yml
@@ -324,7 +324,7 @@ definitions:
         type: 'string'
         default: null
         description: name of user who entered the transaction (may be <tt>null</tt> or empty string for deleted users or system users)
-      oder_id:
+      order_id:
         type: 'integer'
         default: null
         description: id of order group, user is assigned to
--- a/doc/swagger.v1.yml
+++ b/doc/swagger.v1.yml
@@ -320,10 +320,6 @@ definitions:
         type: 'integer'
         default: null
         description: id of user who entered the transaction (may be <tt>null</tt> for deleted users or 0 for a system user)
-      user_name:
-        type: 'string'
-        default: null
-        description: name of user who entered the transaction (may be <tt>null</tt> or empty string for deleted users or system users)
       ordergroup_id:
         type: 'integer'
         default: null
--- a/doc/swagger.v1.yml
+++ b/doc/swagger.v1.yml
@@ -269,10 +269,12 @@ definitions:
       id:
         type: integer
       user_id:
-        type: ['integer', 'null']
+        type: 'integer'
+        default: null
         description: id of user who entered the transaction (may be <tt>null</tt> for deleted users or 0 for a system user)
       user_name:
-        type: ['string', 'null']
+        type: 'string'
+        default: null
         description: name of user who entered the transaction (may be <tt>null</tt> or empty string for deleted users or system users)
       amount:
         type: number
--- a/doc/swagger.v1.yml
+++ b/doc/swagger.v1.yml
@@ -57,7 +57,30 @@ paths:
             $ref: '#/definitions/Error403'
       security:
         - foodsoft_auth: ['user:read', 'user:write']
-
+  /user/financial_summary:
+    get:
+      summary: financial summary of the member's ordergroup
+      tags:
+        - 1. User
+      responses:
+        200:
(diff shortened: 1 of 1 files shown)


doc/swagger.v1.yml
```diff
@@ -285,6 +310,51 @@ definitions:
...
+        description: id of order group, user is assigned to
+      credit:
+        type: number
+        default: 0
+        description: actual available credit of user/ordergroup
+      balance:
+        type: 'integer'
+        default: 0
+        description: credit + open claim
+      classes:
```
wvengen, line 339: let's use the full name, I think that's clearer, so `financial_transaction_classes` (if I understand correctly)

doc/swagger.v1.yml
```diff
@@ -285,6 +310,51 @@ definitions:
...
+        type: array
+        default: null
+        description: array of transaction classes and their credit, null if no transaction classes are defined
+        items:
+          $ref: '#/definitions/FinancialClass'
+          
+  FinancialClass:
```
wvengen, line 346: `FinancialTransactionClass`

doc/swagger.v1.yml
```diff
@@ -285,6 +310,51 @@ definitions:
...
+        description: name of user who entered the transaction (may be <tt>null</tt> or empty string for deleted users or system users)
+      oder_id:
+        type: 'integer'
+        default: null
+        description: id of order group, user is assigned to
+      credit:
+        type: number
+        default: 0
+        description: actual available credit of user/ordergroup
+      balance:
```
wvengen, line 335: let's use the same terms as in the code, which I assume might be `available_funds` (not sure)

doc/swagger.v1.yml
```diff
@@ -285,6 +310,51 @@ definitions:
...
+        default: null
+        description: name of user who entered the transaction (may be <tt>null</tt> or empty string for deleted users or system users)
+      oder_id:
+        type: 'integer'
+        default: null
+        description: id of order group, user is assigned to
+      credit:
+        type: number
+        default: 0
+        description: actual available credit of user/ordergroup
```
wvengen, line 334: Is this `account_balance` in the Ordergroup model?

doc/swagger.v1.yml
```diff
@@ -285,6 +310,51 @@ definitions:
...
+      balance:
+        type: 'integer'
+        default: 0
+        description: credit + open claim
+      classes:
+        type: array
+        default: null
+        description: array of transaction classes and their credit, null if no transaction classes are defined
```
wvengen, line 342: In my memory there always is one transaction class, but it's good to handle the case when there isn't.
`null` or empty array?

doc/swagger.v1.yml
```diff
@@ -285,6 +310,51 @@ definitions:
...
+    properties:
+      id:
+        type: integer
+        description: id of transaction class
+      name:
+        type: string
+        description: name of transaction class
+      credit:
+        type: number
+        description: actual credit of transaction class
```
wvengen, line 357: Ok, I'm not intimately familiar with how financial transaction classes function, but part of me expects the same fields as above: so `account_balance` as well as `available_funds`. But that doesn't really make sense, since you only have `available_funds` for one financial transaction class, from my limited understanding.

Sorry for the confusion, we somehow need to get this right. Do you have more background in financial transaction classes to explain it a bit?

wvengen (commented):
Great to see work on expanding the API. I've added some comments.

Also: reading this, part of me wishes that we had used `/ordergroup/financial_transactions` instead of `/user/financial_transactions`. But that aside.

doc/swagger.v1.yml
```diff
@@ -285,6 +310,51 @@ definitions:
         format: date-time
         description: when the transaction was entered
     required: ['id', 'user_id', 'user_name', 'amount', 'note', 'created_at']
+    
+  FinancialSummary:
+    type: object
+    properties:
+      id:
+        type: integer
```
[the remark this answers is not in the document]
Nalaxon: It was meant to be an id of this json chunk. But I see, it is really not necessary, I guess there is no need/possibility of sorting or such in the frontend for the summary... expect, you request summaries of multiple users/ordergroups?

We could depricate `/user/financial_transactions`, make `/ordergroup/financial_transactions` as a copy.
Finally remove  `/user/financial_transactions` in v2?

If you want to go this way, I would change `user_id` to `ordergroup_id`

doc/swagger.v1.yml
Nalaxon: As you prefer.

Lets use empty then, to not use a non existing array?

doc/swagger.v1.yml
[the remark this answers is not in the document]
wvengen: An `ordergroup_id` would make sense, so a front-end could sort on this.

I'd be somewhat in favour of the `/ordergroup` namespace, yes, and deprecate `/user/financial_transactions`.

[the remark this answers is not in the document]
wvengen: When requesting multiple `FinancialSummaries`, we could always add a property when necessary (removing one is more complicated, as it breaks backward compatibility). That would be in a different endpoint (e.g. `/ordergroups/:id/financial_summary` or `/ordergroups/financial_summaries`).

doc/swagger.v1.yml
wvengen: As a front-end developer, I would prefer an empty array - less `null` checking.
In a way, the array does exist, because there is a list of financial transaction classes, which is empty.

But I'll leave it to you, as a user of the API, to see what is most logical.

doc/swagger.v1.yml
```diff
@@ -285,6 +310,51 @@ definitions:
...
+      user_id:
+        type: 'integer'
+        default: null
+        description: id of user who entered the transaction (may be <tt>null</tt> for deleted users or 0 for a system user)
+      user_name:
+        type: 'string'
+        default: null
+        description: name of user who entered the transaction (may be <tt>null</tt> or empty string for deleted users or system users)
```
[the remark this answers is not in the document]
Nalaxon: `order_id ` should be `ordergroup_id`.
deleted `user_name`

doc/swagger.v1.yml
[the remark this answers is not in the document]
Nalaxon: Thanks, I will do it in the weekend or next week

Nalaxon:
@paroga alredy implmeneted the API :)
So I close this one...

PR #794, opened 2020-12-29
WIP: CI testing only

JuliusR:
This is a dummy PR to test our CI setup.


JuliusR:
I believe I understand the docker-hub behavior now. There are two types of build processes: Some are test builds (which are NOT pushed to the registry), and some are automated builds (which are pushed to the registry using a configurable docker tag).

PR #795, opened 2020-12-29
Introduce Payments

chsigi:
Payments allow a FoodCoop to track various payments, that an
ordergroup owes. This is intended to make tracking yearly membership
fees easier (for FoodCoop Allmunde :-))

I did not regenerate db/schema.rb.

Excuse the weird Ruby/ActiveRecord, these are my first steps in the ruby world.


wvengen:
Hi, thanks for your pull request, and great you've ventured into Ruby on Rails!

First a question about the idea of this change: in what way are payments different than financial transactions?
And would perhaps financial transaction classes be a solution for the same thing (which are like separate member accounts for different purposes)?

RayOei:
This issue hasn't seen any follow-up and clarification. I would suggest to close it.

PR #800, opened 2021-01-10
Fixes #799

### LINKED ISSUE #799: Permission bug in "pickup days" page
**Current behavior**

Assuming there is a user that has no other role than the `pickup_role` (only one work group assignment with only that one role).

When I'm using the "pickup days" page (`PickupsController#index` -> `OrdersController#receive`) with that user they are shown the "Add article" dropdown and the edit buttons for each table row.
However neither of those two controls do anything (We get an HTTP code 302 for the XHR calls that are being made, but those errors are not displayed to the user.)
Also when posting the receive page, the user is redirected to the `OrdersController#show` page, where many buttons don't have permission checks and lead to errors.

https://user-images.githubusercontent.com/783944/104121911-234b0c80-5342-11eb-8471-46e15562f32d.mp4

**Expected behavior**

Users with the `pickup_role` may successfully add order articles on the `OrdersController#receive` page, but aren't shown the "edit" buttons (unless they have the `orders` or `finance` role too).
When posting the receive page, users without the `orders` or `finance` role should be redirected back to the "pickup days" page.

lentschi:
Fixes by not displaying the edit buttons, if the user lacks the sufficient permissions, and allowing to add order articles for `finished` orders, if the user has at least the `pickup` role.

Also makes redirect after `receive` to be sensitive to user's permissions to avoid showing pages, on which the user lacks privileges.

### DIFF
--- a/app/controllers/order_articles_controller.rb
+++ b/app/controllers/order_articles_controller.rb
@@ -1,16 +1,15 @@
 class OrderArticlesController < ApplicationController
-
-  before_action :authenticate_finance_or_orders
+  before_action :fetch_order, except: :destroy
+  before_action :authenticate_finance_or_invoices, except: [:new, :create]
+  before_action :authenticate_finance_orders_or_pickup, except: [:edit, :update, :destroy]
 
   layout false  # We only use this controller to serve js snippets, no need for layout rendering
 
   def new
-    @order = Order.find(params[:order_id])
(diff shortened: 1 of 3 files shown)


wvengen (approved):
Thank you! Also for DRYing a little as a bonus.
I'll merge soon.

Issue #802, opened 2021-02-01
"remember me" option for signing in


carchrae:
one co-op member recently asked if i could add a 'remember me' option on the sign in.  i was confused, since the site always remembers my login.  however, after looking at the cookie (expiry is "Session") and then some head scratching about why i never had to login, i discovered that chrome makes a session cookie last forever if you have it set to "continue where i left off"  
 (  https://stackoverflow.com/questions/10617954/chrome-doesnt-delete-session-cookies  )

since there is no expiry set on the foodsoft cookie, it has an expiry of  "session" by default.

if it is simple, i suggest we add a checkbox ("remember me")  to the login form that uses the "permanent" cookie https://api.rubyonrails.org/v4.2.11/classes/ActionDispatch/Cookies.html 

if that is not straightforward to implement, i will simply change the cookie configuration to set an explicit max age.  given the default behaviour of chrome (and apparently other browsers that implement "continue where i left off") i think a better default would be to actually set the expiry to permanent so it is consistent across browser setups.

i will implement such a feature/pr unless someone beats me to it.

wvengen:
Hmmm, good one. I think this would be a very good idea.

PR #803, opened 2021-02-01
set cookie expiration

carchrae:
a quick fix for #802

caveat - i have not tested this on the master branch (rails 5) but it works on my fork which uses rails 4.2

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -18,7 +18,7 @@ module ActionDispatch
   end
 end
 
-Rails.application.config.session_store :slug_cookie_store, key: '_foodsoft_session'
+Rails.application.config.session_store :slug_cookie_store, key: '_foodsoft_session', expire_after: 1.year
 
 # Use the database for sessions instead of the cookie-based default,
 # which shouldn't be used to store highly confidential information


wvengen:
Thanks for the PR! While this would be better with 'continue where I left off', doesn't this make it a non-session-cookie, hence keeping the cookie when the browser window is closed with standards-like behaviour?

carchrae:
a session cookie is simply one with no expiry set.  (and that behaviour can be different as mentioned in the issue).  this pr just sets an expiry of one year.    

it is contentious because maybe some user will not want that behaviour (eg, they want to have to log in), but frankly i don't see many strong arguments for that as a default.  this change will mean you have to click log out to log out.

wvengen:
Ok, fair enough, thanks for explaining.
If anyone has an opinion on this, please join in, otherwise I'd be favoured to merging this after some time.

paroga:
i'm ok with adding a "remember me" option, but i don't think that a session cookie should have an expire date, regardless what browsers do. i don't like this to get merged

RayOei:
I am not sure how big of a user issue this is but it may need some more scrutiny.
Unfortunately the mentioned dismissal doesn't contain any explanation. 
A quick look indicates that the cookie implementation could be improved? See [security](https://guides.rubyonrails.org/security.html). As we have payment capabilities in FS, with iDEAL for one, I feel this needs a bit more attention.

PR #804, opened 2021-02-05
Migrations for new `received` order state (s. #779)

lentschi:
Added migrations for the new `received` order state:
All orders that have been finished associated to order articles that have `units_received` other than `nil` should have the `received` state.

**Note:** This is the first migration since the upgrade to Rails 5 -> The format of `db/schema.rb` has changed. I've tried this locally with the `docker-compose` default setup (mariadb) and it worked fine.
However I noticed that `db:migrate` produced lines like these:
`create_table "article_categories", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|`

That is, of course, a problem with non-mysql-like DBs (should be [fixed in Rails 6.1](https://github.com/rails/rails/pull/39365)) , so I simple manually stripped the `options` part and it now seems to work fine in mysql-like DBs *and* sqlite.

### DIFF
--- a/db/migrate/20210205090257_introduce_received_state_in_orders.rb
+++ b/db/migrate/20210205090257_introduce_received_state_in_orders.rb
@@ -1,0 +1,11 @@
+class IntroduceReceivedStateInOrders < ActiveRecord::Migration[5.2]
+  def up
+    Order.where(state: 'finished').each do |order|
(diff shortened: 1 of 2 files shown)


paroga:
can you please make a separate PR for the schema migration to the new rails version, to keep the history more readable

PR #806, opened 2021-02-05
Closes #805 - updated `db/schema.rb` to Rails 5.2 dump

### LINKED ISSUE #805: Upgrade schema to Rails 5.2
Rails has been upgraded to 5.2 in 899a83bec424864e6bbe4ffb38ee1d210d013912, but `db/schema.rb` remained in its old format.
-> It should be updated to its new format (to [keep the git history clean for future migrations](https://github.com/foodcoops/foodsoft/pull/804#issuecomment-773933039)).

lentschi:
The format of `db/schema.rb` has changed in Rails 5 and we should upgrade ours to keep the git history clean for future migrations.

**Note:** I noticed that `db:migrate` produced lines like these:
`create_table "article_categories", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|`

That is, of course, a problem with non-mysql-like DBs (should be [fixed in Rails 6.1](https://github.com/rails/rails/pull/39365)) , so I simple manually stripped the `options` part and it now seems to work fine in mysql-like DBs *and* sqlite.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,4 +1,3 @@
-# encoding: UTF-8
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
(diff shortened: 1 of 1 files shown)


paroga:
BTW: i personally do not see a big benefit in creating an issue in addition to a PR. i personally see it as a possible problem since it, has the potential to split a discussion about a change into multiple parts.

lentschi:
> BTW: i personally do not see a big benefit in creating an issue in addition to a PR. i personally see it as a possible problem since it, has the potential to split a discussion about a change into multiple parts.

I'm not sure, what you're getting at there :thinking: Discussions on the issue should be about what to change and how.
Discussions in the PRs should be only about technicalities such as keeping the git history clean.

paroga:
> Discussions in the PRs should be only about technicalities such as keeping the git history clean.

agree, but just opening an issue parallel to the PR adds IMHO no value. anyway I don't mind if you do, just wanted to let you know that you don't need to follow that organizational rule for such simple non-controversial PRs

PR #809, opened 2021-02-05
Rspec tests for #779

lentschi:


spec/models/order_spec.rb
```diff
@@ -17,6 +17,31 @@
...
+    it 'should retrieve open orders in the "open" scope' do
+      expect(Order.open.count).to eq(1)
+      expect(Order.open.first.id).to eq(open_order.id)
+    end
+
+    it 'should retrieve finished, received and closed orders in the "finished" scope' do
+      expect(Order.finished.count).to eq(3)
+      expect(Order.finished.first.id).to eq(finished_order.id)
+      expect(Order.finished.second.id).to eq(received_order.id)
+      expect(Order.finished.third.id).to eq(closed_order.id)
```
paroga, line 35: depending on the order could cause a flacky test. checking if the item is included should be a better way to test that

paroga (changes requested):
please add some additional test for `Order#finished?` and `Order#received?`

spec/models/order_spec.rb
```diff
@@ -40,6 +40,34 @@
       expect(Order.finished_not_closed.where(id: finished_order.id)).to exist
       expect(Order.finished_not_closed.where(id: received_order.id)).to exist
     end
+
+    it 'should return valid boolean states for open orders' do
+      expect(open_order.open?).to be(true)
```
paroga, line 45: i personally like the `.to be true` style (without brackets) more, but i'm ok with merging it as is

lentschi: There's a lot of discussion on this topic (e.g. https://stackoverflow.com/questions/340624/do-you-leave-parentheses-in-or-out-in-ruby) I personally prefer adding brackets as soon as there's a wrapped function call.

But discussing such code style issues in PRs doesn't seem like the right place -> See my earlier question here: https://github.com/foodcoops/foodsoft/pull/762#discussion_r475180328

We could then see if we can agree on a setting to https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MethodCallWithArgsParentheses :wink:

paroga: coding style has a lot of potential for discussions. i personally have no strong opinion on it, just wanted to point out that the [existing code](https://github.com/foodcoops/foodsoft/blob/e3c720589d7544d96b084bad269f50657edca02a/spec/integration/balancing_spec.rb#L79) uses the style without brackets
yes, adding RuboCop is a task I have since a long time on my todo list.

PR #810, opened 2021-02-05
Added factory "requirements"

lentschi:
see https://github.com/foodcoops/foodsoft/pull/765#discussion_r571050986

### DIFF
--- a/spec/factories/group_order.rb
+++ b/spec/factories/group_order.rb
@@ -2,8 +2,8 @@ require 'factory_bot'
 
 FactoryBot.define do
 
-  # requires order
   factory :group_order do
+    order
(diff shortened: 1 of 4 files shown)


wvengen (approved):
Thanks, this brings it more in line with the [FactoryBot best practice](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#best-practices):
> It is recommended that you have one factory for each class that provides the simplest set of attributes necessary to create an instance of that class. If you're creating ActiveRecord objects, that means that you should only provide attributes that are required through validations and that do not have defaults. Other factories can be created through inheritance to cover common scenarios for each class.

PR #811, opened 2021-02-07
On #805: Turn bigint IDs back into normal integers

lentschi:
see https://stackoverflow.com/questions/44390225/how-to-turn-off-bigint-primary-keys-in-rails-5-1
and https://edgeguides.rubyonrails.org/5_1_release_notes.html#active-record-notable-changes

This fixes an issue introduced by me in 6b98fb2dfd9b3e5353330720b7694fe7d38b5ccc. I think this wouldn't be likely to cause any observable errors in the current master, but it breaks the PR https://github.com/foodcoops/foodsoft/pull/762 (should it ever be merged.)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,13 +12,13 @@
 
 ActiveRecord::Schema.define(version: 2021_02_05_090257) do
 
-  create_table "article_categories", force: :cascade do |t|
+  create_table "article_categories", id: :integer, force: :cascade do |t|
     t.string "name", default: "", null: false
     t.string "description"
     t.index ["name"], name: "index_article_categories_on_name", unique: true
   end
 
-  create_table "article_prices", force: :cascade do |t|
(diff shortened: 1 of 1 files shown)


paroga:
> should it ever be merged.

I still want to merge it, since it's very valuable, but I really want to test it with a bunch of real world databases first, which I did not find the time yet. :disappointed:

lentschi:
> I still want to merge it, since it's very valuable, but I really want to test it with a bunch of real world databases first, which I did not find the time yet. disappointed

:+1: Sure - better not hurry something like that! That's why I still left the PR in draft status - I myself am unsure if this might cause havoc.
(Also the tests in this PR currently fail. The reason for this is, that the `shared_article` does not work with the new constraint on `article_category_id`. To fix this I must first find a way to test article sharing locally.)

PR #812, opened 2021-02-08
Add transport to OrderPDF

paroga:


wvengen:
Do we have a test for generating this document?
It would be good to test at least if it can be generated with and without transport (i.e. no error occurs).

paroga:
> Do we have a test for generating this document?

No, we don't have tests for documents. I agree that it would be great to have test, but I don't want to start a whole new category of tests, for this small change.

wvengen (approved):

PR #813, opened 2021-02-08
Add CSV downloads for BankTransaction and Invoice

paroga:


lib/bank_transactions_csv.rb
```diff
@@ -0,0 +1,29 @@
+require 'csv'
+
+class BankTransactionsCsv < RenderCSV
```
wvengen, line 3: `BankTransactionsCSV`?

wvengen:
Other than than, good to merge :+1:

lib/bank_transactions_csv.rb
paroga: it's aligned with `ArticlesCsv`, `OrderCsv` and `OrdergroupsCsv`

PR #814, opened 2021-02-08
Avoid page breaks during tables in OrderPdf 

paroga:


wvengen (approved):
Super \:/

(crossing fingers without a test, but it's an improvement so fine now as well)

PR #815, opened 2021-02-08
Add option to create a new FinancialTransaction at a FinancialLink

paroga:


app/controllers/finance/financial_links_controller.rb
```diff
@@ -36,7 +36,7 @@ def show
   end
 
   def create
-    @financial_link = FinancialLink.new
+    @financial_link = FinancialLink.first_unused_or_create
```
wvengen, line 39: It looks as if this would better be put in a transaction (creating the link and adding the bank transaction).

app/controllers/finance/financial_links_controller.rb
```diff
@@ -64,6 +64,23 @@ def remove_bank_transaction
     redirect_to finance_link_url(@financial_link), notice: t('.notice')
   end
 
+  def new_financial_transaction
+    @financial_transaction = FinancialTransaction.new(financial_link: @financial_link)
+    @financial_transaction.amount = @financial_link.amount
+    @financial_transaction.ordergroup_id = FinancialTransaction.joins(<<-SQL).order(created_on: :desc).pluck(:ordergroup_id).first
+      JOIN bank_transactions a ON financial_transactions.financial_link_id = a.financial_link_id
+      JOIN bank_transactions b ON a.iban = b.iban AND b.financial_link_id = #{@financial_link.id}
+    SQL
```
wvengen, line 73: This would be a candidate for moving to a model or another business-logic-object.
I'd put determining the ordergroup in a separate method somewhere, so that it can be more easily overridden (as it looks like a bit of a heuristic).

wvengen (commented):
I'd need to see this working in practice to fully get the idea, but looks ok from my perspective, with some notes.

app/controllers/finance/financial_links_controller.rb
paroga: why?
1) `first_unused_or_create` is a generic function, needed at different places. 
2) i see no problem happeing when adding the the bank transaction fails

wvengen: 1. To be clear: I mean in the controller, not in the method.
2. That's true! So we can leave it :+1:

app/controllers/finance/financial_links_controller.rb
paroga: done

app/controllers/finance/financial_links_controller.rb
```diff
@@ -102,4 +116,17 @@ def find_financial_link
...
+private
+
+  def financial_transaction_params
+    params.require(:financial_transaction).permit(:financial_transaction_type_id, :ordergroup_id, :amount, :note)
+  end
+
+  def find_best_fitting_ordergroup_id_for_financial_link(financial_link_id)
+    FinancialTransaction.joins(<<-SQL).order(created_on: :desc).pluck(:ordergroup_id).first
+      JOIN bank_transactions a ON financial_transactions.financial_link_id = a.financial_link_id
+      JOIN bank_transactions b ON a.iban = b.iban AND b.financial_link_id = #{financial_link_id}
```
wvengen, line 128: Thanks for separating this out!

We must be careful here to make sure this variable cannot be entered by the user.
In this case, it comes from `@financial_link`, which is a record loaded from the database, so it is fine.
Let's either add a note to the method about this, or add `.to_i`, which is tamper-proof in terms of SQL injection.
(I'm sorry I hadn't seen this earlier.)

paroga: done, better safe than sorry

wvengen (approved):

mjavurek:
I really appreciate this feature, it saves a lot of time for me!

PR #816, opened 2021-02-08
Add option to set all received items to zero

paroga:
When an order could not be delivered, this option makes it easy to
set all amounts to zero with only one click.


wvengen (approved):
Great option.

PR #818, opened 2021-02-08
Add/Improve financial_transaction_* endpoints

paroga:


app/serializers/financial_transaction_serializer.rb
```diff
@@ -2,11 +2,16 @@ class FinancialTransactionSerializer < ActiveModel::Serializer
...
+  attributes :financial_transaction_type_id, :financial_transaction_type_name
 
   def user_name
     show_user object.user
   end
 
+  def financial_transaction_type_name
+    object.financial_transaction_type.name
+  end
+
```
wvengen, line 14: I wonder if we should add a nested `financial_transaction_type` object here, or keep it with `id` and `name`?
Indeed account name and IBAN wouldn't really make sense here, so then we'd need separate serialization variants, so perhaps this approach is the best.

wvengen (commented):
Good to see this expanded.
Part of me doubts between having separate classes and types endpoints vs. a single classes endpoint with embedded types. I don't use these features enough to be able to judge, but the current approach _is_ the most basic / KISS approach, so let's do it (albeit in many cases one would get them both and do the joining client-side).

wvengen (approved):

app/serializers/financial_transaction_serializer.rb
paroga: i had the same thoughts, end ended with the current code :wink:

Issue #821, opened 2021-02-15
Increase readability of IBANs by adding whitespaces 


mariandl:
Request from ["Michael"](https://forum.foodcoops.at/t/leserlichkeit-von-iban/4718/3):

To increase the readabiliy of IBANs it can be useful to display them seperated by whitespaces:

- IBAN: EU10 0000 0000…

twothreenine:
Tried to do it [in this commit](https://github.com/twothreenine/foodsoft/commit/85f26a8d9555289f38d9919178268c8c66072cc3) in my fork.
Important note: I couldn't test it for finance/bank_transactions/show.html.haml

Seems I cannot create a pull request for a specific commit, only a branch?!
Also I'm very new to Ruby, so please check it carefully

PR #824, opened 2021-02-16
better readable IBAN

twothreenine:
Display IBANs in groups of 4 digits with whitespaces in between
solves #821 
I changed it in various views I could find.
Important note: I couldn't test it for finance/bank_transactions/show.html.haml
Also I'm very new to Ruby, so please check it carefully

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,7 +25,7 @@ module ApplicationHelper
     content_tag :span, number_to_currency(amount), class: class_name
   end
 
-  # Splits an IBAN into groups of 4 digits (.iban tag will add margins between each group)
+  # Splits an IBAN into groups of 4 digits displayed with margins in between
   def format_iban(iban)
     iban.scan(/..?.?.?/).map{ |item| content_tag(:span, item, style: "margin-right: 0.5em;") }.join.html_safe
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,6 +25,17 @@ module ApplicationHelper
     content_tag :span, number_to_currency(amount), class: class_name
   end
 
+  # Adds whitespaces to an IBAN for better readable display (groups of 4 digits)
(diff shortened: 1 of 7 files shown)


paroga:
cool, thanks for the PR.
one problem with your change is, that it's then not possible to just double-click it to mark it for copy&paste. to solve this you could add a "Copy to clipboard"-button or use some little CSS-tricks (as described in the [German Wikipedia](https://de.wikipedia.org/wiki/Internationale_Bankkontonummer#Schreibweise)).

if the generated HTML looks like
```html
<span class="iban">
<span>DE07</span><span>1234</span><span>1234</span><span>1234</span><span>1234</span>12
</span>
```
you could format it with the following CSS to add some padding:
```css
.iban > span {
  margin-right: 0.5em;
}
```
As long as there is no whitespace between the different `<span>` elements, it can be still double-clicked.

what do you think about that?

fruestueck:
Your already working on this? Awesome!

> if the generated HTML looks like
> 
> ```
> <span class="iban">
> <span>DE07</span><span>1234</span><span>1234</span><span>1234</span><span>1234</span>12
> </span>
> ```

Agreed. This is also preferrable as it would avoid introduction of whitespaces. Whitespaces could break insertion into banking forms (due to the character limit, and or bad support for whitespace within the iban)

twothreenine:
I managed to implement that and it works for the most pages, however when you copy an IBAN from the suppliers/show page, four whitespaces appear in front of it ("    AT1234123...") regardless of how many margins there are within the number.
I don't know how to fix this.
Also again I couldn't test it for finance/bank_accounts/index and finance/bank_transactions/show.

paroga:
> when you copy an IBAN from the suppliers/show page, it contains four whitespaces

I looks like a wired CSS effect to me, which we could just ignore for now, since it's very uncommon that people will try to copy the IBAN there.

`finance/bank_accounts/index` and `finance/bank_transactions/show` worked for me too.

PR #829, opened 2021-02-17
Split order close method

wvengen:
For clearer code and easier monkey-patching by plugins.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -25,7 +25,7 @@ class Order < ApplicationRecord
   validate :keep_ordered_articles
 
   # Callbacks
-  after_save :save_order_articles, :update_price_of_group_orders
+  after_save :save_order_articles, :update_price_of_group_orders!
   before_validation :distribute_transport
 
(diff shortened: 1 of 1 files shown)


app/models/order.rb
```diff
@@ -391,8 +383,21 @@ def distribute_transport
 
   # Updates the "price" attribute of GroupOrders or GroupOrderResults
   # This will be either the maximum value of a current order or the actual order value of a finished order.
-  def update_price_of_group_orders
-    group_orders.each { |group_order| group_order.update_price! }
+  def update_price_of_group_orders!
+    group_orders.each(&:update_price!)
```
paroga, line 387: i think this is more a method, which should be optimized, since it happens often :wink:

paroga (approved):

app/models/order.rb
wvengen: Yes, good one.
There are quite a few things that can be optimized here, but this one is indeed called often.

PR #831, opened 2021-02-17
Add method to parse bank transactions as JSON

paroga:
This implements parsing of the Account Information Service format as
defined in the Berlin Group Group NextGenPSD2 XS2A Framework, which
is widely used across various European banks.

This is a first step to replace the current bank import features with
a standardized JSON interface.

### DIFF
--- a/app/models/bank_account.rb
+++ b/app/models/bank_account.rb
@@ -24,4 +24,8 @@ class BankAccount < ApplicationRecord
     end
     count
   end
+
+  def last_transaction_date
+    bank_transactions.order(date: :desc).first&.date
+  end
 end
--- a/lib/bank_account_information_importer.rb
+++ b/lib/bank_account_information_importer.rb
@@ -1,0 +1,55 @@
+class BankAccountInformationImporter
+
+  def initialize(bank_account)
+    @bank_account = bank_account
+  end
+
+  def import!(content)
+    return nil if content.empty?
+    data = JSON.parse content, symbolize_names: true
+
+    return 0 if data.empty?
+    booked = data.fetch(:transactions, {}).fetch(:booked, [])
+
+    ret = 0
+    booked.each do |t|
+      amount = parse_account_information_amount t[:transactionAmount]
+      entityName = amount < 0 ? t[:creditorName] : t[:debtorName]
+      entityAccount = amount < 0 ? t[:creditorAccount] : t[:debtorAccount]
+
+      @bank_account.bank_transactions.where(external_id: t[:transactionId]).first_or_create.update({
+        date: t[:bookingDate],
+        amount: amount,
+        iban: entityAccount && entityAccount[:iban],
+        reference: t[:remittanceInformationUnstructured],
+        text: entityName,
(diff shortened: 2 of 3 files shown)


app/models/bank_account.rb
```diff
@@ -24,4 +24,55 @@ def assign_unlinked_transactions
...
+
+  def last_transaction_date
+    bank_transactions.order(date: :desc).first.try(:date)
+  end
+
+  def import_account_information!(content)
+    return nil if content.empty?
+    data = JSON.parse content, symbolize_names: true
+
+    booked = data.try(:fetch, :transactions, nil).try(:fetch, :booked, nil) || []
```
wvengen, line 36: Do you agree that this use of `try` reduces readability?
What about:
```
return nil if data.empty?

booked = data.fetch(:transactions, {}).fetch(:booked, [])
```

app/models/bank_account.rb
```diff
@@ -24,4 +24,55 @@ def assign_unlinked_transactions
...
+    save!
+
+    ret
+  end
+
+  private
+
+  def parse_account_information_amount(value)
+    value && value[:amount].to_f
+  end
```
wvengen, line 77: It seems to me that this doesn't really belong in the model, but in an importer class that takes the file and updates the database.

wvengen (commented):
Nice :) And great you added tests.
Too bad banks in NL are lagging with this.

app/models/bank_account.rb
paroga: agree

app/models/bank_account.rb
paroga: moved to `lib`

lib/bank_account_information_importer.rb
```diff
@@ -0,0 +1,55 @@
...
+    %w(closingBooked expected authorised openingBooked interimAvailable forwardAvailable nonInvoiced).each do |type|
+      value = balances[type]
+      if value then
+        balance = value
+        break
+      end
+    end
+
+    @bank_account.balance = parse_account_information_amount(balance) || @bank_account.bank_transactions.sum(:amount)
+    @bank_account.import_continuation_point = booked.first.try(:fetch, :entryReference, nil)
```
wvengen, line 42: `booked.first&.fetch(:entryReference, nil)`

app/models/bank_account.rb
```diff
@@ -24,4 +24,8 @@ def assign_unlinked_transactions
     end
     count
   end
+
+  def last_transaction_date
+    bank_transactions.order(date: :desc).first.try(:date)
+  end
```
wvengen, line 30: I'm a little hesitant, because it invites developers to use the method, and it can't be preloaded.
That said, it makes sense to put it here, so let's leave it.

wvengen: Also, I think we can use the `&.` operator, that in my view is more readable (and less lenient):
`bank_transactions.order(date: :desc).first&.date`

wvengen (commented):
Thanks for the changes.
Some small things, otherwise good to go!

PR #835, opened 2021-02-18
Add API v1 finance_overview endpoint

paroga:


config/routes.rb
```diff
@@ -265,6 +265,7 @@
 
         namespace :user do
           root to: 'users#show'
+          get :finance_overview,  to: 'users#finance_overview'
```
wvengen, line 268: perhaps "financial_overview" is cleaner here

app/controllers/api/v1/user/users_controller.rb
```diff
@@ -6,4 +6,22 @@ def show
     render json: current_user
   end
 
+  def finance_overview
```
wvengen, line 9: What are your thoughts on scope `user:read` vs. `finance:user`?
Since this really is a financial overview, I would slightly favour the latter, with `user:read` more for user details and membership. What do you think?

app/controllers/api/v1/user/users_controller.rb
```diff
@@ -6,4 +6,22 @@ def show
...
+          {
+            id: c.id,
+            name: c.display,
+            amount: ordergroup["sum_of_class_#{c.id}"]
+          }
+        }
+      }
+    }
+  end
+
```
wvengen, line 26: I assume you've thought about whether to do this with ActiveModelSerializer.

Other question: where will we expose Ordergroup information? We don't really, right now, and it is perhaps related to this.

app/controllers/api/v1/user/users_controller.rb
paroga: i have no clear thought about our scopes yet. i'll go with your proposal

app/controllers/api/v1/user/users_controller.rb
paroga: i don't see a benefit of a serializer in this case. i'd refactor, if we decide that there's a better way

do you you mean with _Ordergroup information_? we expose our ordergroup-related `finacial_transactions` already via `/user/financial_transactions`.

config/routes.rb
paroga: agree

app/controllers/api/v1/user/users_controller.rb
wvengen: True. The API doesn't currently return e.g. the name of the user's ordergroup, or whether it has one.
If we were to add that, would there be overlap with `financial_overview`?

One way would be to let `/api/user` also return group memberships, and have `/api/ordergroup` show ordergroup fields. Another way would be to only have `/api/user/ordergroup` with ordergroup fields (and then perhaps `/api/user/workgroups` for workgroup memberships). The user's ordergroup endpoint would either include the financial overview, or there would be a separate `/api/user/financial_overview` endpoint.

Since any choice we make will stick for quite a while, I'd like to put some thought to it.
What do you think?

app/controllers/api/v1/user/users_controller.rb
wvengen: :+1:
In that case I would propose an ordergroup controller in the API that includes this method.

app/controllers/api/v1/user/users_controller.rb
paroga: IIRK the basic idea of `/api/user` was to make dealing with access roles easier: every user can request `/api/user/financial_transactions` to get her own transactions, but only finance-people can request `/api/financial_transactions` to get **all** transactions.
sticking to that idea `/api/user/finance_overview` should be the way to go. I see no benefit in moving it to `/api/user/ordergroup/finance_overview`, since a user has only one and we would need to move the `/api/user/financial_transactions` to `/api/user/ordergroup/financial_transactions` then too. IMHO renaming `/api/user` to `/api/me` or `/api/my` would make the basic idea more clear, but we settled with `/api/user` already.

wvengen: And what you've done now seems even better, on second thought: putting it in an ordergroup controller. :+1:

wvengen (approved):

PR #836, opened 2021-02-20
Improve reference calculator

twothreenine:
- fix rounding error: previously, some numbers like 38.20 (which is something like 38.2000000001) were ceiled to 38.21 (was there a specific reason to use "ceil" instead of "round"?)
- always display final amount with two decimals
- add currency symbol to input fields and final amount
- better(?) headline for transaction types - you could also use "type" (as in transaction type) which would be closer to the model, but I think "purpose" is better understandable and fitting in this context.

### DIFF
--- a/app/views/home/reference_calculator.haml
+++ b/app/views/home/reference_calculator.haml
@@ -12,7 +12,7 @@
         $('input[data-has-name-short]').each(function() {
           var bank_account = $(this).attr('data-bank-account') || default_bank_account;
           var name = $(this).attr('name');
-          var value = Math.ceil(parseFloat($(this).val()) * 100) / 100;
+          var value = Math.round(parseFloat($(this).val()) * 100) / 100;
           if (value > 0) {
             var item = items[bank_account];
             if (!item) {
@@ -32,7 +32,7 @@
           console.dir(item);
           if (item) {
             ele.show();
-            ele.find('.amount').text(item.amount);
+            ele.find('.amount').text(item.amount.toFixed(2));
             ele.find('.reference').text((#{raw BankTransactionReference.js_code_for_user(@current_user)})(item.values));
           } else {
             ele.hide();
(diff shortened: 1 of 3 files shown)


app/views/home/reference_calculator.haml
```diff
@@ -12,7 +12,7 @@
         $('input[data-has-name-short]').each(function() {
           var bank_account = $(this).attr('data-bank-account') || default_bank_account;
           var name = $(this).attr('name');
-          var value = Math.ceil(parseFloat($(this).val()) * 100) / 100;
+          var value = Math.round(parseFloat($(this).val()) * 100) / 100;
```
paroga, line 15: What's the rational for this change? I choose `ceil` to make sure, that people always get an amount which includes all of the required money. If 3 people they to pay 1/3 of 10 entering `3.333` will be rounded to `3.33` and results in `9.99`. So the foodcoop misses `0.01` otherwise it gets `0.02` plus (in case people do not enter the exact values anyway).

paroga (commented):
LGTM, beside the `Math.ceil`

app/views/home/reference_calculator.haml
twothreenine: As I already mentioned, with `ceil` some numbers with two decimals are wrongly ceiled to the next 0.01 step. Try `1.09`, `1.10`, `1.11`, `1.12` for example. The reason is that they are handled as `float` and not cut off at the last digit entered.
refcalc

> If 3 people they to pay 1/3 of 10 entering 3.333 

I don't understand that sentence, why should several people (an ordergroup consisting of 3 people?) enter a single fraction into the calculator? Or do you mean if someone wants to pay 10.00 in total, split evenly on 3 transaction types?
I don't see why this calculator should handle more than two decimals in the first place. The input field turns red as a warning if more than two decimals are entered.

PR #838, opened 2021-02-23
Update de.yml for minimum order quantity - currency independant

mjavurek:
"Wert" repräsentiert für mich eher einen Geldbetrag als "Menge", daher wird die deutsche Übersetzung aus meiner Sicht klarer.

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -177,8 +177,8 @@ de:
         fax: Fax
         iban: IBAN
         is_subscribed: abonniert?
-        min_order_quantity: Mindestbestellmenge
-        min_order_quantity_short: Menge (mind.)
+        min_order_quantity: Mindestbestellwert
+        min_order_quantity_short: Wert (mind.)
         name: Name
         note: Notiz
         order_howto: Kurzanleitung Bestellen


wvengen:
Thank you for your PR! I have been confused by the meaning of this. The field is called _quantity_, but at some point someone [added some logic](https://github.com/foodcoops/foodsoft/blob/007ff70c486be5d93cb189fa4e13327270436e62/app/models/order.rb#L311) related to _monetary amount_. I have used it in the past for a textual description. And indeed, the field is a string, as can be seen in https://github.com/foodcoops/foodsoft/blob/master/db/schema.rb#L473. This is not an ideal situation.

Maybe we need to find clarity on its use in the (user) mailing-list. And introduce a monetary field separate from a 'free-form' quantity field (if it is used at all). Perhaps there are other uses (e.g. weight or crates or whatever).

paroga:
> I'd suggest adding a text field order_note to supplier, which prefills the order_note field of each order opened for this supplier.

What's the benefit compared to copying the order? If we add a `order_note` to the `Order` I'd prefer to add it in **addition** to the existing note.

I like the idea of extending the custom fields to the supplier and your migration path for existing data.

twothreenine:
Sorry, I meant a field `supplier.order_note `which prefills / acts as a default for the `order.note`.

> What's the benefit compared to copying the order? If we add a order_note to the Order I'd prefer to add it in addition to the existing note.

You wouldn't have to click "copy" each time and could open orders normally, while you would still be able to adapt it in case the conditions differ for a specific order. (Also, copying an order does not copy the note atm)
What's the benefit of an additional `order.order_note `field?

Issue #722, opened 2020-02-26
[v4.6.99] Transaction collections ignores decimals


twothreenine:
When entering transactions via finance/transactions/new_collection, the amounts are processed as integers, ignoring the decimals. Only by entering singular transactions to a specific ordergroup, it is possible to enter amounts with decimals.

twothreenine:
I found out that it works with decimal point, just not with decimal comma.
Entering a singular transaction works both with decimal point and decimal comma.
There are more cases in which only decimal point works, like received units per group in an order:
![split-](https://user-images.githubusercontent.com/22921479/83052070-172e6b00-a04f-11ea-97d5-039d99205e29.png)

And then there is a case where decimal point seems to work at first, but it is not processed: On the "receive" page
![receive](https://user-images.githubusercontent.com/22921479/83052295-67a5c880-a04f-11ea-9b64-65ff48f6e194.png)
After submitting, the numbers are rounded to the nearest integers.

wvengen:
Thanks for your bug report. That's quite odd, especially the receive screen-case.
The code used for allowing both comma and point is really old, it might suffer from bit-rot. Would be good to review.
The database should have a decimal for these fields.

twothreenine:
> When entering transactions via finance/transactions/new_collection, the amounts are processed as integers, ignoring the decimals. Only by entering singular transactions to a specific ordergroup, it is possible to enter amounts with decimals.

This has been resolved in #921 👍

Could we use a similar logic to fix the balancing menu case, too, or is that a different problem?

PR #839, opened 2021-02-25
add currency symbols & improve amount fields in various views

twothreenine:
Currency symbols added to various forms and other views.
The new_collection form now also supports decimal commas, fixing #722 (at least the initial issue)


twothreenine:
I'd like some feedback on this (and #840 ) before I move on to the next fixes/features ;)

paroga:
sorry for the late reply, unfortunately this is not an easy change.

You changed the type of the input fields to `number`. I tried already something similar some time ago, but stopped it, since I wasn't able to get browsers to behave consistent. Did you test **ALL** changed fields with **multiple** browsers and **multiple** locales? This change has potential to break a lot.

The main issue happens when the view prints numbers in german locale (e.g. `12,34`), but the input field of type number expects english locales (e.g. `12.34`).

RayOei:
Seems related to the (newer) #1060
It is also in `Draft` so...  should it be taken together?

twothreenine:
Some of the commits in this PR are obsolete, so it shouldn't be merged as it is.
Regarding the minimum order quantity, I made an alternative suggestion in #1054 - what do you think?
The amount fields in new_collection and new transaction have been fixed to allow commas.
Adding currency symbols in various views is still relevant but far from critical.

RayOei:
As it is still in `draft` (thus not finished I assume) and apparently outdated in parts, I would suggest to close this PR and transfer anything which _is_ useful to a (or more) new branch(es). That way the things which are useful remain in the pipeline.
It is also much better to have specific branches for specific changes to not convolute a branch with (too) many different things which makes merging small changes also much easier and more likely to happen. Which is ultimately also more rewarding for any contributor in the end 😁

twothreenine:
Ok, I noted it in a new issue and will close this.

> It is also much better to have specific branches for specific changes to not convolute a branch with (too) many different things which makes merging small changes also much easier and more likely to happen.

Yeah, this was one my first PRs and I didn't have much experience.

PR #840, opened 2021-02-25
Improve various dl-horizontal views

twothreenine:
I raised the width of dl-horizontal titles (used on pages like suppliers/show, group_orders/show, and here finance/invoices/show as well), as some titles, esp. in German, were cut off before. I'm not sure if this could have negative effects on small screens like on mobile devices. I guess there has been a reason this wasn't changed.
Also I added some 'line spacing' via margin-bottom to increase the readability of those pages.

With these changes, we could also use dl-horizontal for the invoices/show page, which would again improve readability and make it easier to compare dates, amounts, deposit etc.

### DIFF
--- a/app/views/finance/invoices/show.html.haml
+++ b/app/views/finance/invoices/show.html.haml
@@ -2,74 +2,76 @@
 
 - total = 0
 
-%p
-  %b= heading_helper(Invoice, :created_at) + ':'
-  = format_time(@invoice.created_at)
-%p
-  %b= heading_helper(Invoice, :created_by) + ':'
-  = show_user(@invoice.created_by)
-%p
-  %b= heading_helper(Invoice, :supplier) + ':'
-  = @invoice.supplier.name
+.row-fluid
+  .span6
+    %dl.dl-horizontal
+      %dt= heading_helper(Invoice, :created_at) + ':'
+      %dd= format_date(@invoice.created_at)
 
-- if @invoice.deliveries.any?
-  %p
-    %b= heading_helper(Invoice, :deliveries) + ':'
-    %span><
-      - @invoice.deliveries.order(:date).each_with_index do |delivery, index|
-        - sum = delivery.sum
-        - total += sum
-        = ', ' if index > 0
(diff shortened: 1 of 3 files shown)


app/views/finance/invoices/show.html.haml
```diff
@@ -2,74 +2,76 @@
...
+            = ' (' + number_to_currency(sum) + ')'
+
+      - if @invoice.orders.any?
+        %dt= heading_helper(Invoice, :orders) + ':'
+        %dd><
+          - @invoice.orders.order(:ends).each_with_index do |order, index|
+            - sum = order.sum
+            - total += sum
+            = ', ' if index > 0
+            = link_to format_date(order.ends), new_finance_order_path(order_id: order) #, style: "margin-right: 0.2em;"
```
paroga, line 34: can you please remove the unnecessary comment

paroga (commented):
small issue, otherwise good to go. thanks!

twothreenine:
Just a note for the foodcoopsat fork: Please be aware that this collides with [this commit](https://github.com/foodcoopsat/foodsoft/commit/ef13628cd20d6aab6bd11f269bbb54d575a52a11) (number of indents has to aligned)

paroga:
sorry I missed the change for the `created_at` in my last review: Is there a reason why you changed it to date-only? Sometimes it's quite interesting when members created invoices. E.g. sometimes I wondered why I missed to pay an invoice and then saw that it has just been created a few minutes after paying all unpaid ones.

twothreenine:
The date was not formatted with `format_time`. When you switch to English, you still see the German representation, unlike the other dates in the invoice, which would be confusing when you are trying to compare them.
`format_datetime` gives a very long representation like `Tue, 16 Mar 2021 22:27:55 +0000` and I thought the time is unnecessary anyway.
We could add a short version of `format_datetime` which just returns the date as in `format_date` + `hh:mm`

paroga:
we can change `format_time` in [ApplicationHelper](https://github.com/foodcoopsat/foodsoft/blob/master/app/helpers/application_helper.rb#L6-L8) to provide a localized representation, but i'd consider that an independent issue

Issue #788, opened 2020-11-17
Mail broken for single coop install


alexanderharm:
Hi, I tried setting up a public mailserver using the docker image. However the logs always complain about this:

```
  /usr/src/app/vendor/bundle/ruby/2.6.0/gems/railties-5.2.4.3/lib/rails/railtie.rb:190:in `public_send'
  /usr/src/app/vendor/bundle/ruby/2.6.0/gems/railties-5.2.4.3/lib/rails/railtie.rb:190:in `method_missing'
```

I tried to set all env's as outlined in <https://github.com/foodcoops/foodsoft/blob/master/config/environments/production.rb>. Not sure how to fix this. Any pointers?

alexanderharm:
I verified credentials again and tested with Discourse as well since it uses the same mailer. It works flawlessly there.

The Foodsoft server however, returns an error 500 when posting to e. g. /f/invites

wvengen:
Thanks for reporting the issue. Can you give a bit more context: when exactly is this happening? Can you give the full stracktrace?
You can also look at the [foodcoops.net setup](https://github.com/foodcoops/foodcoops.net/blob/master/docker-compose.yml#L18), where I expect sending mails to work.

alexanderharm:
The foodcoop.net setup uses a separate container for mail delivery. I was hoping that I could deliver straight to my provider instead.

wvengen:
That's right. You should be able to use a direct connection too, but then some authentication is necessary, which I haven't tested. This should be standard Rails configuration (see [this guide](https://guides.rubyonrails.org/v5.2/action_mailer_basics.html#action-mailer-configuration), which you probably have read already).

If you use a non-encrypted connection (e.g. with CRAM-MD5 authentication, or a temporary password), you could use tcpdump / wireshark / ... to see what happens in the connection.

alexanderharm:
I will test and come back...

alexanderharm:
I fixed one issue (sending messages) because of Redis problems. However, inviting users still fails:

```
I, [2021-02-18T11:24:58.175080 #52]  INFO -- : [58d429d6-92a6-4a41-851e-bd78df2175e2]   Parameters: {"authenticity_token"=>"Da+WevP0maM0F57zv5azKUfbrWEP0vhHGZUu0WGV8Q7Sy0LPuacFYSPxFqK+G8aD1TUIww7kuI8eiDfRamC+8g==", "foodcoop"=>"f", "id"=>"8"}
[...]

PR #841, opened 2021-02-26
Fix single coop issues

wvengen:
- non-background mailing didn't work
- the error handler had an issue

#788

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -11,15 +11,6 @@ class ErrorsController < ApplicationController
 
   private
 
-  def select_foodcoop
-    foodcoop = params[:foodcoop]
-    if FoodsoftConfig.allowed_foodcoop? foodcoop
-      FoodsoftConfig.select_foodcoop foodcoop
-    else
-      FoodsoftConfig.select_default_foodcoop
-    end
-  end
-
   def current_layout
     current_user ? 'application' : 'login'
   end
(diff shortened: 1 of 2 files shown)


app/controllers/errors_controller.rb
```diff
@@ -11,15 +11,6 @@ def show
...
 
-  def select_foodcoop
-    foodcoop = params[:foodcoop]
-    if FoodsoftConfig.allowed_foodcoop? foodcoop
-      FoodsoftConfig.select_foodcoop foodcoop
-    else
-      FoodsoftConfig.select_default_foodcoop
-    end
-  end
-
```
wvengen, line 22: This is already provided by the `ApplicationController`, and that handles single coop installs better.

paroga: i'd create a separate commit for that, since it's unrelated

lib/foodsoft_config.rb
```diff
@@ -75,6 +75,11 @@ def init(filename = APP_CONFIG_FILE)
       self.scope = config[:default_scope] or raise "No default_scope is set"
       # Set defaults for backward-compatibility
       set_missing
+      # When multicoops is enabled, the foodcoop is selected on each request.
+      # When multicoops is disabled, we need to do some initialization.
+      unless config[:multi_coop_install]
```
paroga, line 80: do we really need the `unless`. if not, i'd omit it

paroga (approved):

lib/foodsoft_config.rb
wvengen: We don't (I'll test) :+1:

app/controllers/errors_controller.rb
wvengen: Ah, thanks - I intended to but then forgot!

PR #842, opened 2021-02-27
Received mailer for #766

lentschi:
As discussed with @paroga the "received" mail should be decoupled from the reset of #766


config/locales/de.yml
```diff
@@ -1283,6 +1283,17 @@ de:
 
 
         Viele Grüße von %{foodcoop}
+    order_received:
+      subject: 'Bestellung entgegengenommen: %{name}'
+      text0: |
+        Liebe Bestellgruppe %{ordergroup},
+        die Bestellung für "%{order}" wurde geliefert.
```
paroga, line 1290: add a newline?

config/locales/en.yml
```diff
@@ -1294,6 +1294,17 @@ en:
 
 
         Kind regards from %{foodcoop}.
+    order_received:
+      subject: 'Order delivery registered: %{name}'
+      text0: |
+        Dear %{ordergroup},
```
paroga, line 1300: add newline after "Dear"?

app/views/mailer/order_received.text.haml
```diff
@@ -0,0 +1,13 @@
...
+  = raw "===\n" + t('.abundant_articles') + ":"
+  - @abundant_articles.each do |order_article|
+    - article = order_article.article
+    = raw t('.article_details', name: article.name, ordered: order_article.units_to_order, received: order_article.units_received, unit: article.unit)
+
+- unless @scarce_articles.empty?
+  = raw "===\n" + t('.scarce_articles') + ":"
+  - @scarce_articles.each do |order_article|
+    - article = order_article.article
+    = raw t('.article_details', name: article.name, ordered: order_article.units_to_order, received: order_article.units_received, unit: article.unit)
```
paroga, line 13: missing newline at end of file

app/views/mailer/order_received.text.haml
```diff
@@ -0,0 +1,13 @@
+= raw t '.text0', ordergroup: @group_order.ordergroup_name, order: @order.name
+
```
paroga, line 2: what do you think about adding the ordered articles too? (maybe with received units after distribution)

something like:
```
you ordered:
 * 2 kg carrots  12€
 * 7 kg potatoes  34€

the following articles where received with a different amount
* potatoes

according to the received amount the designated amount for your ordergroup is:
 * 2 kg carrots  12€
 * 5 kg potatoes.
```

if we put all the information into the `received` mail, the `ordered` mail should get obsolete for most (all?) use cases.

paroga (commented):
LGTM, just some small ideas. if you don't agree, i'm ok to merge it as is

app/views/mailer/order_received.text.haml
lentschi: > what do you think about adding the ordered articles too? 

Hm... That would be too verbose for my own personal use case as I wouldn't use the e mail to check what I've ordered (Are there users that do that...?), but just as an info that I should go and fetch stuff and to warn me something has been delivered too much or too little of. (I would then use a PDF export or the self service page to check the `availability` and `ordered` columns when fetching articles)
But if there is a use case for that, I could maybe add another sub config option ('only include articles where the delivered amount deviates') to show the email in its current format and if not in your suggested format...?

> maybe with received units after distribution

Well, when used with my new self service page this value is hardly required: They'd just view the composite `availability` column that auto-updates as other users fetch stuff. (Nothing really is _distributed_. Things are _taken_.)
For the other user cases I could add that though, yes. (I'll just wait for an answer to the above topic, as adapting this would change the email's format completely.)

PR #844, opened 2021-03-01
Add Rubocop

paroga:


.rubocop_todo.yml
```diff
@@ -0,0 +1,991 @@
...
+  Exclude:
+    - 'plugins/current_orders/foodsoft_current_orders.gemspec'
+    - 'plugins/discourse/foodsoft_discourse.gemspec'
+    - 'plugins/documents/foodsoft_documents.gemspec'
+    - 'plugins/links/foodsoft_links.gemspec'
+    - 'plugins/messages/foodsoft_messages.gemspec'
+    - 'plugins/polls/foodsoft_polls.gemspec'
+    - 'plugins/printer/foodsoft_printer.gemspec'
+    - 'plugins/uservoice/foodsoft_uservoice.gemspec'
+    - 'plugins/wiki/foodsoft_wiki.gemspec'
```
wvengen, line 30: Could this become something like `plugins/*/*.gemspec` in `.rubocop.yml`?
As these plugins are bundled with Foodsoft, they'd have the same minimum version as Foodsoft.

wvengen (commented):
Super!!
One comment, and the specs seem to fail (no, I didn't review _all_ changes).

.rubocop_todo.yml
paroga: done

paroga:
i reduced this change to whitespace only, so that it should not change any behavior

wvengen (approved):
:tada:

PR #846, opened 2021-03-02
Make usage of Docker for development easier by avoiding COPY operations

paroga:


Dockerfile-dev
```diff
@@ -25,15 +25,6 @@ USER app
 
 WORKDIR /home/app/src
 
-# Copy files needed for installing gem dependencies, and install them.
-COPY Gemfile Gemfile.lock ./
-COPY plugins ./plugins
-RUN bundle config build.nokogiri "--use-system-libraries" && \
```
wvengen, line 31: does `--use-system-libraries` still have any use, e.g. not having to compile libxml and friends at `bundle install` time?

doc/SETUP_DEVELOPMENT_DOCKER.md
```diff
@@ -28,19 +28,20 @@ afterwards.
 
 ## Setup
 
-Then setup foodsoft development (this will take some time, containers needs
-to be pulled from docker registry and a lot dependencies needs to be installed)
+Then start the databasee server and setup foodsoft development (this will take
```
wvengen, line 31: small typo

doc/SETUP_DEVELOPMENT_DOCKER.md
```diff
@@ -28,19 +28,20 @@ afterwards.
...
+    docker-compose -f docker-compose-dev.yml run --rm foodsoft \
+      bundle exec rake foodsoft:setup_development_docker
 
-Then create an initial database (here seeded with `small.en`) by running
+Optionally an initial database (here seeded with `small.en`) can be loaded by running
 
-    docker-compose -f docker-compose-dev.yml up -d mariadb && sleep 15
     docker-compose -f docker-compose-dev.yml run --rm foodsoft \
-      bundle exec rake db:create db:schema:load db:seed:small.en
+      bundle exec rake db:schema:load db:seed:small.en
```
wvengen, line 44: Let's be consistent in whether to use `\` or not (here and elsewhere in the file).

lib/tasks/foodsoft_setup.rake
```diff
@@ -34,10 +34,20 @@ namespace :foodsoft do
     setup_development
     setup_database
     start_mailcatcher
-    puts yellow "All done! Your foodcoft should be running smoothly."
+    puts yellow "All done! Your foodsoft should be running smoothly."
```
wvengen, line 37: What about "Your Foodsoft installation should be running smoothly." ? (or s/installation/setup/)

wvengen (commented):
Great to make this better.

doc/SETUP_DEVELOPMENT_DOCKER.md
paroga: IMHO it's consistent. It's only missing when there are multiple commands (vs. multiline commands)

wvengen: Yes, I see now that later invocations with `docker-compose-dev` don't use it - which is good.

paroga:
I removed the user completely, since it works only if the uid `1000` matches (usually the first user in the system)...

wvengen:
That's right, about the uid.
I'm a little hesitant about running Foodsoft as `root` by default, though (also to be more similar to production).
Could we avoid that?

paroga:
as long as we can't assure that the uid in and outside the container matches this brings a bunch of permission issues. I agree with not running as root for production, but for development in docker? I found no better EASY solution for it, but feel free to bring up something better.

wvengen:
Ok, maybe we can allow `root` for development. Let's see if it puts new developers off or not ;)
The only option would be to add all relevant files and dirs in Foodsoft's root directory as volumes - not pretty.

wvengen (approved):
Let's do it! :-)

PR #849, opened 2021-03-05
Add BankGateway

paroga:
This is the first step to replace the integrated import functionality with a clean call to an external api endpoint. The full implementation can be found in a [branch in my repository](https://github.com/paroga/foodsoft/tree/bgi).

### DIFF
--- a/app/views/admin/finances/index.html.haml
+++ b/app/views/admin/finances/index.html.haml
@@ -3,6 +3,7 @@
 - content_for :actionbar do
   = link_to t('.new_financial_transaction_class'), new_admin_financial_transaction_class_path, remote: true, class: 'btn btn-primary'
   = link_to t('.new_bank_account'), new_admin_bank_account_path, remote: true, class: 'btn'
+  = link_to t('.new_bank_gateway'), new_admin_bank_gateway_path, remote: true, class: 'btn'
 
 - content_for :sidebar do
   %p= t('.first_paragraph').html_safe
@@ -12,3 +13,6 @@
 
 %h2= t('.bank_accounts')
 #bank_accounts_table= render 'bank_accounts'
+
+%h2= t('.bank_gateways')
+#bank_gateways_table= render 'bank_gateways'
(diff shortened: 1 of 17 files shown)


app/views/admin/finances/index.html.haml
```diff
@@ -12,3 +13,6 @@
 
 %h2= t('.bank_accounts')
 #bank_accounts_table= render 'bank_accounts'
+
+%h2= t('.bank_gateways')
+#bank_gateways_table= render 'bank_gateways'
```
wvengen, line 18: Perhaps we need to add `- if @bank_gateways.any?` (and on that note, `@bank_accounts.any?`) to make this prettier when not using this functionality. Or let it also depend on `use_iban`?

wvengen (commented):
It's a little hard for me to review, even with the full implementation, but as far as I can see, it looks fine. I don't full get the unattended user thing, would you mind to explain?

Also, this could really use some documentation. For developers on how the different parts relate to each other. It's probably straightforward for you - but for newcomers some higher-level description would be good. Perhaps as yardoc class documentation, or else something in `doc/`.
And ultimately also for users, in what functionality this brings and how to set this up, and e.g. which countries and banks support this kind of interaction.

app/views/admin/finances/index.html.haml
paroga: i don't see the benefit of hiding it here, since it's a configuration page. we don't hide `bank_accounts` either

paroga:
> I don't full get the unattended user thing, would you mind to explain?

if you have a cron-job for importing transactions (aka unattended import) the gateway still needs a user on whose behalf the import happens and who is responsible for configuring the access. usually the credentials to log into you bank account are bound to one specific person and not to an organization

i 100% agree that we need documentation how the bank gateway api works, but i don't see a need to document the data model of this PR. a sane developer can look into `app/models/bank_gateway.rb` and read `has_many :bank_accounts` to see how it relates in the remaining model.

app/views/admin/finances/index.html.haml
wvengen: My bad, I was thinking of finances overview, good to have here!

wvengen:
Ok, if we have user documentation, then the data model will be clear enough.

wvengen (approved):
Curious to see what will follow :)

PR #851, opened 2021-03-07
Use ActiveJob instead of Resque directly

paroga:
This allows us to directly pass entities to the job.

This is the new version of #752.


config/application.rb
```diff
@@ -47,6 +47,9 @@ class Application < Rails::Application
     # TODO Disable this. Uncommenting this line will currently cause rspec to fail.
     config.action_controller.permit_all_parameters = true
 
+    config.active_job.queue_adapter     = :resque
+    config.active_job.queue_name_prefix = "foodsoft_#{Rails.env}"
```
wvengen, line 51: Can we include an environment variable, so that one could run more easily multiple instances of Foodsoft with one Redis instance? e.g.

```ruby
config.active_job.queue_name_prefix = ENV.fetch('JOB_QUEUE_PREFIX', "foodsoft_#{Rails.env}")}
```

config/initializers/active_job_select_foodcoop.rb
```diff
@@ -0,0 +1,22 @@
...
+        FoodsoftConfig.select_multifoodcoop arguments.shift
+        orig_deserialize(arguments)
+      end
+    end
+  end
+end
+
+ActiveSupport.on_load(:after_initialize) do
+  ActiveJob::Arguments.include FoodsoftActiveJobArguments
+end
```
wvengen, line 22: Cool. I would include this manually in each job, if possible. This allows people to write a Foodsoft plugin that does _not_ use a foodcoop scope (e.g. an overview of all instances).

wvengen (commented):
Good work, some comments.

config/initializers/active_job_select_foodcoop.rb
paroga: nope, since we need it for the internal rails jobs (e.g. for active storage)

wvengen: Good point. Then we'll need find a solution later if such a plugin appears.

wvengen (approved):

PR #852, opened 2021-03-07
Add support for ActiveStorage

paroga:


wvengen (approved):
Thanks!

One question: do we need to do something with docker-compose regarding the volume?

paroga:
> do we need to do something with docker-compose regarding the volume?

not with the one in this repository. it would be cleaner to configure it in the production setup (if no other storage backend will be configured), but it's not required since there is a `VOLUME` statement in `Dockerfile` a new unnamed volume should be created automatically

wvengen:
> should be created automatically

Great, that's what I wanted to check.
Would you mind updating the foodcoops.net docker-compose setup as well, so that we are prepared?
Reminder: we'll need to add something about this to the release notes.

kidhab:
> Would you mind updating the foodcoops.net docker-compose setup as well, so that we are prepared?

Is this still necessary? If not it would be nice to merge this into master.

PR #853, opened 2021-03-07
Use decimal to store number of billed/received units

paroga:


wvengen (approved):
I don't think this would have any serious side-effects. Perhaps when redistributing (after receive), we may need to check that no more units are distributed than have come in (rounding). Did you check that? If yes, then feel free to merge.

paroga:
i didn't check that, but I also don't agree to restrict that here. `units_received` and `units_billed` are not 100% connected to each other. we have the red exclamation mark in the balancing page to indicate differences. as long as people only use integer numbers nothing changes...

wvengen:
Yes, I'm ok with the balancing page.
When receiving, people can now enter partial amounts, and I'm not sure how these are being distributed.
But, as entering integers indeed remains the same, I'd be ok with merging. We can fix that later if necessary.

wvengen:
@paroga would you mind finishing this, as #868 depends on this? (I'll ask you just to be sure it's really finished.)

Issue #848, opened 2021-03-05
Add option to add financial link when accounting an order & improve access to financial link feature


twothreenine:
Financial links are a really nice feature with which you can group associated financial transactions, invoices, and bank transactions.
Currently it is a bit hidden: In the new_collection transaction form, the option is only shown if there is a bank account, while in the `Manage accounts` menu it is only shown if there already is at least one financial link.
I suggest to add a config option for this (like `use financial links on/off` in the finance tab) to make it better accessible and since it could also be useful if you don't have a bank account in your foodsoft instance.

Additionally, there should be an option (checked by default if the config option is on) to create a financial link when accounting an order. It should include all ordergroup charges as well as the invoice (in case there is one linked to the order).
Currently, when you want to add a financial transaction to a financial link, all the non-linked ordergroup charges show up, making the popup very congested, so it would be nice if they are linked.

PR #860, opened 2021-03-16
Improve access to financial link feature & add option to create link when balancing an order

twothreenine:
Implementation of #848 
I didn't know how to automatically check the checkbox to create a financial link on the confirm page, but I don't know if that would be desired anyway.
I guess we also would have to update [automatic assignment to invoice](https://github.com/foodcoops/foodsoft/blob/007ff70c486be5d93cb189fa4e13327270436e62/app/models/bank_transaction.rb#L46) to check if an invoice already has a financial link and if so, use it instead of creating a new one and overriding it. Since I cannot really test this without any bank connector, it doesn't make sense for me trying to implement it.
Also I don't know what the Rubocop error (classlength) means.

### DIFF
--- a/app/controllers/finance/balancing_controller.rb
+++ b/app/controllers/finance/balancing_controller.rb
@@ -80,7 +80,8 @@ class Finance::BalancingController < Finance::BaseController
   def close
     @order = Order.find(params[:id])
     @type = FinancialTransactionType.find_by_id(params.permit(:type)[:type])
-    @order.close!(@current_user, @type)
+    @link = FinancialLink.new if params[:create_financial_link]
+    @order.close!(@current_user, @type, @link)
     redirect_to finance_order_index_url, notice: t('finance.balancing.close.notice')
   rescue => error
     redirect_to new_finance_order_url(order_id: @order.id), alert: t('finance.balancing.close.alert', message: error.message)
--- a/app/views/finance/balancing/confirm.html.haml
+++ b/app/views/finance/balancing/confirm.html.haml
@@ -4,6 +4,11 @@
     %p
       %b= heading_helper FinancialTransaction, :financial_transaction_type
       = select_tag :type, options_for_select(FinancialTransactionType.order(:name).map { |t| [ t.name, t.id ] })
+  - if FoodsoftConfig[:use_financial_links]
+    %p
+      %label
+        = check_box_tag :create_financial_link, true, params[:create_financial_link]
+        = t('.create_financial_link')
   %table.table.table-striped{:style => "width:35em"}
     - for group_order in @order.group_orders
       %tr{:class => cycle('even', 'odd')}
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -266,13 +266,15 @@ class Order < ApplicationRecord
   end
 
   # Sets order.status to 'close' and updates all Ordergroup.account_balances
-  def close!(user, transaction_type = nil)
+  def close!(user, transaction_type = nil, financial_link = nil)
     raise I18n.t('orders.model.error_closed') if closed?
 
     update_price_of_group_orders!
 
+    invoice.update_attribute :financial_link, financial_link if invoice && financial_link
+
     transaction do                                        # Start updating account balances
-      charge_group_orders!(user, transaction_type)
+      charge_group_orders!(user, transaction_type, financial_link)
 
       if stockit?                                         # Decreases the quantity of stock_articles
         for oa in order_articles.includes(:article)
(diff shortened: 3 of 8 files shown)


paroga:
> I didn't know how to check the checkbox by default

i didn't test it, but you could use `params.fetch(:create_financial_link, true)` to provide `true` as default value

> the new config option should be set to true in case there is at least one financial link

you could add a (database) migration file for that

> I guess we also would have to update automatic assignment to invoice

i don't think that linking an order with an invoice is correct. IMHO we should implement the same schematic like in `FinancialTransactionsController`: https://github.com/foodcoops/foodsoft/blob/007ff70c486be5d93cb189fa4e13327270436e62/app/controllers/finance/financial_transactions_controller.rb#L104-L113

when balancing an order with two groups it should result in 3 transactions: e.g. _ordergroup A_ "-10€", _ordergroup B_ "-20€" and _ordergroup `nil`_ (=foodcoop) "+30€".
the invoice should usually only have corresponding `BankTransaction`s. linking between `Invoice` and `Order` happens with `invoice_id` on `Order`

> Also I don't know what the Rubocop error (classlength) means.

the order.rb files gets to big (and hard to maintain). for now you could increase the maximum in the rubocop configuration: https://github.com/foodcoops/foodsoft/blob/007ff70c486be5d93cb189fa4e13327270436e62/.rubocop_todo.yml#L286-L287


BTW: it would be great to add tests for changes on the model

twothreenine:
> i didn't test it, but you could use `params.fetch(:create_financial_link, true)` to provide `true` as default value

Wouldn't that result in creating a financial link in case the config option is deactivated and the checkbox therefore not shown?

> i don't think that linking an order with an invoice is correct. [...] when balancing an order with two groups it should result in 3 transactions: e.g. ordergroup A "-10€", ordergroup B "-20€" and ordergroup nil (=foodcoop) "+30€".

Ah, that makes sense! I'll add an option to create a foodcoop transaction. Do you think its value should be gross_amount (without margin), fc_amount (with margin) or groups_amount (the total amount the groups are actually charged)? Or perhaps gross_amount as one foodcoop transaction and, in case fc_amount varies, the difference as a another foodcoop transaction?

paroga:
> Wouldn't that result in creating a financial link in case the config option is deactivated and the checkbox therefore not shown?

depends where you put it. if you put it [in the view](https://github.com/foodcoops/foodsoft/blob/ca327f75fe649406c6df83e617d9bd7ff020523b/app/views/finance/balancing/confirm.html.haml#L10) and you don't create an `<input type=hidden>` in the "hidden case" it should work as expected. if you put it into [the controller](https://github.com/foodcoops/foodsoft/blob/ca327f75fe649406c6df83e617d9bd7ff020523b/app/controllers/finance/balancing_controller.rb#L83) it will default to `true`

> I'll add an option to create a foodcoop transaction. 

it must be the sum of the values you subtracted from the ordergroups. having a non-zero sum of a `FinanzialLink` indicates an error in your bookkeeping

app/views/finance/balancing/confirm.html.haml
```diff
@@ -4,6 +4,17 @@
     %p
       %b= heading_helper FinancialTransaction, :financial_transaction_type
       = select_tag :type, options_for_select(FinancialTransactionType.order(:name).map { |t| [ t.name, t.id ] })
+  - if FinancialTransaction.where(ordergroup: nil).any?
```
paroga, line 7: what do you think about merging both `if` into `if FoodsoftConfig[:use_financial_links]`? I think the check for an "foodcoop transaction" is somehow outdated now (applies to `new_collection.html.haml` too). I would also prefer to make both checkboxes checked by default, since that's the common case. we can do that also in a follow up commit, if you don't want to include it here.

twothreenine: I was already thinking about adding another config option "use_foodcoop_transactions", what do you think about that?
Or one option "use double-entry accounting" for both financial links and foodcoop transactions?

app/controllers/finance/balancing_controller.rb
```diff
@@ -80,7 +80,20 @@ def confirm
...
+    @link = FinancialLink.new if params[:create_financial_link]
+    @order.close!(@current_user, @type, @link)
+
+    if params[:create_foodcoop_transaction]
+      ft = FinancialTransaction.new({
+                                      financial_transaction_type: @type,
+                                      user: @current_user,
+                                      amount: @order.sum(:groups),
+                                      note: @order.transaction_note,
+                                      financial_link: @link,
```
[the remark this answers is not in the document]
twothreenine: I fixed it now, but I can't figure out why it still doesn't work (no foodcoop transaction is created)

app/views/finance/balancing/confirm.html.haml
paroga: i'd prefer to keep the number of options low. IMHO showing `user_financial_links` as _Use financial links (usable for "double-entry accounting")_ should be enough.

app/controllers/finance/balancing_controller.rb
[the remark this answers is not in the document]
paroga: `transaction_note` is a private method you cannot access here

I know, that very similar code exists somewhere else in the codebase, but the same problem applies there too (and should be fixed):

this change mixes responsibilities of controller and model, which results in hard to test code. the better way would be to add two parameters to `Order.close!` and call it then as `@order.close!(@current_user, @type, create_financial_link: params[:create_financial_link], create_foodcoop_transaction: params[:create_foodcoop_transaction])` Then it's very easy to add tests for this change.

paroga (changes requested):
could you please create an independent PR for the translation changes or force-push your changes with a clean history, so I don't need to squash them and can rebase them instead when merging. it's up to you which git-workflow you want to use

twothreenine:
replaced by #1064

PR #865, opened 2021-04-09
Fixes bundle install error

### LINKED ISSUE #864: bundle install fails due to yanked version of locked mimemagic
The command `bundle install` on a fresh install fails with the error message:

```
Your bundle is locked to mimemagic (0.3.5), but that version could not be found in any of the sources listed in your Gemfile. If you haven't changed sources, that means the
author of mimemagic (0.3.5) has removed it. You'll need to update your bundle to a version other than mimemagic (0.3.5) that hasn't been removed in order to install.
```

This is due to the `mimemagic` gem having been yanked in the version locked by `Gemfile.lock` (0.3.5):
https://rubygems.org/gems/mimemagic/versions

See also:
https://github.com/rails/rails/issues/41750#issuecomment-810808957

lentschi:
Fixes #864

This is the result of running `bundle update --conservative mimemagic`
as suggested by https://github.com/rails/rails/issues/41750#issuecomment-810808957

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -300,7 +300,9 @@ GEM
     mime-types (3.3.1)
       mime-types-data (~> 3.2015)
     mime-types-data (3.2021.0212)
-    mimemagic (0.3.5)
+    mimemagic (0.3.10)
+      nokogiri (~> 1)
+      rake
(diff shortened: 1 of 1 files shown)


wvengen:
Thank you!

PR #866, opened 2021-04-19
de.yml: Aufgaben "Verantwortliche"

mjavurek:
"Verantwortlichen" should be grammatically correct "Verantwortliche". It is used as a label for a list of persons.
aufgabe2


wvengen:
Thanks for your fix, not everyone speaks German, so it could be that inconsistencies creep in.
This ultimatly needs to be fixed in the translation system, see https://github.com/foodcoops/foodsoft/wiki/Translation, but I'll accept it here too.

PR #868, opened 2021-04-23
Issue 830

lentschi:
**What this fix does:**

1. Allow decimals on the `receive` page (@paroga's commit)
2. Convert any colon entered in a `delta_input` to a comma internally (To deal with values generated by `number_with_precision`)
3. Don't immediately post the ajax form, if the value is _still_ invalid. (E.g. if the user enters a non-numeric/negative/... value) The field is then just displayed with a red border signifying that it hasn't been posted. (Maybe we still should improve UX here...?)

### DIFF
--- a/app/assets/javascripts/delta_input.js
+++ b/app/assets/javascripts/delta_input.js
@@ -19,7 +19,7 @@ function data_delta_update(el, direction) {
   var delta = $(el).data('delta');
   var granularity = $(el).data('granularity');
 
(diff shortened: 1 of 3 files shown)


wvengen:
Depends on #853.

mortbauer:
Hello everybody, @wvengen, @lentschi  seems #853 is merged so this might be good to go?

wvengen (approved):

PR #876, opened 2021-05-25
Use docker-compose v2 since extends is still supported

decentral1se:
See https://github.com/docker/compose/issues/4315.

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: '2'
 services:
 
   foodsoft:


decentral1se:
Alternatively, we can use YAML repeat nodes to avoid the use of the deprecated key? Here is a diff that works:

```diff
diff --git a/docker-compose-dev.yml b/docker-compose-dev.yml
index ba3ec548..f9790885 100644
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -1,21 +1,15 @@
-version: '3'
+version: "3"
 services:
 
-  foodsoft:
-    extends: foodsoft_worker
-    command: ./proc-start web
-    ports:
-      - "3000:3000"
-
   foodsoft_worker:
-    build:
+    build: &build
       context: .
       dockerfile: Dockerfile-dev
     command: ./proc-start worker
-    volumes:
+    volumes: &volumes
       - bundle:/usr/local/bundle
       - .:/app
-    environment:
+    environment: &environment
       - DATABASE_URL=mysql2://root:secret@mariadb/development?encoding=utf8mb4
       - REDIS_URL=redis://redis:6379
       - TEST_DATABASE_URL=mysql2://root:secret@mariadb/test?encoding=utf8mb4
@@ -23,6 +17,14 @@ services:
       - MAILCATCHER_ADDRESS=mailcatcher
       - MAILCATCHER_PORT=25
 
+  foodsoft:
+    build: *build
+    volumes: *volumes
+    environment: *environment
+    command: ./proc-start web
+    ports:
+      - "3000:3000"
+
   mailcatcher:
     image: tophfr/mailcatcher
     ports:
```

wvengen:
Ouch, that's a welcome correction! I'm ok with 'staying' with version 2 (until we start to use version 3+ features, at which point we can see what the best way is to do this, then).

PR #877, opened 2021-05-25
Use docker-compose.yml (more convenient)

decentral1se:
It is handier. Especially if you're new to docker-compose.


wvengen:
It is true that it is not the friendliest for newcomers.

But it will be harder for non-developers wanting to use Foodsoft in production. We now have two different `Dockerfile`s, one for development and one for production, where production is the default (so people don't inadvertantly use the development version in production, which could lead to security issues). It's likely we'll get a product `docker-compose.yml` too at somepoint, the same applies here.

How do other projects handle the distinction between development and production, and the default?

decentral1se:
Yes right, fair points. I think the two approaches I've seen are `-f ...` or using `docker/{dev/prod}` folder structure. I guess it works out fine as is then and it certainly didn't stop me from working locally with the setup. I will just close this off then, thanks for considering.

wvengen:
Ok, thanks for your input!

Issue #854, opened 2021-03-08
Article Export to CSV: first column header "status" missing


mjavurek:
The first column in the exported articel CSV looks like:
`"";Order number;Name;Note;Manufacturer;Origin;Unit;Price (net);VAT;Deposit;Unit quantity;"";"";Category`
or in the German version
`"";Bestellnummer;Name;Notiz;Produzent;Herkunft;Einheit;Nettopreis;MwSt;Pfand;Gebindegröße;"";"";Kategorie`
The first column header is an empty string (""). If such a CSV is imported, everything works fine. If the CSV is opened in LibreOffice and then saved as ODS, an error occurs during the import: `A problem has occured: undefined method 'strip' for 1:Integer` 
If the word "status" is typed into cell A1 of the ODS, the import works fine.
So I think the exported CSV should contain a header for column 1 for compatibility with other file formats.

mjavurek:
Thanks for fixing it so fast. 
I then noticed that the status data itself is also not exported.

kidhab:
> I then noticed that the status data itself is also not exported.

I assume this is by design. From the article upload view:

>The rows shown here are examples. When there is an "x" in the first column, the article is outlisted and will be removed. This allows you to edit the spreadsheet and quickly remove many articles at once, for example when articles become unavailable with the supplier.

If the first column will be included in the export the column will be filled with `frue` or `false` statements.

PR #884, opened 2021-06-09
fill availability column + add translation

vurtn:
ref: #854

The availability column is now filled and the button to export the articles is translated in french.

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -484,7 +484,7 @@ de:
       already_imported: schon importiert
       not_found: Keine Artikel gefunden
     index:
-      change_supplier: Lieferant wechseln ...
+      change_supplier: Lieferant wechseln
       download: Artikel herunterladen
       edit_all: Alle bearbeiten
       ext_db:
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -348,6 +348,7 @@ fr:
       not_found: Aucun produit correspondant n'a été trouvé
     index:
       change_supplier: Changer de fournisseur-e...
+      download: Exporter les produits
       edit_all: Tout modifier
       ext_db:
         import: Rechercher/Importer
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -347,7 +347,7 @@ fr:
       already_imported: déjà importé
       not_found: Aucun produit correspondant n'a été trouvé
     index:
-      change_supplier: Changer de fournisseur-e...
+      change_supplier: Changer de fournisseur-e
       download: Exporter les produits
       edit_all: Tout modifier
       ext_db:
--- a/lib/articles_csv.rb
+++ b/lib/articles_csv.rb
@@ -23,7 +23,7 @@ class ArticlesCsv < RenderCSV
   def data
     @object.each do |o|
       yield [
-        '',
+        o.availability ? I18n.t('simple_form.yes') : I18n.t('simple_form.no'),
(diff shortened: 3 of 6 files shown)


config/locales/fr.yml
```diff
@@ -348,6 +348,7 @@ fr:
       not_found: Aucun produit correspondant n'a été trouvé
     index:
       change_supplier: Changer de fournisseur-e...
+      download: Exporter les produits
```
decentral1se, line 351: Would you mind cleaning up the `...` from the other translations?

```
config/locales/nl.yml:      change_supplier: Leverancier wisselen ...
config/locales/en.yml:      change_supplier: Change supplier ...
config/locales/es.yml:      change_supplier: Cambiar proveedor ...
config/locales/de.yml:      change_supplier: Lieferant wechseln ...
config/locales/fr.yml:      change_supplier: Changer de fournisseur-e...
```

decentral1se (commented):
:rocket:

the-star-rover:
`...` removed in change_supplier.

lib/articles_csv.rb
```diff
@@ -23,7 +23,7 @@ def header
   def data
     @object.each do |o|
       yield [
-        '',
+        o.availability ? I18n.t('simple_form.yes') : I18n.t('simple_form.no'),
```
wvengen, line 26: It would be good if the import would recognize this (I vaguely recall `x` or empty) - did you check it corresponds?

config/locales/de.yml
```diff
@@ -484,7 +484,7 @@ de:
       already_imported: schon importiert
       not_found: Keine Artikel gefunden
     index:
-      change_supplier: Lieferant wechseln ...
+      change_supplier: Lieferant wechseln
```
wvengen, line 487: What about a caret after the text in the page? Or is this for something else?

lib/articles_csv.rb
twothreenine: > I vaguely recall `x` or empty

Not quite, `x` removes/deletes the article altogether.
I opened #949 to propose the addition of another key to import articles as `not available`. We should then use the same key in the exported CSV.

kidhab: There's anothe approach at the article import:
> When there is an "x" in the first column, the article is outlisted and will be removed.

So one can use the first column to mass delete articles of a supplier. The import doesn't process any available column.

kidhab:
@vurtn Can you split this MR into different ones: one for the export and one for the updates to the languages files?

I think after that we can merge it because it's already an improvement and it doesn't interfere with the article import (see my comment above).

lib/articles_csv.rb
twothreenine: @kidhab I see you merged it now although the import behavior does not correspond (yet).
I also see an issue with the translated yes/no: This would mean the import would have to check all yes/no locales for each availability cell (or else the imported CSV would have to be in the user's locale, which could be a problem if one user exports it & another one imports it, and also for Lieferscraping).
Did you consider my proposal #949 ?

kidhab: At the moment the behaviour of the article's import and export functionality differs because they fulfill another purpose. This MR add missing feature based on the current purpose of the export functionality. And from my point of view it shouldn't take years to implement it.

Your proposal addresses a slightly different aspect of the Foodsoft. I think this is something that should be implemented. But this shouldn't block such small changes.

PR #887, opened 2021-09-09
Add Rails 6 backport for ActiveRecord

paroga:
This fixes the "can't create Thread: Resource temporarily unavailable" error.


kidhab:
We applied this at the global hosting some weeks ago and didn't seen any of the described errors since then. What about a merge?

wvengen:
There are some small linter issues. Is someone willing to fix this, or shall I merge because it is really important for production?

wvengen:
Will we get a linter issue with every PR when we don't resolve the above? I'm a little hesitant to merge.

PR #901, opened 2021-12-09
Enable to configure feedback recipient address

kidhab:
Enable to set a different address for mails send through a Foodsoft's feedback form. Make it possible to send error notifications to an admin group and feedback/ support requests to a support mailing list.

### DIFF
--- a/app/mailers/mailer.rb
+++ b/app/mailers/mailer.rb
@@ -103,7 +103,7 @@ class Mailer < ActionMailer::Base
     @user = user
     @feedback = feedback
 
-    mail to: FoodsoftConfig[:notification][:error_recipients],
+    mail to: feedback_recipients,
          from: user,
          subject: I18n.t('mailer.feedback.subject')
   end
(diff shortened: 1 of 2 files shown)


kidhab:
Usually a clear note in the update instructions should be enough to point to this change.

What about this:
```
unless defined? feedback_recipients
    feedback_recipients = error_recipients
end
```
or
```
feedback_recipients = error_recipients if feedback_recipients.nil?
```

I'm not familar with Ruby. Your feedback is appreciated.

app/mailers/mailer.rb
```diff
@@ -103,7 +103,7 @@ def feedback(user, feedback)
     @user = user
     @feedback = feedback
 
-    mail to: FoodsoftConfig[:notification][:error_recipients],
+    mail to: FoodsoftConfig[:notification][:feedback_recipients],
```
wvengen, line 106: Ideally this would go in [FoodsoftConfig#get_default_config](https://github.com/foodcoops/foodsoft/blob/master/lib/foodsoft_config.rb#L254). But as the fallback is a value depending on another configuration item, it won't work.

What about:
```ruby
def feedback(user, feedback)
  #...
  mail to: feedback_recipients,
  # ...
end

private

def feedback_recipients
  # use the (new) feedback_recipients option, but fallback to error_recipients for backwards compatibility
  FoodsoftConfig[:notification][:feedback_recipients] || FoodsoftConfig[:notification][:error_recipients]
end
```
```

kidhab: Thanks for your feedback. I added the priavte method and squashed the commit.

wvengen (changes requested):
Great! Ready to merge if indentation is solved :)

kidhab:
It is solved now. Thanks for the hint.

wvengen (approved):
Thanks!

PR #902, opened 2021-12-12
Update SETUP_DEVELOPMENT.md

mjavurek:
How to avoid errors arising during step 4. 

If /usr/bin/mkdir is not available, the compilation of nokogiri fails:
```
make: /usr/bin/mkdir: Kommando nicht gefunden
make: *** [Makefile:202: .sitearchdir.-.nokogiri.time] Fehler 127
make install failed, exit code 2

```
see also: 
https://stackoverflow.com/questions/64653051/make-usr-bin-mkdir-command-not-found-during-gem-install-nokogiri-in-ubuntu

### DIFF
--- a/doc/SETUP_DEVELOPMENT.md
+++ b/doc/SETUP_DEVELOPMENT.md
@@ -69,7 +69,18 @@ If instead you just want to run Foodsoft without changing its code, please refer
         # CentOS/Redhat
         sudo yum install v8 community-mysql-devel libxml2-devel libxslt-devel libffi-devel readline-devel file-devel
 
-3. Install Ruby dependencies:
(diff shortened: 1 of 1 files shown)


wvengen:
Ouch ... could this be fixed in nokogori instead?

RayOei:
This seems related to specific versions issues with (older) Ubuntu and/or Ruby. See additional [comment](https://stackoverflow.com/questions/64653051/make-usr-bin-mkdir-command-not-found-during-gem-install-nokogiri-in-ubuntu#comment120916443_64653051)

With recent installs this didn't seem to be a problem and `mkdir` _should_ normally be available anyway. Adding it seems a bit far fetched.

PR #905, opened 2021-12-20
Update to recent version

kidhab:


wvengen:
That sounds like a good idea.
The `Dockerfile`s already use `2.6` so don't need updating.
Does `.ruby-version` support omitting the patch level?

kidhab:
The file's contents only representing a string of the Ruby version to use with this project. I assume it's usualy the first place to look at if you want to know what Ruby version a project uses. Also (and maybe more importantly) RVM and Rbenv will automatically detect it and set the correct Ruby version.
So yes, it should be save to omit a version as long as the new one is working with the Foodsoft.

wvengen:
So, anything against `2.6`?

kidhab:
I think I missunderstood you. Do you ask if it's save to remove the patch level from the `.ruby-version` file, don't you?
In order to work like expected the version has to be complete including the patch level.

E.g. `rbenv` parses the file and uses precisely that included version for running the application.

wvengen:
> In order to work like expected the version has to be complete including the patch level.

Thanks, that's what I wanted to check!

PR #907, opened 2021-12-23
Feature/tw automatic group order invoice generation

Viehlieb:
Offers the possibility to generate Invoices for given Group Orders.
Generation is possible after an order is closed.
If order is 'settled' then there's additionally the possibility to let the software send e-mails automatically on button press. The corresponding invoices will be sent as attachment via mail to the order group members.

Notes:
To use automatic sending of invoices, select option in "Admin/Finance" config
You can enter a "payment method " in "Admin/Finance"
To make things work, it is mandatory to enter a tax_number in "Admin/Foodcoop" config

As for now "payment method" determines the "body" of the invoice. 
An example of new options and the resulting invoice can be found here:

tax_number
finance
invoice

Additionally for older and already settled orders, group order invoices can be generated manually under "finance/balancing"
generate
download

Invoices will be generated given the momentarily given information about invoicee and invoicer (tax_number, contact details...). SO take care when delting an invoice.

### DIFF
--- a/app/controllers/group_order_invoices_controller.rb
+++ b/app/controllers/group_order_invoices_controller.rb
@@ -1,0 +1,38 @@
+class GroupOrderInvoicesController < ApplicationController
+  include Concerns::SendGroupOrderInvoicePdf
+
+  def show
+    @group_order_invoice = GroupOrderInvoice.find(params[:id])
+    if FoodsoftConfig[:contact][:tax_number]
+      respond_to do |format|
+        format.pdf do
+          send_group_order_invoice_pdf @group_order_invoice if FoodsoftConfig[:contact][:tax_number]
+        end
+      end
+    else raise RecordInvalid
+      redirect_back fallback_location: root_path, notice: 'Something went wrong', :alert => I18n.t('errors.general_msg', :msg => "#{error} " + I18n.t('errors.check_tax_number'))
+    end
+  end
+
+  def destroy
+    goi = GroupOrderInvoice.find(params[:id])
+    @order = goi.group_order.order
+    goi.destroy
+    respond_to do |format|
+      format.js
+      format.json { head :no_content }
+    end
+  end
+
+  def create
+    go = GroupOrder.find(params[:group_order])
+    @order = go.order
+    goi = GroupOrderInvoice.find_or_create_by!(group_order_id: go.id)
+    respond_to do |format|
+      format.js
+    end
+    redirect_back fallback_location: root_path
+  rescue => error
+    redirect_back fallback_location: root_path, notice: 'Something went wrong', :alert => I18n.t('errors.general_msg', :msg => error)
+  end
+end
--- a/app/models/group_order_invoice.rb
+++ b/app/models/group_order_invoice.rb
@@ -1,0 +1,60 @@
+class GroupOrderInvoice < ApplicationRecord
+
+  belongs_to :group_order
+  validates_presence_of :group_order
+  validates_uniqueness_of :invoice_number
+  validate :tax_number_set
+  after_initialize :init, unless: :persisted?
+
+  def generate_invoice_number(count)
+    trailing_number = count.to_s.rjust(4, '0')
+    unless  GroupOrderInvoice.find_by(invoice_number: Time.now.strftime("%Y%m%d") + trailing_number)
+      Time.now.strftime("%Y%m%d") + trailing_number
+    else
+      generate_invoice_number(count.to_i + 1)
+    end
+  end
+
+  def tax_number_set
+    if FoodsoftConfig[:contact][:tax_number].blank?
+      errors.add(:group_order_invoice, "Keine Steuernummer in FoodsoftConfig :contact gesetzt")
+    end
+  end
+
+  def init
+    self.invoice_number = generate_invoice_number(1) unless self.invoice_number
+    self.invoice_date = Time.now unless self.invoice_date
+    self.payment_method = FoodsoftConfig[:group_order_invoices]&.[](:payment_method) || I18n.t('activerecord.attributes.group_order_invoice.payment_method') unless self.payment_method
+  end
+
+  def name
+    I18n.t('activerecord.attributes.group_order_invoice.name') + "_#{invoice_number}"
+  end
+
+  def load_data_for_invoice
+    invoice_data = {}
+    order = group_order.order
+    invoice_data[:supplier] = order.supplier.name
+    invoice_data[:ordergroup] = group_order.ordergroup
+    invoice_data[:group_order] = group_order
+    invoice_data[:invoice_number] = invoice_number
+    invoice_data[:invoice_date] = invoice_date
+    invoice_data[:tax_number] = FoodsoftConfig[:contact][:tax_number]
+    invoice_data[:payment_method] = payment_method
+    invoice_data[:order_articles] = {}
+    group_order.order_articles.each do |order_article|
+      # Get the result of last time ordering, if possible
+      goa = group_order.group_order_articles.detect { |goa| goa.order_article_id == order_article.id }
+
+      # Build hash with relevant data
+      invoice_data[:order_articles][order_article.id] = {
+        :price => order_article.article.fc_price,
+        :quantity => (goa ? goa.quantity : 0),
+        :total_price => (goa ? goa.total_price : 0),
+        :tax => order_article.article.tax
+      }
+    end
+    invoice_data
+
+  end
+end
--- a/app/controllers/concerns/send_group_order_invoice_pdf.rb
+++ b/app/controllers/concerns/send_group_order_invoice_pdf.rb
@@ -1,0 +1,12 @@
+module Concerns::SendGroupOrderInvoicePdf
+  extend ActiveSupport::Concern
+
+  protected
+
+  def send_group_order_invoice_pdf group_order_invoice
+    invoice_data = group_order_invoice.load_data_for_invoice
+    invoice_data[:title] = "Rechnung für " + invoice_data[:supplier] # TODO Internationalise und load_data rausschmeißen
+    pdf = GroupOrderInvoicePdf.new group_order_invoice.load_data_for_invoice
+    send_data pdf.to_pdf, filename: pdf.filename, type: 'application/pdf'
+  end
+end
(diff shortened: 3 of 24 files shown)


wvengen:
This seems like a great feature. Thanks for taking the time to draft this!
What exactly is the use of removing an invoice?

Viehlieb:
> 
Hey. Excuse my late reply.
When incorrect settings are set in the admin configs, there should be the possibility to delete the old and generate a new invoice with correct settings applied.

Viehlieb (commented):
If I provide a link to article details it was not possible to click the link because of the :hover feature. 
Behavior before commit:
Hover over articles would render the article infos and if no article is hovered there's a green box showing
Behavior after commit:
Hover over articles will render article infos asbefore.
Click article, and hover over other articleswill render article infos as before.
BUT the 'clicked' info will be the 'fallback info' which will remain in the green box (article clicked last). So it is possible to click alink in the green bo
Screenshot from 2022-04-05 13-24-25
x

Viehlieb:
Is there anything more needed to merge this feature into master branch?

paroga:
first: sorry for the late reply
second: thanks for the contribution
third: this is a huge change and therefore hard to review: could split it smaller self-containing parts? e.g.: change the model in a separate commit

there are a bunch of style violations and i have no idea why they are not marked by the CI system :disappointed:
if i'm not wrong there are a bunch of changes which require cleanup before merge. it seams there a some dummy "hello world" files and debug outputs introduces.
please feel free to correct my if i'm wrong. i just skimmed the PR

Viehlieb:
Hey. Thanks for the hints. 
It made me look into the code again and I am working on fixing the points you mentioned.

1) I noticed, there was no authentication clause to deal wirth group order invoices.
With b6854bc41609fe42bcb5bae733f9ffc4524d3ea6 I added a neccessary update of the group_order_invoice_controller.
:authenticate_finance, so only for finance authenticated users can see/download and generate group_order_invoices.
2) Removed the "Hello World" Dummy files.
3) Todo Style issues
Let me know, if there is anything more you notice.

app/controllers/group_order_invoices_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  def show
+    @group_order_invoice = GroupOrderInvoice.find(params[:id])
+    if FoodsoftConfig[:contact][:tax_number]
+      respond_to do |format|
+        format.pdf do
+          send_group_order_invoice_pdf @group_order_invoice if FoodsoftConfig[:contact][:tax_number]
+        end
+      end
+    else raise RecordInvalid
+      redirect_back fallback_location: root_path, notice: 'Something went wrong', :alert => I18n.t('errors.general_msg', :msg => "#{error} " + I18n.t('errors.check_tax_number'))
```
paroga: * what kind of error do you want to catch here?
* please use translations for user visible strings.
* please do not use `notice` and `alert` at the same time
* there is usually no need to use `I18n.t` and you could use just `t` in the controllers
* please use newer syntax `name: value` instead of `:name => value`

app/controllers/group_order_invoices_controller.rb
```diff
@@ -0,0 +1,38 @@
+class GroupOrderInvoicesController < ApplicationController
+  include Concerns::SendGroupOrderInvoicePdf
+
+  def show
+    @group_order_invoice = GroupOrderInvoice.find(params[:id])
+    if FoodsoftConfig[:contact][:tax_number]
+      respond_to do |format|
+        format.pdf do
+          send_group_order_invoice_pdf @group_order_invoice if FoodsoftConfig[:contact][:tax_number]
```
paroga: you check already for the `tax_number` in line 6. what's the purpose of the second check here?

app/controllers/group_order_invoices_controller.rb
```diff
@@ -0,0 +1,38 @@
+class GroupOrderInvoicesController < ApplicationController
+  include Concerns::SendGroupOrderInvoicePdf
+
+  def show
+    @group_order_invoice = GroupOrderInvoice.find(params[:id])
+    if FoodsoftConfig[:contact][:tax_number]
```
paroga: why can't we show an invoice without an tax number?

app/controllers/group_order_invoices_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+        end
+      end
+    else raise RecordInvalid
+      redirect_back fallback_location: root_path, notice: 'Something went wrong', :alert => I18n.t('errors.general_msg', :msg => "#{error} " + I18n.t('errors.check_tax_number'))
+    end
+  end
+
+  def destroy
+    goi = GroupOrderInvoice.find(params[:id])
+    @order = goi.group_order.order
```
paroga, line 24: what's the purpose of `@order` here?

app/controllers/group_order_invoices_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    goi.destroy
+    respond_to do |format|
+      format.js
+      format.json { head :no_content }
+    end
+  end
+
+  def create
+    go = GroupOrder.find(params[:group_order])
+    @order = go.order
```
paroga, line 50: what's the purpose of @order here?

app/controllers/group_order_invoices_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    goi = GroupOrderInvoice.find_or_create_by!(group_order_id: go.id)
```
paroga: why `find_or_create` in the `create` method?

app/controllers/group_order_invoices_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    respond_to do |format|
+      format.js
+    end
+    redirect_back fallback_location: root_path
```
paroga, line 55: why do you redirect here to `root_path`?

app/controllers/group_order_invoices_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  rescue => error
+    redirect_back fallback_location: root_path, notice: 'Something went wrong', :alert => I18n.t('errors.general_msg', :msg => error)
```
paroga, line 57: the comments form line 13 apply here too

app/models/group_order_invoice.rb
```diff
@@ -0,0 +1,60 @@
...
+  belongs_to :group_order
+  validates_presence_of :group_order
+  validates_uniqueness_of :invoice_number
+  validate :tax_number_set
+  after_initialize :init, unless: :persisted?
+
+  def generate_invoice_number(count)
+    trailing_number = count.to_s.rjust(4, '0')
+    unless  GroupOrderInvoice.find_by(invoice_number: Time.now.strftime("%Y%m%d") + trailing_number)
+      Time.now.strftime("%Y%m%d") + trailing_number
```
paroga: * is there a way to make the format configureable?
* please create a variable for values you use multiple times

app/models/group_order_invoice.rb
```diff
@@ -0,0 +1,60 @@
...
+    else
+      generate_invoice_number(count.to_i + 1)
```
paroga, line 11: * recursive calling `generate_invoice_number` here is a very bad idea, since you create a database query all the time which is slow and blocks everything else
* you could either store the invoice number somewhere in the database or find the last invoice via a single query to the database (something like `SELECT invoice_number ... WHER invoice_number LIKE '...%' ORDER BY invoice_number DESC LIMIT 1`)

app/models/group_order_invoice.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+  end
+
+  def tax_number_set
+    if FoodsoftConfig[:contact][:tax_number].blank?
+      errors.add(:group_order_invoice, "Keine Steuernummer in FoodsoftConfig :contact gesetzt")
+    end
```
paroga, line 20: this method does not access the database and therefore should be removed from the model and move to a better location (BTW: there are user visible string which are not translated)

app/models/group_order_invoice.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def init
+    self.invoice_number = generate_invoice_number(1) unless self.invoice_number
+    self.invoice_date = Time.now unless self.invoice_date
+    self.payment_method = FoodsoftConfig[:group_order_invoices]&.[](:payment_method) || I18n.t('activerecord.attributes.group_order_invoice.payment_method') unless self.payment_method
```
paroga, line 26: why is the payment method stored in `FoodsoftConfig`?
please do not use translated strings in the model. depending on the locale of the current user different data would be used and stored in the database

app/models/group_order_invoice.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def tax_number_set
+    if FoodsoftConfig[:contact][:tax_number].blank?
+      errors.add(:group_order_invoice, "Keine Steuernummer in FoodsoftConfig :contact gesetzt")
+    end
+  end
+
+  def init
+    self.invoice_number = generate_invoice_number(1) unless self.invoice_number
```
paroga, line 25: please use a `before_validation` hook instead (see e.g. [`set_password` in the `User` model](https://github.com/foodcoops/foodsoft/blob/f7f8774c433d58bcdf0c03fe93a7967fcf0e6437/app/models/user.rb#L122-L127))

app/models/group_order_invoice.rb
```diff
@@ -0,0 +1,60 @@
...
+    self.invoice_date = Time.now unless self.invoice_date
+    self.payment_method = FoodsoftConfig[:group_order_invoices]&.[](:payment_method) || I18n.t('activerecord.attributes.group_order_invoice.payment_method') unless self.payment_method
+  end
+
+  def name
+    I18n.t('activerecord.attributes.group_order_invoice.name') + "_#{invoice_number}"
```
paroga, line 30: the translation should be moved from the model to a controller, since the model should be language independent

app/models/group_order_invoice.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def load_data_for_invoice
```
paroga, line 33: why do you copy everything here? usually working with a `GroupOrderInvoice` instance should be the way to go.

paroga (changes requested):
i looked only into the first commit of your PR for now

RayOei:
As far as I can make out, no follow-up exists. 
I do not know the actual use case but I am inclined to close this in its current state?
_Added tags for review._

PR #910, opened 2022-01-09
Change instructions to rbenv

### LINKED ISSUE #899: rvm vs. rbenv
The [instructions](https://github.com/foodcoops/foodsoft/blob/master/doc/SETUP_DEVELOPMENT.md) recommend `rvm`. On debian, I prefer [`rbenv`](https://packages.debian.org/bullseye/rbenv) because I can install it with `apt`.

Is there any reason to prefer `rvm` over `rbenv`?

kidhab:
I made the following changes:
- drop all rvm commands in favor of rbenv
- remove dependency on Postgresql as this document doesn't cover it any further
- remove  links to dependencies of the OS (as some of them are outdated and it means less maintenance)
- change `rake` commands to `rails`

Closes #899


wvengen (approved):
Thanks for clearing this up and moving to `rbenv`!

PR #912, opened 2022-01-18
Add default time_zone

### LINKED ISSUE #908: Open orders are not displayed if time zone is missing or behind
I'm not sure if this is a bug. But maby we can enter some useful defaults to a new database (by using seeds).

**Steps to reproduce**
- Set the time zone in the Foodsoft settings to null (= no entry for `foodcoop.demo.time_zone` in the database) or to a time zone that's behind the browser's time zone
- Create a new order - don't change the start date (the current date and time are preselected)
- Open the dashboard or the group_orders overview - the new order is missing

I assume this is the associated database query:
```
SELECT  1 AS one FROM `orders` WHERE `orders`.`state` = 'open' AND (starts <= '2022-01-02 13:52:39.899261') LIMIT 1
```
Because of the different time zones the newly created order will not be shown.

kidhab:
There is no default time zone setting in newly created Foodsoft instances. Closes #908

### DIFF
--- a/lib/foodsoft_config.rb
+++ b/lib/foodsoft_config.rb
@@ -257,6 +257,7 @@ class FoodsoftConfig
         use_apple_points: true,
         # English is the default language, and this makes it show up as default.
         default_locale: 'en',
+        time_zone: 'Berlin',
         currency_unit: '€',
         currency_space: true,
         foodsoft_url: 'https://github.com/foodcoops/foodsoft',


wvengen:
Thanks!
What is a sensible default value? English language suggests London or GMT.
But currently most users are in Western-Europe, so Berlin/Paris/Amsterdam/CET/CEST makes sense.
I'm fine with Berlin.

PR #913, opened 2022-01-18
Add ability to include custom gems

kidhab:
This might be useful if you want to include gems that are not part of this project. E.g. when a Foodsoft hosting provider want to include [turnout](https://github.com/biola/turnout) to inform about maintenance.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -117,3 +117,7 @@ group :test do
   gem 'apivore', require: false
   gem 'hashie', '~> 3.4.6', require: false # https://github.com/westfieldlabs/apivore/issues/114
 end
+
+# Use a local Gemfile to include custom dependencies that might be relevant for
+# hosting environments or developement.
+eval_gemfile 'Gemfile.local' if File.exist? 'Gemfile.local'


wvengen:
Interesting idea.
`Gemfile.lock` remains an issue here, how to handle that? Can we separate the two ... ?

kidhab:
I would leave the `Gemfile.lock` as it is. It's up to the user to update that file if one adds additional gems to `Gemfile.local`.

wvengen:
When people start using this, `Gemfile.lock` would be updated after a `bundle install`.
I'd expect this change to introduce random `Gemfile.lock` changes in PRs.

Are there drawbacks to `echo "\ngem 'foo'" >>Gemfile && bundle install` for a hosting provider?

paroga:
I do not think this is a clean solution, since you would need to keep it in sync with `Gemfile.lock`. IMO hosting providers should install only with `bundle install --deployment`, which requires that [`Gemfile.lock` is up to date](https://bundler.io/man/bundle-install.1.html#DEPLOYMENT-MODE)

kidhab:
Maybe it would be better to do it the other way and include the `Gemfile` in a `Gemfile.local`:
```
# Gemfile.local
eval File.read('Gemfile')
```
And then run `bundle install --gemfile Gemfile.local`

PR #916, opened 2022-01-24
One-click development environment via Gitpod.io (browser-based VSCode with cloud backend)

Mob-Programming-Meetup-Austria:
Hi,

during our last mob-programming meetup session we were experimenting with this project and prepared a configuration for creating a development environment on https://gitpod.io.

The instructions in `doc/SETUP_DEVELOPMENT_DOCKER.md` were very helpful for figuring out how to use docker-compose in the `.gitpod.yml` for preparing the docker containers.

You can test the idea with the badge we added to the `README.md`. If you like it and also want to keep the badge you just need to update the URL to https://gitpod.io/#https://github.com/foodsoft/foodsoft.

Thanks to all contributors during our mob-programming meetup session:

* @bjelline
* @dpolivaev
* @haraldreingruber 
* @koushik-ms

It was a really awesome group-effort 🙏😊


wvengen:
This is quite cool, thanks for the work!
Could we move the instructions for running with gitpod to a separate document, like the other approaches? That keeps the README clean. If all developers are switching to using gitpod, we may make it more prominent.
We could consider keeping the button, though, so people recognize it directly.

Just played around with it, found some things that still could be improved:
- in a console: `rake -T` errored with "couldn't find bundler"
- in a console: `bundle exec rake -T` complained about a wrong bundler version
- it would be nice if the Foodsoft page is shown by default (instead of mailcatcher)

haraldreingruber:
Thanks for the fast feedback and for trying out the solution.

Regarding the README, it was of course just an idea. Moving it to a separate page makes totally sense.

> We could consider keeping the button, though, so people recognize it directly.

So you mean keeping the button where it is but moving all the other instructions to a new page?
Would you keep the button in this section? Maybe you can quickly sketch your idea with a small markdown snippet?

I will try to look into your improvement requests next weekend.

haraldreingruber:
> in a console: rake -T errored with "couldn't find bundler"
> in a console: bundle exec rake -T complained about a wrong bundler version

My first idea was to use `ruby:2.6` as the base image for Gitpod, but then docker-compose is missing. For now, I think using the default Gitpod base image and installing the right versions is a better idea.

> it would be nice if the Foodsoft page is shown by default (instead of mailcatcher)

I've changed mailcatcher and phpmyadmin to notify only. Looks like open preview on startup only works just for a single port.

Looking forward to your feedback and further improvement ideas!

.gitpod.yml
```diff
@@ -0,0 +1,34 @@
+# This configuration file was mainly inspired by ./doc/SETUP_DEVELOPMENT_DOCKER.md
+
+tasks:
+  - name: Rails
+    init: |
+      # install correct ruby+tools versions
+      rvm install 2.6.9 # should be similar as in /Dockerfile-dev 
+      rvm use 2.6.9
```
wvengen, line 8: you could perhaps use `rvm install $(cat .ruby-version)` and `rvm use $(cat .ruby-version)`

.gitpod.yml
```diff
@@ -0,0 +1,34 @@
...
+      gem install bundler:1.17.3
```
wvengen, line 9: Would `gem install bundler` suffice - or is there a dependency on the specific version? I'd hope to make Ruby version upgrades somewhat smooth in this way.

wvengen (commented):
Thanks for the changes! Added one idea.

.gitpod.yml
haraldreingruber: Yep, these are all great suggestions. I've integrated them in the current version of the PR.

haraldreingruber:
@wvengen I think we were able to address the mentioned improvement ideas.

I am not sure if we could also avoid running `bundle install` in the main workspace, as it is also done in the foodsoft docker container, and it could be confusing that different ones are used for the running app and the workspace. 
But I could imagine it might be helpful during development and for running rake tasks.
The alternative could be adding instructions how to open a shell inside the docker container... but I am not enough familiar with a Rails+Docker workflow to know which would be the preferable option.

haraldreingruber:
To be honest, I am still trying to figure out the best setup for this. Also, some things are not working smoothly when a workspace is paused and restarted (probably has to do with which folders and docker container state is persisted by Gitpod).

Executing the rake commands via `docker-compose -f docker-compose-dev.yml run --rm foodsoft ...` (similar as described in the SETUP_DEVELOPMENT_DOCKER.md) or inside the foodsoft docker container directly works well so far.

Maybe this workflow is more convenient compared to installing ruby and all gems in the main gitpod environment too.

One way or the other, I think it might be a good idea to clearly state that the Gitpod setup is still experimental and ask for improvement suggestions?

haraldreingruber:
After using this setup for implementing #716 I think the simplest solution is not to install ruby and all bundles in the gitpod environment directly, but instead give a hint how to open up a terminal inside the foodsoft container, where everything is installed already and all the commands can be used as expected.

This avoids having to install everything twice, once in the docker environment and once in Gitpod.

This can be tested here:
https://gitpod.io/#https://github.com/mob-programming-meetup/foodsoft

I've pushed the other approaches to branches for easier comparison.
* Installing ruby and bundles via `gitpod.Dockerfile`: https://github.com/mob-programming-meetup/foodsoft/tree/feature/gitpod-dockerfile
* Installing ruby and bundles via Gitpod init-tasks: https://github.com/mob-programming-meetup/foodsoft/tree/feature/gitpod-init-tasks

@wvengen @paroga What do you think?

haraldreingruber:
This PR doesn't change any internals of the project. Is anything still blocking the merge?

paroga:
i didn't check in detail, but since it does not touch any existing code and there a no objections from anybody else i just merged it

paroga:
Shouldn't the URL in the README change from `https://github.com/mob-programming-meetup/foodsoft` to `https://github.com/foodcoops/foodsoft`?

haraldreingruber:
Ah yes, good catch.

It was pointing to our fork, so that the .gitpod.yml is picked up when opening the link.
Now that it's merged to master, it should of course point to the main repo.

Will submit a new PR.

PR #917, opened 2022-01-24
Add validation of more article fields

kidhab:
If one adds a some text  that is longer than 255 characters into these articles fields the browser shows an unspecific error messages. With this validation an understandable message will be shown.


kidhab:
Let me know if you prefer grouping of the fields like:
```Ruby
validates_length_of :note, :origin, :manufacturer, :order_number, :maximum => 255
```

wvengen:
With migrating to higher Rails versions, I'd consider moving the use of `validates`, one line per attribute.

PR #925, opened 2022-02-13
Update midi-smtp-server because v2.3.2 was yanked

haraldreingruber:
Looks like the `midi-smtp-server` version referenced in the Gemfile.lock was yanked.

Is there any reason not updating to the latest version would be preferable? I guess updating to v2.3.3 would be an alternative.


paroga:
`bundle update` in bca29bea705eb4759c4a9fe87ca24b04625551b2

PR #927, opened 2022-02-14
fix translation missing: xxx.orders.state.received

mjavurek:
Added missing translation for de.orders.state.received.
grafik


paroga:
can you add the lines to the other languages too?

mjavurek:
> can you add the lines to the other languages too?

done.

mjavurek:
Thanks!

PR #929, opened 2022-02-20
Configure base_controller of Doorkeeper

paroga:


paroga:
@wvengen do you know why you didn't use `base_controller` to configure `doorkeeper` in fd96b6ccc10d3fcbf973442182c6b575ee577adf?

wvengen:
@paroga Actually, no. Your approach seems a lot cleaner (when it keeps working well)!

PR #938, opened 2022-04-09
Fix Typo

kidhab:


wvengen:
I see it was also translated in the translation system (Crowdin) :+1:

PR #954, opened 2022-09-20
Spelfouten, maar ook verbeteringen in Nederlands

hamaryns:
Hi all,

I noticed some severe spelling errors in Dutch (betaald should be betaalt).  Furthermore suggestions for better translations, more in line with localisation style in other projects: verb at the end in commands etc.

If accepted, I will continue on this file.  If you want me to do this in another manner, please say so.

Cheers, H.

### DIFF
--- a/config/locales/nl.yml
+++ b/config/locales/nl.yml
@@ -370,22 +370,22 @@ nl:
         new_user: Nieuwe gebruiker
         new_users: toevoegen
         show_deleted: Verwijderde gebruikers tonen
-        title: Gebruikers admin
+        title: Gebruikersbeheer
       new:
         title: Nieuwe gebruiker toevoegen
       restore:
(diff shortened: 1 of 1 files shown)


wvengen (approved):
This looks like an improvement, thank you!

Note that it's important to look at the translated strings in the application too, because it may fit or not. These changes are just tweaks, so they are fine. :+1:

wvengen:
We have also been using crowdin for translations. It used to be the case that that is the preferred place to work on translations. Did you look at that as well? https://github.com/foodcoops/foodsoft/wiki/Translation

hamaryns:
> We have also been using crowdin for translations. It used to be the case that that is the preferred place to work on translations. Did you look at that as well? https://github.com/foodcoops/foodsoft/wiki/Translation

Ah, I hadn’t seen that, will use from now on.

PR #957, opened 2022-10-13
Show instance name at login screen

kidhab:
This commit is useful for multi_coop_installs. At the moment the login screen of every multi_coop instance looks the same.
It can happen that users log in to another instance because they mistyped the URL.
With this change to name of a foodsoft instance will be shown as the title of the login screen.

### DIFF
--- a/app/views/sessions/new.html.haml
+++ b/app/views/sessions/new.html.haml
@@ -6,6 +6,8 @@
 
 - title t('.title')
 
+.lead= FoodsoftConfig[:name]
+
 %noscript
   .alert.alert-error
     != t '.nojs', link: link_to(t('.noscript'), "http://noscript.net/")


app/views/sessions/new.html.haml
```diff
@@ -4,7 +4,10 @@
       $('#nick').focus();
     });
 
-- title t('.title')
+- if FoodsoftConfig[:instance_name].present?
+  - title FoodsoftConfig[:instance_name]
```
paroga, line 8: why not `FoodsoftConfig[:instance_name] || t('.title')` to avoid duplication of the config access?

what's the benefit of `instance_name` over the already existing `name`?

config/app_config.yml.SAMPLE
```diff
@@ -167,6 +167,14 @@ default: &defaults
   # default to allow automatically adding new articles on sync only when less than 200 articles in total
   #shared_supplier_article_sync_limit: 200
 
+# Multi_coop example configuration
+# In order to make this work set "multi_coop_install: true"
+#anotherfoodcoop:
+#  <<: *defaults
+#  database:
+#    database: foodsoft_anotherfoodcoop
+#  instance_name: anotherfoodcoop
```
paroga, line 176: please move it to the other options like `name`, since it's not more or less special

paroga (commented):
IMHO the message is misleading, since the name is already shown in the bottom left of the login screen

is there a reason why this is not configurable via the setting pages?

app/views/sessions/new.html.haml
kidhab: There is no benefit if every foodcoop changes their foodcoop's name at the configuration page. Actually a lot of foodcoops didn't do it. But maybe this is a special case at our hosting platform.
As you suggested it would be better to use `FoodsoftConfig:[:name]` to make it editable by users.

config/app_config.yml.SAMPLE
```diff
@@ -167,6 +167,13 @@ default: &defaults
   # default to allow automatically adding new articles on sync only when less than 200 articles in total
   #shared_supplier_article_sync_limit: 200
 
+# Multi_coop example configuration
+# In order to make this work set "multi_coop_install: true"
+#anotherfoodcoop:
+#  <<: *defaults
+#  database:
+#    database: foodsoft_anotherfoodcoop
+
```
paroga, line 176: what's the benefit of adding this lines? IMO everything is already present in the file.

app/views/sessions/new.html.haml
```diff
@@ -4,7 +4,7 @@
       $('#nick').focus();
     });
 
-- title t('.title')
+- title FoodsoftConfig[:name] || t('.title')
```
paroga, line 7: I'm not really happy with this. I'd prefer to keep the `t('.title')` in some form on the login page and use something like `t('.title', FoodSoftConfig[:name])` to pass the name to the translation file, but this will probably result in (too) long lines. what you think about just adding something like `h2 FoodsoftConfig[:name]` after the `title`?

kidhab: Done. Indeed it looks better now with a subtitle.

PR #961, opened 2022-10-25
Add sd_notify

kidhab:
With this [gem](https://github.com/agis/ruby-sdnotify) Puma [supports](https://github.com/puma/puma/blob/master/docs/systemd.md) systemd's Type=notify and watchdog service. This enabled systemd to automatically restart Puma if it locks up.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,6 +41,7 @@ gem 'acts_as_tree'
 gem 'rails-settings-cached', '= 0.4.3' # caching breaks tests until Rails 5 https://github.com/huacnlee/rails-settings-cached/issues/73
 gem 'resque'
 gem 'puma'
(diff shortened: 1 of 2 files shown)


paroga:
LGTM, but please fix `bundle install` error from the GitHub checks:

```
You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.

If this is a development machine, remove the
/home/runner/work/foodsoft/foodsoft/Gemfile freeze
by running `bundle config --delete deployment`.

The dependencies in your gemfile changed

You have added to the Gemfile:
* sd_notify
```

kidhab:
Thanks for this hint. It's should be fine now.

PR #964, opened 2022-10-28
Add sum column to order overview

### LINKED ISSUE #822: Add column displaying the "total value of order" to order list
Request by "Hans":

To increase usability it can be helpful to add a column containing the total value of each order to the order list:

image

kidhab:
Closes #822

### DIFF
--- a/app/views/orders/index.html.haml
+++ b/app/views/orders/index.html.haml
@@ -24,6 +24,7 @@
           %th= heading_helper Order, :pickup
           %th= heading_helper Order, :ends
           %th= heading_helper Order, :note
+          %th= heading_helper Order, :sum
           %th{colspan: "2"}
       %tbody
         - unless @open_orders.empty?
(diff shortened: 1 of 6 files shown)


paroga:
thanks for targeting that issue, but did you check the performance of your change?

[the queries](https://github.com/foodcoops/foodsoft/blob/427561e5ae9bc07585aa727e7d73bd8ebe2315ad/app/controllers/orders_controller.rb#L13-L14) do not join on any of the required associated tables for the `sum` method, resulting in a IMO relevant [N+1 problem](https://stackoverflow.com/questions/97197) if there are many orders

kidhab:
I did test the performance only with the demo instance at foodcoops.net. What is a possible solution - to optimize the `sum` method?

paroga:
a good solution would be to calculate all the values directly via the database and change all usages of the `sum`. An easy solution would be to add all required required models with [`includes`](https://apidock.com/rails/ActiveRecord/QueryMethods/includes) to [the queries](https://github.com/foodcoops/foodsoft/blob/427561e5ae9bc07585aa727e7d73bd8ebe2315ad/app/controllers/orders_controller.rb#L13-L14)

BTW: you also missed to add the `sum` for the `finished_orders`, which are in the same table

RayOei:
No follow-up has been provided. Propose to close.

PR #965, opened 2022-10-30
Show order note as tooltip

### LINKED ISSUE #886: Note text is cut off after about 30 characters in "Manage Orders" table, no easy way to view full note text
I was expecting to be able to view the full text when hovering over it with the mouse. I was also not expecting to see the text cut off in the backend, resulting in unused whitespace when the column is much wider than 30 characters. The full note text is nowhere found in the HTML table.
IMO it would be more useful to have the full text in the HTML and use a CSS property like "[text-overflow: ellipsis](https://www.w3schools.com/cssref/css3_pr_text-overflow.asp)". For example, in the header row, the "Missing Units" column header is abbreviated, with the option to view the full text on mouse over. Why is this not the case with note texts, either for orders or articles?
See attached screenshots.
Screenshot 2021-08-21 at 17-35-38 Foodsoft - Edit order Schmidt
Screenshot 2021-08-21 at 17-37-25 Foodsoft - Manage orders
Screenshot 2021-08-21 at 17-40-49 Foodsoft - Manage orders
Screenshot 2021-08-21 at 17-55-10 Foodsoft - Orders

kidhab:
Closes #886

### DIFF
--- a/app/views/orders/index.html.haml
+++ b/app/views/orders/index.html.haml
@@ -33,7 +33,7 @@
               %td= order.name
               %td= format_date(order.pickup) unless order.pickup.nil?
               %td= format_time(order.ends) unless order.ends.nil?
-              %td= truncate(order.note)
+              %td= truncate(order.note, length: 25, tooltip: true)
               %td= link_to t('.action_end'), finish_order_path(order),
                 data: {confirm: t('.confirm_end', order: order.name)}, method: :post,
                 class: 'btn btn-small btn-success'


twothreenine:
Sounds awesome. Do you have a screenshot how this will look now?
Could also notes with more than 255 characters be displayed properly with this?
Then we could remove/change the character limit of the note.

kidhab:
As far as I remember I choosed the limit of 255 characters because it's used as a limit for another tooltip.

I assume that the tooltip also works with no limit.

Bildschirmfoto vom 2023-03-30 17-42-59

PR #969, opened 2022-11-21
Replace apivore with rswag for api tests

yksflip:
This PR replaces apivore with [rswag](https://github.com/rswag/rswag).
It's still in an early stage, but I thought it would be good to share some thoughts and the progress, as it will be quite a big changeset...

* moving to OpenAPI Spec 3.0.1
* there is a swagger ui which is routed by default from: http://localhost:3000/api-docs/index.html
* the swagger file moved from `/docs/swagger.v1.yml` to `/swagger/v1/swagger.yml`
* swagger file is generated by running ` RAILS_ENV=test rails rswag`, should this be added to the CI?

### ToDo's

* fix rubocop violations
  * [x] `config/routes.rb:1:1: C: Metrics/BlockLength: Block has too many lines. [212/210]`
  * [x] `spec/support/api_helper.rb:8:9: C: RSpec/VariableName: Use snake_case for variable names.`
* rswag translation api endpoints
    * [x] /user
      * [x]  /user/financial_overview
      * [x] /user/financial_transactions
      * [x] /user/group_order_articles
    * [x] /financial_transactions
    * [x] /orders
    * [x] /order_articles
    * [x] /article_categories
    * [x] /financial_transaction_classes
    * [x] /financial_transaction_types
    * [x] /config
    * [x] /navigation


yksflip:
thanks for your feedback! 
Do we actually need to have the swagger.yaml in the tree?
We could generate it in the Dockerfile at build time and we could reference in the docs to smth like https://app.foodcoops.net/api-docs/index.html or any other demo instance.

There's one thing left:
financial_transactions_controller.rb doesn't trows an 422 Error for invalid params. See:  https://github.com/foodcoops/foodsoft/pull/969/files#diff-1c933b556e77c0e5911d866a7c6fbd16b6ed8f195f40e2f079703ae4c074c03fR48 
I guess this would be another issue?

Beside of that I think this rswag stuff seems to work! I know this is quite a lot to review, so I'd be happy for any ideas how we can merge this best.

yksflip:
I've created follow-up issues for the bug in the financial_transaction controller (see todo in the spec file) and for generating the swagger.yml at build time in the Dockerfile.
I think this can be merged now. 
@lentschi do you wan't to have a look first or should I just move on? 
Final thoughts anyone?

carchrae:
i haven't reviewed this, but i'm excited to try it out some time.  i would much rather have a vue.js app running as my customised front end rather than making lot of changes to haml files and then having a nightmare upgrading when a new foodsoft version comes out.

so, thank you.  also, glad you are using swagger.  i haven't used it in ruby, but i have used it in nodejs and java projects before.

spec/requests/api/user/financial_transactions_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+      it_handles_invalid_token_with_id
+      it_handles_invalid_scope_with_id 'user has no ordergroup, is below minimum balance, self service is disabled, or missing scope'
+
+      response '404', 'financial transaction type not found' do
+        schema '$ref' => '#/components/schemas/Error404'
+        let(:financial_transaction) { { amount: 3, financial_transaction_type_id: 'invalid', note: 'lirum larum' } }
+        run_test!
+      end
+
+      # TODO: fix controller to actually send a 422 for invalid params?
```
lentschi, line 48: My personal opinion about TODOs in the codebase is this: Never leave them in there unless there is at least an issue referenced to them. (E.g. `# TODO: https://github.com/foodcoops/foodsoft/issues/xxx - fix controller to actually send a 422 for invalid params?`) Otherwise these TODOs just stay in there forever in my experience.

Apart from that: Is commenting this test out not concealing a real bug?: _Shouldn't_ there be a validation error response when `note` is missing? (I truly don't know as I haven't got a lot of experience with this particular controller. :blush: )

.rubocop_todo.yml
```diff
@@ -266,7 +266,7 @@ Metrics/AbcSize:
 # Configuration parameters: CountComments, CountAsOne, ExcludedMethods, AllowedMethods, AllowedPatterns, IgnoredMethods, inherit_mode.
 # AllowedMethods: refine
 Metrics/BlockLength:
-  Max: 210
+  Max: 212
```
lentschi, line 269: Just out of curiosity: What are these extra two characters required for?

.rubocop_todo.yml
```diff
@@ -451,6 +451,24 @@ RSpec/DescribedClass:
...
+    - 'spec/requests/api/configs_spec.rb'
+    - 'spec/requests/api/financial_transaction_classes_spec.rb'
+    - 'spec/requests/api/financial_transaction_types_spec.rb'
+    - 'spec/requests/api/financial_transactions_spec.rb'
+    - 'spec/requests/api/navigations_spec.rb'
+    - 'spec/requests/api/order_articles_spec.rb'
+    - 'spec/requests/api/orders_spec.rb'
+    - 'spec/requests/api/user/financial_transactions_spec.rb'
+    - 'spec/requests/api/user/group_order_articles_spec.rb'
+    - 'spec/requests/api/user/users_spec.rb'
```
lentschi, line 468: Assuming this directory always contains rswag tests, maybe this might be handier? (untested!):
```suggestion
    # exclude for rswag tests:
    - 'spec/requests/api/**/*_spec.rb'
```

... or maybe better still: have a separate `.rubocop.yml` for this directory, which inherits the base...?

spec/requests/api/user/financial_transactions_spec.rb
yksflip: Yes, I think it's an actual bug, and we should fix it. But I also didn't want to throw it in this PR here ... I'll reference https://github.com/foodcoops/foodsoft/issues/999 there (nice issue number :D) and we can hopefully fix it soon ..

.rubocop_todo.yml
yksflip: I don't rember 100% but I think it's the swagger_docs block in the swagger_helper.rb ... and it's huge.
Maybe there is a way to split it up, but before rswag everything of that was a huge block in swagger.yml

.rubocop_todo.yml
yksflip: yes!

PR #970, opened 2022-11-28
Controller Tests

yksflip:
This PR adds controller tests for: 

* application_controller.rb
* articles_controller.rb
* concerns/auth_concern.rb
* finance/balancing_controller.rb
* finance/base_controller.rb
* home_controller.rb
* login_controller.rb

We found that the  `articles_controller` sync method throws a DoubleRenderError, should we open a new Issue/PR for that?
see spec/controllers/articles_controller_spec.rb:179


spec/controllers/articles_controller_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+        ['unit', [article_a, article_b]],
+        ['unit_reverse', [article_b, article_a]],
+        ['article_category', [article_a, article_b]],
+        ['article_category_reverse', [article_b, article_a]],
+        ['note', [article_a, article_b]],
+        ['note_reverse', [article_b, article_a]],
+        ['availability', [article_a, article_b]],
+        ['availability_reverse', [article_b, article_a]]
+      ]
+      sortings.each do |sorting|
```
paroga, line 32: if all `_reverse` sorting has  the same expected results: how do you test that the code for `name_reverse` is does not sort by e.g. `unit`, which would lead to the same result. IMHO this test gives a false impression and should either done in a way that ensures that the correct part is tested (e.g. with more than two articles) or should be removed, to be clear about missing test coverage.

spec/controllers/articles_controller_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+      expect(response).to render_template('articles/edit_all')
+    end
+  end
+
+  describe "#update_selected" do
+    let(:order_article) { create :order_article, order: order, article: article_c }
+
+    before do
+      order_article
+    end
```
paroga, line 123: i'd suggest to add a second order at line 15 instead. IMHO there should be no need for a dedicated `order_article` factory.

spec/controllers/articles_controller_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+    it 'triggers an article csv' do
+      get :index, params: { foodcoop: FoodsoftConfig[:default_scope], supplier_id: supplier.id }, format: :csv
+      expect(response.header["Content-Type"]).to include("text/csv")
+      expect(response.body).to include(article_a.unit, article_b.unit)
+    end
+  end
+
+  describe "new" do
+    it 'renders form for a new article' do
+      get :new, params: { foodcoop: FoodsoftConfig[:default_scope], supplier_id: supplier.id }, xhr: true
```
paroga, line 48: thins applies to all lines of this kind: i think there is a recurring schema withe the `get`. I think that a function (like `get_with_foodcoop_default_scope_and_created_supplier` with a better and shorter name) would help reducing the code duplication and increase readability

spec/controllers/articles_controller_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+    it 'updates selected articles' do
+      get :update_selected, params: { foodcoop: FoodsoftConfig[:default_scope], supplier_id: supplier.id, selected_articles: [article_a.id, article_b.id] }
+      expect(response).to have_http_status(:redirect)
+    end
+
+    it 'destroys selected articles' do
+      get :update_selected, params: { foodcoop: FoodsoftConfig[:default_scope], supplier_id: supplier.id, selected_articles: [article_a.id, article_b.id], selected_action: "destroy" }
+      article_a.reload
+      article_b.reload
+      expect(article_a.deleted? && article_b.deleted?).to be_truthy
```
paroga, line 134: this applies to all lines of this kind: please create two `expect` lines for that (then you can use the rspec helpers also)

spec/controllers/articles_controller_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+    # TODO: double render error in controller
+    it 'throws double render error' do
+      expect do
+        post :sync, params: { foodcoop: FoodsoftConfig[:default_scope], supplier_id: supplier.id }
+      end.to raise_error(AbstractController::DoubleRenderError)
+    end
+
+    xit 'updates particles from spreadsheet' do
+      post :sync, params: { foodcoop: FoodsoftConfig[:default_scope], supplier_id: supplier.id, articles: { "#{article_a.id}": attributes_for(:article), "#{article_b.id}": attributes_for(:article) } }
+      puts "XXX flash #{flash.inspect}"
```
paroga, line 188: committed debug line :wink:

spec/controllers/concerns/auth_concern_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+  end
+
+  describe 'authenticate_or_token' do
+    describe 'logged in' do
+      let(:token_verifier) { TokenVerifier.new('xyz') }
+      let(:token_msg) { token_verifier.generate }
+      let(:user) { create :user }
+
+      it 'authenticates token' do
+        login user
```
paroga, line 175: `before { login user }`?

spec/controllers/home_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+      before { login fin_user }
+
+      it 'removes user from group' do
+        membership = fin_user.memberships.first
+        get :cancel_membership,
+            params: { foodcoop: FoodsoftConfig[:default_scope],
+                      group_id: fin_user.groups.first.id }
+        expect(response).to have_http_status(:redirect)
+        expect(response).to redirect_to(my_profile_path)
+        expect(flash[:notice]).to match(/#{I18n.t('home.ordergroup_cancelled', :group => membership.group.name)}/)
```
paroga, line 184: `group:` instead of `:group =>`

spec/controllers/login_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        expect(response).to render_template('login/accept_invitation')
+      end
+    end
+
+    describe 'with invalid token' do
+      it 'redirects to login' do
+        get :accept_invitation, params: { foodcoop: FoodsoftConfig[:default_scope], token: invite.token + 'XX' }
+        expect(response).to have_http_status(:redirect)
+        expect(response).to redirect_to(login_url)
+        expect(flash[:alert]).to match(I18n.t('login.controller.error_invite_invalid'))
```
paroga, line 24: what do you think about a helper method for checking the `flash`? it's the same pattern multiple times

spec/support/spec_test_helper.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# spec/support/spec_test_helper.rb
+module SpecTestHelper
+  def login_admin
+    login(:admin)
+  end
+
+  def login(user)
+    user = User.where(:nick => user.nick).first if user.is_a?(Symbol)
```
paroga, line 10: why not `User.find_by_nick`?
i'd prefer `String` over `Symbol`.
AFAIK a `Symbol` does not support `.nick`

spec/support/spec_test_helper.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# spec/support/spec_test_helper.rb
+module SpecTestHelper
+  def login_admin
```
paroga, line 5: i don't see a usage of this (exception throwing?) function

spec/support/spec_test_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def current_user
+    User.find(session[:user_id])
+  end
+end
+
+# spec/spec_helper.rb
+RSpec.configure do |config|
+  config.include SpecTestHelper, :type => :controller
```
paroga, line 23: `type:` instead of `:type =>`

spec/support/spec_test_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+    session[:scope] = FoodsoftConfig[:default_scope] # Save scope in session to not allow switching between foodcoops with one account
+    session[:locale] = user.locale
+  end
+
+  def current_user
+    User.find(session[:user_id])
+  end
+end
+
+# spec/spec_helper.rb
```
paroga, line 21: what's the benefit of this comments?

spec/controllers/concerns/auth_concern_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+        end
+      end
+    end
+
+    describe '#deny_access' do
+      it "redirects to root_url" do
+        login user
+        routes.draw { get "deny_access" => "dummy_auth#call_deny_access" }
+        get :call_deny_access, params: { foodcoop: FoodsoftConfig[:default_scope] }
+        expect(response).to redirect_to(root_url)
```
paroga, line 78: when do you use `_path` and when `_url`?

paroga (changes requested):
first: once again many thanks for the great work!
i just skimmed the changes and didn't look into detail for now. most of comments apply multiple times, but i didn't mark every single occurrence

the rubucop checks for it are not enforced, since we have a lot of existing code, but it would be great to stick to the [Ruby Style Guide for string literals](https://rubystyle.guide/#consistent-string-literals)

i know that's it's a hard to split this PR, when all parts would require the `Gemfile` changes, but it would make the review process easier if the PR would be smaller. maybe start with a single controller test?

yksflip:
thanks for your feedback! We'll look into your comments and I'll change the scope of the PR to fit just the Gemfile and one ControllerSpec at first.

app/controllers/home_controller.rb
```diff
@@ -64,7 +64,7 @@ def ordergroup
   # cancel personal memberships direct from the myProfile-page
   def cancel_membership
     if params[:membership_id]
-      membership = @current_user.memberships.find!(params[:membership_id])
+      membership = @current_user.memberships.find(params[:membership_id])
```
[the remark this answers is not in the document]
yksflip: i think a `find!` method does not exists? 
```
     NoMethodError:
       undefined method `find!' for #<ActiveRecord::Associations::CollectionProxy []>
       Did you mean?  find
```
find_by!` returns a `ActiveRecord::RecordNotFound` instead of `nil` compared to `find_by`...
I added another test to  check that `find` also returns a `ActiveRecord::RecordNotFound` Exception

spec/controllers/login_controller_spec.rb
yksflip: I'm not sure if I grasp your idea correctly ...
what i'd come up with could be smth like 
```ruby
  def expect_flash_alert_to_match(msg)
    expect(flash[:alert]).to match(msg)
  end
 ```
I wouldn't see so much benefit of this refactoring ... But If you prefer it, I can do that ofc

PR #972, opened 2022-12-05
Add home controller test

yksflip:
Co-authored-by: viehlieb <pf@pragma-shift.net>
Co-authored-by: Tobias Kneuker <tk@pragma-shift.net>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -112,7 +112,6 @@ group :test do
   gem 'rspec-core'
   gem 'rspec-rerun'
   gem 'i18n-spec'
-  gem 'rails-controller-testing'
   # code coverage
   gem 'simplecov', require: false
   gem 'simplecov-lcov', require: false
--- a/Gemfile
+++ b/Gemfile
@@ -112,6 +112,7 @@ group :test do
   gem 'rspec-core'
   gem 'rspec-rerun'
   gem 'i18n-spec'
+  gem 'rails-controller-testing'
   # code coverage
   gem 'simplecov', require: false
   gem 'simplecov-lcov', require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -350,10 +350,6 @@ GEM
       sprockets-rails (>= 2.0.0)
     rails-assets-listjs (0.2.0.beta.4)
       railties (>= 3.1)
-    rails-controller-testing (1.0.5)
-      actionpack (>= 5.0.1.rc1)
-      actionview (>= 5.0.1.rc1)
-      activesupport (>= 5.0.1.rc1)
     rails-dom-testing (2.0.3)
       activesupport (>= 4.2.0)
       nokogiri (>= 1.6)
@@ -610,7 +606,6 @@ DEPENDENCIES
   rack-cors
   rails (~> 5.2)
   rails-assets-listjs (= 0.2.0.beta.4)
-  rails-controller-testing
   rails-i18n
   rails-settings-cached (= 0.4.3)
   rails_tokeninput
@@ -647,4 +642,4 @@ DEPENDENCIES
   whenever
 
 BUNDLED WITH
-   1.17.3
+   2.4.5
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -642,4 +642,4 @@ DEPENDENCIES
   whenever
 
 BUNDLED WITH
-   2.4.5
+   1.17.3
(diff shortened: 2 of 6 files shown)


wvengen:
At first glance, great! I can imagine this is useful especially for the Rails upgrade. :+1:

yksflip:
thanks wvengen! 
@paroga  what do you think? Anything needed to get this merged? :)

spec/controllers/home_controller_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+  describe 'GET index' do
+    describe 'NOT logged in' do
+      it 'redirects' do
+        get_with_defaults :profile
+        expect(response).to have_http_status(:redirect)
+        expect(response).to redirect_to(login_path)
+      end
+    end
+
+    describe 'logegd in' do
```
lentschi, line 17: typo:
```suggestion
    describe 'logged in' do
```

Gemfile
```diff
@@ -112,6 +112,7 @@ group :test do
   gem 'rspec-core'
   gem 'rspec-rerun'
   gem 'i18n-spec'
+  gem 'rails-controller-testing'
```
lentschi, line 115: Honestly, considering what is being said in the linked [deprecation message](https://github.com/rails/rails/issues/18950#issue-57746192), I'm not sure, if it's a good idea to use this gem at all:
> Testing what instance variables are set by your controller is a bad idea. 
> [...]The same goes for assert_template.

But I don't feel very strongly about it, so if you really want to use it and noone else has objections, I'm fine with it!

spec/controllers/home_controller_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+      before { login user }
+
+      it 'assigns tasks' do
+        get_with_defaults :index
+
+        expect(assigns(:unaccepted_tasks)).not_to be_nil
+        expect(assigns(:next_tasks)).not_to be_nil
+        expect(assigns(:unassigned_tasks)).not_to be_nil
+        expect(response).to render_template('home/index')
```
lentschi, line 26: See my comment about the `rails-controller-testing` gem above. But if this gem _is_ going to be used after all, couldn't you use this instead?:

```suggestion
        assert_template 'home/index'
```

app/controllers/home_controller.rb
```diff
@@ -18,15 +18,15 @@ def reference_calculator
       @bank_accounts = @types.includes(:bank_account).map(&:bank_account).uniq.compact
       @bank_accounts = [BankAccount.last] if @bank_accounts.empty?
     else
-      redirect_to root_url, alert: I18n.t('group_orders.errors.no_member')
+      redirect_to root_path, alert: I18n.t('group_orders.errors.no_member')
```
lentschi, line 21: The PR says this is about adding a test, not about changing redirect behavior. Has this maybe slipped in from some other branch? (Other than that I'm fine with it!)

spec/spec_helper.rb
```diff
@@ -21,6 +21,10 @@
 
 RSpec.configure do |config|
   # We use capybara with webkit, and need database_cleaner
+  config.before(:suite) do
+    DatabaseCleaner.clean_with(:truncation)
+  end
```
lentschi, line 26: Just out of curiosity: Why did you add this? (Tests seem to run fine without it. Could it be this has accidentally slipped in from some other change?)

lentschi (commented):
Looks fine to me in general; added some minor remarks

app/controllers/home_controller.rb
yksflip: I tried to make the usage of  _path and _url uniform, but no need to do it here...
The change in l67 is actually a fix see https://github.com/foodcoops/foodsoft/pull/970#discussion_r1039455645

Gemfile
yksflip: I think you are right, we should try to conform with the rails conventions.
I'll rework the controller tests and remove tests that step out of the boundaries and try to add the parts as integration tests.

Gemfile.lock
```diff
@@ -642,4 +642,4 @@ DEPENDENCIES
   whenever
 
 BUNDLED WITH
-   1.17.3
+   2.4.5
```
lentschi, line 645: I think, since there are now no changes to the `Gemfile` left, `Gemfile.lock` should remain untouched too. right?

lentschi (changes requested):

Gemfile.lock
yksflip: yess you're right!

wvengen:
Hooray! :tada:

Issue #956, opened 2022-10-13
Upgrade Rails to 6.x


t3k4y:
Foodsoft urgently needs an Rails upgrade since 5.2 is not supported anymore.
(see https://endoflife.date/rails )
Rails 6 will end soon so it will be only an intermediate step.

See also issue on ruby-version: https://github.com/foodcoops/foodsoft/issues/939

t3k4y:
In context of our project  for further developing foodsoft ( https://prototypefund.de/project/weiterentwicklung-von-foodsoft/ ) we are currently working on an pull request solving this issue.

Best wishes, T.

paroga:
100% agree. i did the update from [4 to 5](https://github.com/foodcoops/foodsoft/commit/2557645f4fa223ba6ea562df0443199b73824f20), but there were a lot of incompatibilities, which need (a) code changes and (b) **extensive** manual testing (since the test coverage is not super high). I'm looking forward to merging the tests from the [yksflip fork](https://github.com/yksflip/foodsoft/tree/8_increase_test_coverage) into `master`.

IMHO we should "directly" upgrade to the most recent rails version (similar to my [4.2 to 5.2 upgrade](https://github.com/foodcoops/foodsoft/compare/2892d5272d6d564b1c8fd80acaa59aa291816d96...899a83bec424864e6bbe4ffb38ee1d210d013912) commits)

Do you have people who are willing to do the manual testing?

t3k4y:
Hi paroga, 
thank you for your superfast reply! 

We are currently working on an upgrade to Rails 7.x
It seems pretty straight forward except the apivore incompatibility mentioned in https://github.com/foodcoops/foodsoft/issues/939#issuecomment-1099907198
Our approach is to rewrite the specs and swagger-file using the rswag-gem ( https://github.com/rswag/rswag ) 

In the next days @yksflip will start a pull request with the test code we have written to raise the coverage making the upgrade a little more save.

We are working together with a foodcoop willing to do manual testing in the field. 

Best wishes,
T.

Viehlieb:
Hey folks,
The Upgrade to rails 6.1 looked good and as @t3k4y mentioned, we are directly head to 7.0 and ruby 2.7.2.
Currently we are facing issues with the

gem 'twitter-bootstrap-rails'

This gem supports rails 7, but comes with bootstrap 3, which makes code changes necessary anyways (for instance navbar and modal changed behavior). 

[...]

PR #979, opened 2023-02-17
ruby 2.7.2 and rails 7 upgrade

yksflip:
This PR has the required changes for the ruby and rails upgrade, see  #956 

it'd make sense to first merge these:

2. [x]  rswag api tests #969 
3. [x] rebase master
4. [x] make ci run again
5. [x] run rubocop autofixes
8. [x] create issue for ruby 3 upgrade
6. [x] fix order matrix pdf generation error (sql injection possible?)
    ```
    Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): "SUM(COALESCE(group_order_articles.result, group_order_articles.quantity))".This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql().
    ``` 
9. [x] check flaky tests
10. [x] do some more manual acceptance testing (real life scenarios) 
11. [x] test production (docker) setup

---

1. [ ]  controller tests #970 #972   this isn't a blocker actually. would be nice for better coverage, but let's get the upgrade done first.

happy for any comments or ideas how we can proceed with this! :)


kidhab:
How can we move forward with this and all related MR? I think it is more difficult to merge this if the development continues and the MRs and the master branch diverge more and more.
Maybe it's fine to solve the merge conflicts and just merge it? After all the opener worked in a team for about 6 months on  the Foodsoft code and got support from the folks at Pragma Shift who work on a daily basis with RoR. Also there's a Foodsoft instance out there running for with the RoR update for some months now, isn't it?

Maybe we can ask some Foodcoops which are using the global hosting to use a testing instance with these MR to get more Feedback?

And maybe we can fund some work to solve all related conflicts if it's a lack of time?

kidhab:
With Ruby [2.7.8](https://www.ruby-lang.org/en/news/2023/03/30/ruby-2-7-8-released/) the 2.7 series reached EOL. Maybe this is a good topic for the next community call.

yksflip:
> one question related to the reviewing process: Would it be possible to do the rubocop autofixes (changed due to the rubocop upgrade) in a separate PR? (That new PR could be easily merged without much review as rubocop's autofixes are fairly stable afaik. It would also make this PR - the actual ruby & rails upgrade - easier to review.) Or is that simply not possible due to ruby syntax changes? (I didn't check)

not sure if i understand correctly .. so there are still a lot rubocop violations that need to be autofixed .. I think it would make sense to do it in this PR, but in a seperate commit. I think for reviewing you can select what commits to display, so one could disable the rubocop commit ... The changes that are already in here are needed for syntax reasons I think.

yksflip:
> With Ruby [2.7.8](https://www.ruby-lang.org/en/news/2023/03/30/ruby-2-7-8-released/) the 2.7 series reached EOL. Maybe this is a good topic for the next community call.

why is software aging so fast?  :sweat_smile: 
Should talk about it ... but hopefully going to ruby 3.1 shouldn't be a big deal i guess?

yksflip:
After some more testing I feel pretty confident about this changes. If nobody has any concerns, I'd go on with a optimistic merge ... @wvengen @paroga @lentschi

config/initializers/cors.rb
```diff
@@ -0,0 +1,16 @@
...
+
+# Rails.application.config.middleware.insert_before 0, Rack::Cors do
+#   allow do
+#     origins "example.com"
+#
+#     resource "*",
+#       headers: :any,
+#       methods: [:get, :post, :put, :patch, :delete, :options, :head]
+#   end
+# end
```
wvengen, line 16: For the API, CORS is important. It could be that we do CORS setup elsewhere, in that case we need to move it here, or document that we are doing it elsewhere.

wvengen (approved):
Did a quick scan, looks ok to me in general. Most important thing is that it works in practice, including the different parts of the application. Since most coops don't use each and every feature, it's important to be fully aware of all nooks, ideally testing happens by multiple groups with different approaches. That said, I have confidence that this upgrade is fine, and would work, so a new release noting "you may want to wait upgrading until the next minor version" would probably work as well.

One question about CORS.

In any case, you've done a great work, thank you so much!

config/initializers/cors.rb
wvengen: i.e. it would be nice if running a browser-based app that uses the Foodsoft API would remain working.

yksflip: thanks for your comment! I moved it from config/application.rb to the correct initializer and did a quick test with the foodsoft-shop application

wvengen (approved):
Super, thanks for your quick update! :)

PR #983, opened 2023-03-05
introduce importmaps

yksflip:
This commit introduces importmaps. They allow to use modern javacript ESM within rails without webpack, yarn etc.
see https://github.com/rails/importmap-rails for more details.
This will be useful to add new features e.g. actiontext.
Depends on the rails upgrade #979  

As it  would require larger changes, we decided to keep the legacy javascript in it's old structure for now.


kidhab:
I think this would complete the Ruby/Rails update. Can we merge it?

wvengen (approved):
Super!

(not sure why we include already minified jquery, since seems to be minification already, but that's another issue)

PR #989, opened 2023-03-29
Specify an URL to redirect after logout via settings

kidhab:
This change make it possbile to specify an URL via `logout_redirect_url` at `config/app_config.yml`. After logging out the application redirects to this URL.

We use is to redirect the logout of the [Demo](https://app.foodcoops.net) instance to the [info](https://foodcoops.net/demo/) page.

Maybe this is also useful for other foodcoops.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -21,7 +21,11 @@ class SessionsController < ApplicationController
 
   def destroy
     logout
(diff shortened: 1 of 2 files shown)


yksflip:
Great, thank you!  Maybe it would be good to have it in the  `app_config.yml.SAMPLE` too?

kidhab:
> Great, thank you! Maybe it would be good to have it in the `app_config.yml.SAMPLE` too?

done.

Issue #967, opened 2022-11-19
Prevent adding user to multiple ordergroups


twothreenine:
Since it is not intended for a user to be in multiple ordergroups and this leads to various errors like #631, an error message should appear when a user tries to add a user who is already in an ordergroup to another ordergroup (either by mistake or by not knowing that a user must not be in multiple ordergroups).

aban1:
I will be attempting this soon :)

kidhab:
Can you explain how to add someone to two ordergroups? Playing with the [Demo](https://app.foodcoops.net) I can't reproduce it.
![Bildschirmfoto vom 2023-04-12 23-36-45](https://user-images.githubusercontent.com/32387157/231593054-7d43e99c-d20b-4c3a-9046-165bfe439b79.png)

twothreenine:
> Can you explain how to add someone to two ordergroups? Playing with the [Demo](https://app.foodcoops.net) I can't reproduce it. ![Bildschirmfoto vom 2023-04-12 23-36-45](https://user-images.githubusercontent.com/32387157/231593054-7d43e99c-d20b-4c3a-9046-165bfe439b79.png)

There is no error message when you do that while creating an ordergroup.

aban1:
> Can you explain how to add someone to two ordergroups? Playing with the [Demo](https://app.foodcoops.net) I can't reproduce it.
> ![Bildschirmfoto vom 2023-04-12 23-36-45](https://user-images.githubusercontent.com/32387157/231593054-7d43e99c-d20b-4c3a-9046-165bfe439b79.png)



I don't know the password to the demo website but running it locally when I create two order groups with the same admin account, the account joins both. Here are the screenshots below. 


![Screen Shot 2023-04-12 at 5 54 24 PM](https://user-images.githubusercontent.com/56775972/231603111-b8d31d9b-f2c0-46df-9982-4c022d79fac1.png)



![Screen Shot 2023-04-12 at 5 54 10 PM](https://user-images.githubusercontent.com/56775972/231603067-73dda7a8-afdc-4d9b-a4d7-861ee604fff4.png)

PR #993, opened 2023-04-11
Prevent adding user to multiple order groups

aban1:
I was aiming to resolve [this issue](https://github.com/foodcoops/foodsoft/issues/967) of adding an error message for when someone tries to join multiple order groups. I actually found that there was a method that almost does this, however it only sent an error if a user tried to join a third order group and not a second.


kidhab:
Thanks for your contribution. Can you please remove the commits that contain comments as the only change?

aban1:
Hello so sorry about that, I updated the pull request to only include
relevant changes.

Thank you!

kidhab:
There are still two commits left that addresses other issues. Can you please remove them from this MR?

aban1:
Again so sorry I forgot to switch branches before starting to work on a new task. I just removed those two from this PR. Thank you for your patience.

kidhab:
I just tested it. Your change prevents updating an existing ordergroup because the uniqueness check fails for existing ordergroup members. Can you try it on you instance?

PR #996, opened 2023-04-12
Ingredients List

### LINKED ISSUE #955: Adding ingredients of articles in the foodsoft order page
Hello,

We would like to write all ingredients for the articles in the order, so people see the ingredients at a glance when they order and do not need to browse websites of producers (which in half of the cases are not available because foodcoops tend to buy from small producers). This information is very important for people with allergies or food intolerances, vegans or people who need to do diets for medical reasons, etc. Also most people like to know what's in a product. So this information should be interesting to nearly everybody. That's why we'd love to see the ingredients information integrated in the order page of the foodsoft software.

At the moment there is this green area that appears when you do a mouse over a product in the order. But unfortunately there's not enough space to fit all ingredients there. E.g. the text field "Notizen/notes" which would allow 255 characters get truncated as you see in picture 1.

There would be the field "Herkunft" where we could write alle ingredients, but then one would need to scroll a lot as you see in picture 1.

I guess it is not an option to allow more room for this green area because it worsens the mobile use of the foodsoft. 

But there is already a solution for this problem in the foodsoft. There is the order summary page in the foodsoft (which is shown after an order is saved) and there you can click on an article and see more information right below. You can see this in picture 2.

Would it possible to fit in the ingredients in the order also? 

Is there somebody who wants to implement that? I can offer my help with testing or feedback in regards to looks/content

We use Foodsoft v4.7.99

aban1:
Aiming to resolve #955 

I added ingredients as a separate piece of data for every article of food. This way one can store special information about the food and also the ingredients, rather than placing the ingredients in the notes column. I thought it was important to keep those two separate. Now there is a hover tool tip for ingredients in the order page, and in the order summary page. 

Screen Shot 2023-04-12 at 4 56 03 PM
The mouse is hovering over the Pad Thai tool tip in the screenshot above (order page). 

Screen Shot 2023-04-12 at 5 17 43 PM
There is a tooltip for both notes and ingredients in the order summary page above

Screen Shot 2023-04-12 at 5 19 47 PM
There is now a spot to store ingredients while creating a new article


RayOei:
Based on the responses, I labelled it a `won't fix` and therefore candidate to close.
Personally (and checked with some of our members) I don't see any need for this information, and notes would suffice.
If there is a need bumping (increase) the type should also be easier? Adding a specific additional field which is likely not useful for a big number of articles is also wasteful.
If it _is_ needed, the proposed solution with a plugin seems more logical.

Issue #1000, opened 2023-05-10
Wrong page and total pages in app/controllers/concerns/collection_scope.rb


SomeMichael:
While working on the api I came across a little bug in app/controllers/concerns/collection_scope.rb

page was given as "0" when in api/admin/users/index
total pages was wrong in certain circumstances, since division of two integers yields an integer in ruby.

Fixed here: f4de70d1770c8b60a0296b54f452870f905077f9

yksflip:
thats great, thank you! Feel free to open a PR. We'll try to queue this up as soon as possible (https://github.com/orgs/foodcoops/projects/1)
Maybe you'd like to also write a test for this? :)

SomeMichael:
A test is included [here](https://github.com/SomeMichael/foodsoft/blob/b4348ff33c87477bdbdf10ba3e8756435a0c9a96/spec/requests/api/admin/users_spec.rb#L47-L50)

PR #1001, opened 2023-05-12
API admin/users, bugfix page/total pages

SomeMichael:
As suggested by @yksflip in #1000, I created a PR.

But there doesn't seam to be a way to only include specific commits into the PR. Just let me know if I should put the commits to different branches (E.g. to separate the proposed bugfix)


wvengen:
Hi @SomeMichael thank you for your PR.
To include only specific commits in the PR, you can rebase it on master. If there are still a lot of commits, you may have diverged branches, and need to checkout a clean master branch from this repo, and e.g. cherry-pick the commits you want to include.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/keeping-your-pull-request-in-sync-with-the-base-branch
https://docs.github.com/en/get-started/using-git/about-git-rebase

Also, are you sure pages start at 1 and not at 0 (just to be sure)?

RayOei:
There hasn't been any follow-up so therefore I added `won't fix`.
The actual issue seems more or less trivial? Not checked though if it still exists and #1000 is still open.

PR #1003, opened 2023-06-09
downgrade haml to make deface work

### LINKED ISSUE #980: haml upgrade breaks deface
We found out that haml >= 6.0 breaks the plugins capabilities to alter views.
The deface gem depends on haml, but apparently does not work with haml >= 6.0. No Error thrown, but no functionality given. The documentation says it should work with haml. 

Alterations that have been done inside plugin's engine.rb, e.g. add another NavItem work regardless. Only .deface files are problemativ.

suggested solution
downgrade haml to 5.X and deface will work again

yksflip:
resolves #980


yksflip:
great, then let's merge it :)

PR #1013, opened 2023-06-21
Introduce Message Formatting with Actiontext Trix Editor

yksflip:
This introduces to possibility to send formatted messages and attachments with the messages plugin.

* [x] fix render attachment to actually display a link to attachement
* [x] scheduled cleanup for attachments
* [x] max file size
*~~[ ] test email reply with attachment, is it shown in foodsoft then?~~
  tried  around a bit but still not sure what's going on there, I'd guess the message data will just be shown as a plaintext.
  I think if this is a use case we can have a look at some other point ...

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,1 +1,4 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
+import "trix"
+import "@rails/actiontext"
+import "./trix-editor-overrides"
--- a/app/views/active_storage/blobs/_blob.html.haml
+++ b/app/views/active_storage/blobs/_blob.html.haml
@@ -1,0 +1,9 @@
+%figure{class: "attachment attachment--#{blob.representable? ? "preview" : "file"} attachment--#{blob.filename.extension}"}
+  - if blob.representable?
+    = image_tag blob.representation(resize_to_limit: local_assigns[:in_gallery] ? [ 800, 600 ] : [ 1024, 768 ])
+  %figcaption.attachment__caption
+    - if caption = blob.try(:caption)
+      = caption
+    - else
+      %span.attachment__name= link_to blob.filename, blob
+      %span.attachment__size= number_to_human_size blob.byte_size
--- a/db/migrate/20230215085312_migrate_message_body_to_action_text.rb
+++ b/db/migrate/20230215085312_migrate_message_body_to_action_text.rb
@@ -1,0 +1,32 @@
+class MigrateMessageBodyToActionText < ActiveRecord::Migration[7.0]
+  include ActionView::Helpers::TextHelper
+
+  class Message < ApplicationRecord
+    has_rich_text :body
+  end
+
+  def change
+    reversible do |dir|
+      dir.up do
+        rename_column :messages, :body, :body_old
+        Message.all.each do |message|
+          message.update(body: simple_format(message.body_old))
+          message.body.update(record_type: :Message) # action_text_rich_texts uses STI record_type field and has to be set to the real model
+        end
+        remove_column :messages, :body_old, :text
+      end
+      dir.down do
+        execute "ALTER TABLE `messages` ADD `body_old` text"
+        execute "UPDATE `messages` m
+                  INNER JOIN `action_text_rich_texts` a
+                    ON m.id = a.record_id
+                set m.body_old = a.body"
+        Message.all.each do |message|
+          message.update(body_old: strip_tags(message.body_old))
+        end
+        execute "DELETE FROM `action_text_rich_texts` WHERE `action_text_rich_texts`.`record_type` = 'Message'"
+        execute "ALTER TABLE `messages` CHANGE `body_old` `body` TEXT CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NULL DEFAULT NULL;"
+      end
+    end
+  end
+end
--- a/plugins/messages/app/helpers/messages_helper.rb
+++ b/plugins/messages/app/helpers/messages_helper.rb
@@ -5,7 +5,7 @@ module MessagesHelper
       body = ''
     else
       subject = message.subject
-      body = truncate(message.body, length: length - subject.length)
+      body = truncate(message.body.to_plain_text, length: length - subject.length)
     end
     "<b>#{link_to(h(subject), message)}</b> <span style='color:grey'>#{h(body)}</span>".html_safe
   end
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,9 +49,10 @@ RUN export DATABASE_URL=mysql2://localhost/temp?encoding=utf8 && \
     rm -Rf /var/lib/apt/lists/* /var/cache/apt/*
 
 # Make relevant dirs and files writable for app user
-RUN mkdir -p tmp && \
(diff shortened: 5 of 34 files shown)


yksflip:
two questions that are open to me:

1. Any ideas for a better way in the message .html template than using this extra translations fields?
2. At the moment the db migration, javascript etc. is done globally, even though the message is a plugin. Should I try to move everything to the messages plugin or can it stay there, as Actiontext might be useful in other places e.g. the Wiki?

yksflip:
@lentschi could you have another look at this? :) It' the first rake task I ever wrote, but I think it works as intended.

For the rake task I'd leave a note addressing the new config field in the Release Notes ...

lentschi (approved):

decentral1se (commented):
Quick pass on this, LGTM! Thanks for working on it, definitely nice feature to have.

app/javascript/application.js
```diff
@@ -1 +1,4 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
+import "trix"
+import "@rails/actiontext"
+import "./trix-editor-overrides"
```
paroga, line 4: please end **all** files with newlines. otherwise it generates useless changes, when a the file get modified with an editor that adds those automatically during save

app/views/active_storage/blobs/_blob.html.haml
```diff
@@ -0,0 +1,9 @@
+%figure{class: "attachment attachment--#{blob.representable? ? "preview" : "file"} attachment--#{blob.filename.extension}"}
+  - if blob.representable?
+    = image_tag blob.representation(resize_to_limit: local_assigns[:in_gallery] ? [ 800, 600 ] : [ 1024, 768 ])
+  %figcaption.attachment__caption
+    - if caption = blob.try(:caption)
+      = caption
+    - else
+      %span.attachment__name=link_to blob.filename, blob
+      %span.attachment__size= number_to_human_size blob.byte_size
```
paroga, line 9: please use a consistent coding style also in the haml file (space after equal-sign)

db/migrate/20230215085312_migrate_message_body_to_action_text.rb
```diff
@@ -0,0 +1,10 @@
+class MigrateMessageBodyToActionText < ActiveRecord::Migration[7.0]
+  include ActionView::Helpers::TextHelper
+  def change
+    rename_column :messages, :body, :body_old
+    Message.all.each do |message|
```
paroga, line 5: if you use a `ActiveRecord` in a migration, create it also in the `Migration`, like in the [other files](https://github.com/foodcoops/foodsoft/blob/817e409a2b142acea41520a88d8e7ad7017668a3/db/migrate/20181204070000_create_stock_events.rb#L2-L4). otherwise migration break if the corresponding `ActiveRecord` gets removed from the code in future or the table definition change in an incompatible way

db/migrate/20230215085312_migrate_message_body_to_action_text.rb
```diff
@@ -0,0 +1,10 @@
...
+      message.update_attribute(:body, simple_format(message.body_old))
+    end
+    remove_column :messages, :body_old, :text
```
paroga, line 8: are you sure that `remove_column` takes three arguments?

plugins/messages/app/helpers/messages_helper.rb
```diff
@@ -5,7 +5,7 @@ def format_subject(message, length)
       body = ''
     else
       subject = message.subject
-      body = truncate(message.body, length: length - subject.length)
+      body = truncate(message.body.to_plain_text, :length => length - subject.length)
```
paroga, line 8: please use the new style with `:` instead of `=>`

paroga (changes requested):
i don't know what you preferred way of working is, but if 
a) you want to squash you changes during merge, the PR contains IMHO too much unrelated changes (one commit should change only as little as possible. e.g. (1) just add the trix editor, (2) adopt messaging to use trix, (3) add `prune_old_attachments`)
b) you want to merge it with the whole commit history, the PR contains too much useless commits like [typo fixes](https://github.com/foodcoops/foodsoft/pull/1013/commits/c367681785d78cff4d4353f5ff9becb8e75c721d).

since developers read code much more often than they write it, creating a clean git history should have some importance. (especially if nobody writes any kind of code documentation ;-))

plugins/messages/app/helpers/messages_helper.rb
lentschi: I agree. However discussing this on every MR seems arduous. I suggest adding `Style/HashSyntax` to our rubocop config to enforce this and then run auto-correct on the whole codebase to have a distinct commit.

This may affect more though - I've created an issue for this: https://github.com/foodcoops/foodsoft/issues/1018

yksflip:
thanks for your review paroga! 
You're right regarding the commit history. I was sloppy here and there, I'll split it into more meaningful commits and try to keep that in mind for future PR!

db/migrate/20230215085312_migrate_message_body_to_action_text.rb
yksflip: yes, it's optional, but will be used in context of the `change` method to make the operation reversible.
https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/remove_column

I realized that the migration isn't reversible at all and I'll rewrite this bit

db/migrate/20230215085312_migrate_message_body_to_action_text.rb
yksflip: I tried to implement it like this and came across a challenge: 
The new message body is a RichText field which writes to the `action_text_rich_texts` table and has a record_type field (Single Table Inheritance). Using a extra class in the migration yields to a record_type like `MigrateMessageBodyToActionText::Message`. In the `up` direction of the migration I could simply update this field to be of type `Message` and that should be okay. But I don't know how to read the body again in the `down` direction, as it would try to find only records with the MigrateMessageBody Type. Any ideas? Is there a way to access the model with a modified `record_type` param?

I've got it working with plain sql, but that feels like a very complicated way. Idk what's better to risk that the Migration stops working if the Message Model gets ever removed or to have a complicated migration file.

```
class MigrateMessageBodyToActionText < ActiveRecord::Migration[7.0]
  include ActionView::Helpers::TextHelper
  class Message < ActiveRecord::Base
    has_rich_text :body
  end

  def change
    reversible do |dir|
      rename_column :messages, :body, :body_old
      dir.up do
        Message.all.each do |message|
          message.update(body: simple_format(message.body_old))
          message.body.update(record_type: :Message) # action_text_rich_texts uses STI record_type field and has to be set to the real model
        end
      end
      dir.down do
        execute "ALTER TABLE `messages` ADD `body_old` text"
        Message.all.each do |message|
          message.update(body_old: message.body.to_trix_html) # message.body is empty here because the record_type doesn't match anymore.
        end
        execute "DELETE FROM `action_text_rich_texts` WHERE `action_text_rich_texts`.`record_type` = 'Message'"
      end
      remove_column :messages, :body_old, :text
    end
  end
end
```

yksflip: I've decided to go wit a custom sql statement for the down path

Issue #1006, opened 2023-06-10
fix test  ./spec/models/order_spec.rb:177


yksflip:
```
  1) Order balancing charges correct amounts creates financial transaction with correct amount
     Failure/Error: expect(goa.result).to be > 0

       expected: > 0
            got:   0
     # ./spec/models/order_spec.rb:177:in `block (3 levels) in <top (required)>'
```

yksflip:
look's like the `login` method in `session_helper.rb` is flaky. 
Here parts of the entered password ended up in the username field ... pretty weird. Maybe adding a `wait:` could help... 


![grafik](https://github.com/foodcoops/foodsoft/assets/16109235/90ceec77-6350-4425-b12f-5f4bddfaac79)
![grafik](https://github.com/foodcoops/foodsoft/assets/16109235/606f0193-91bf-4027-be7c-eb4516cd2fc1)

PR #1015, opened 2023-06-21
Suppress net protocol errors

kidhab:
Adding these gems suppresses the following errors (see also https://github.com/ruby/net-imap/issues/16):
```
/2.7.8/lib/ruby/2.7.0/net/protocol.rb:66: warning: already initialized constant Net::ProtocRetryError
/2.7.8/lib/ruby/gems/2.7.0/gems/net-protocol-0.2.1/lib/net/protocol.rb:68: warning: previous definition of ProtocRetryError was here
/2.7.8/lib/ruby/2.7.0/net/protocol.rb:206: warning: already initialized constant Net::BufferedIO::BUFSIZE
/2.7.8/lib/ruby/gems/2.7.0/gems/net-protocol-0.2.1/lib/net/protocol.rb:214: warning: previous definition of BUFSIZE was here
/2.7.8/lib/ruby/2.7.0/net/protocol.rb:503: warning: already initialized constant Net::NetPrivate::Socket
/2.7.8/lib/ruby/gems/2.7.0/gems/net-protocol-0.2.1/lib/net/protocol.rb:541: warning: previous definition of Socket was here

### CODE REVIEW

**Comment by yksflip** [comment]
great you've found a way to supress  the error messages! 

The ci pipeline fails with this error ...
```
You have already activated uri 0.10.0.2, but your Gemfile requires uri 0.12.1. Since uri is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports uri as a default gem. (Gem::LoadError)
```
have you tried updating the bundler verison already?

kidhab:
I set the exact version to avoid the CI error. Good enough to merge?

yksflip:
great kidhab! thank you! 
I think the failed test is somehow flaky, I've opened already a issue to tackle that. #1006 
LGTM

PR #1021, opened 2023-08-23
Multistage Dockerfile

yksflip:
This PR 
* removes a db dependency for assets precompilation
* adds rswag swaggerize swagger.yml generation and thatfore introduces the nulldb adapter
* remove Dockerfile-dev 
*  unites dev and production image in one File using docker multi-stage


yksflip:
any thoughts @paroga @wvengen ? :)

wvengen:
Awesome!
Not sure though about merging the production and dev `Dockerfile`. I think the `Dockerfile` is complex enough without it. And does it mean by doing `docker build -t foodsoft .` builds both the development and production image?
Also, when building the swagger file, I would expect it to have `RAILS_ENV=production`, e.g. routes might be affected by the environment.

wvengen:
p.s. in other projects, I've used a dummy `DATABASE_URL` without even needing the nulldb adapter - but I guess it's cleaner

yksflip:
thanks for your comments!
Hm yeah I see what you mean. I thought it could reduce complexity and I liked the multi-stage idea, but in the end everything blew up a bit.

You can choose what target to build with e.g. `docker build --target development`, but the production stage requires the development as it copies the _swagger.yml_ file. 
We could generate the swagger.yml in the production stage, but that would require the rswag gem in the prod group.

Good point about the RAILS_ENV, I think I used _test_  as the rswag-specs gem is not part of the production group. Afaic the swagger file is generated by the definitions of the `spec/requests/` tests. But I can double check if it makes any differences.

What a hack with the `DATABASE_URL`, I didn't know. I think I'd even prefer that than having another dependency ...

PR #1023, opened 2023-08-23
article file import via plugin

yksflip:


wvengen:
Super to get this within Foodsoft! Especially relevant for foodcoops running their own instance not having to setup sharedlists in some scenario's.

Part of me thinks this is a good candidate for making this part of Foodsoft core, not a plugin.
Also, code is duplicated with sharedlists: what about making the file-related parts a gem, used by Foodsoft and sharedlists?
In any case, these can be additional improvements - happy to see this coming to Foodsoft, one way or another :tada:

yksflip:
thanks willem! :)
I think I picked the wrong branch,  viehlieb actually already build a gem for this: https://git.local-it.org/Foodsoft/foodsoft_article_import
and I think [11_bnn_import_article_update](https://git.local-it.org/Foodsoft/foodsoft/src/branch/11_bnn_import_article_update) follows exactly should be the right branch :o

Lentschi is working atm on a "share supplier via url" feature within the foodcoopsat fork, that could supersede the sharedlists application, see: https://github.com/foodcoopsat/foodsoft_hackathon/issues/20

wvengen:
Ah, super to hear! :rocket:

PR #1028, opened 2023-09-18
Remove feedback form

kidhab:
This MR removes the feedback form.

Personally I don't think there's a need for a regular Foodsoft user to contact some kind of unknown admin team behind an instances. It would be better to use the internal message system to direct problems to a person or group in charge.

In fact at apps.foodcoops.net all requests that came via the feedback form were non-technical questions directed at other foodcoop members/ groups. We were never able to answer such questions because we are not part of these foodcoops.
I think most people don't recognize the text above the form.


yksflip:
Yes, I agree! I think in general it's good to have some way to aggregate feedback. But I think a link to the forum would be fine too.
Could you fix the rubocop violations? 
Any objections @lentschi @wvengen @paroga ?

kidhab:
> Could you fix the rubocop violations?

Done.

yksflip:
thanks, lgtm! I think the failing tests are flaky. #1006

PR #1036, opened 2023-12-22
feat: only show shared suppliers that are allowed to be shown

### LINKED ISSUE #1033: Access control for shared suppliers
[Sharedlists](https://github.com/foodcoops/sharedlists) provides a database of suppliers and their articles. The Foodsoft can connect to this database and automaticalle import/ update a suppliers articles from the Sharedlists database.

Currently all available suppliers from Sharedlists will be shown at the Foodsoft instance. That is not ideal in case of a multicoop Foodsoft setup (running multiple Foodsoft instances from the same codebase and configuration):

There are suppliers that are only usefull for a single foodcoop or suppliers with different price models per foodcoop (e.g. depending on the distance of the delivery).

It would be nice to have some kind of access control for shared suppliers with the aim that a foodcoop only sees suppliers that are enabled for them.

What would the best way to implement this?

As far as I can see this are the most relevant parts from the Foodsoft and Sharedlists:

https://github.com/foodcoops/foodsoft/blob/master/app/views/suppliers/shared_suppliers.html.haml
https://github.com/foodcoops/foodsoft/blob/master/app/controllers/suppliers_controller.rb#L62
https://github.com/foodcoops/foodsoft/blob/master/app/models/shared_supplier.rb
https://github.com/foodcoops/sharedlists/blob/main/db/schema.rb#L35

kidhab:
[This SharedLists commit](https://github.com/foodcoops/sharedlists/commit/8b8afaaf0a2d26edee8c940a90a1b8e0707c558a) introduces a new *foodcoop* column at the SharedLists suppliers table.

The web interface will show a field that expects the name of a Foodcoop (has to be defined via `[FoodsoftConfig[:name]`). If the field is empty, the supplier will be show to all Foodsoft multisite instances.

If the field contains a string the Foodsoft will use it to filter out suppliers that are only allowed to be shown for a specific Foodcoop instance.

This MR introduces the filter at the Foodsoft side.

Closes #1033

### DIFF
--- a/app/models/shared_supplier.rb
+++ b/app/models/shared_supplier.rb
@@ -6,6 +6,7 @@ class SharedSupplier < ApplicationRecord
 
   has_many :suppliers, -> { undeleted }
   has_many :shared_articles, foreign_key: :supplier_id
+  default_scope { where(foodcoop: [FoodsoftConfig[:name], ""]) }
 
   def find_article_by_number(order_number)
     # NOTE: that `shared_articles` uses number instead order_number


kidhab:
Maybe it would be better to introduce a new config item because it's possible to set `[FoodsoftConfig[:name])` via the web interface (and not only via `app_config.yml`).
If someone changes the value via web this would filter out the supplier.

Anyway let me know what you think in general.

yksflip:
hey,
as already mentioned in #1033, we should talk about how we want to proceed with sharedlist in general. I think the suggestion made by lentschi look very good.
But as the timeline of the foodsoftAT fork is not predictable yet, I'd be okay to go with your changes, so you don't have to wait  for it to finish

yksflip:
Maybe you can have another look at the tests, it look's like your changes broke something.

wvengen:
We could use the foodcoop config key (I think it is sometimes also called 'scope') instead of `name`. It's not ideal, I would expect a one-to-many relation (being able to allow multiple foodcoops), but as this may be a temporary solution anyway -  until we have restructured the shared database approach (which will probably take some time) -  it sounds like a great simple approach to solve a relevant use-case.

kidhab:
> Maybe you can have another look at the tests, it look's like your changes broke something.

Can you re-run the failed job so that I can see the error message again? I don't have the necessary permissions.

wvengen:
> Can you re-run the failed job so that I can see the error message again? I don't have the necessary permissions.

It appears this run has expired, and needs a re-push to run CI again (and if that fails, we could try to re-run in case of a flaky test).

PR #1039, opened 2024-01-26
1037 migrate invoice attachments to active storage slow performance financial links

### LINKED ISSUE #1037: migrate invoice attachments to active storage (slow performance financial links)
Our Foodsoft database contains a constantly rising number of invoices with attachments (PDF, JPEG). The following operations have become continuously slower and slower over the years and last now approx 2 to 3 minutes:
* Display unpaid invoices (unbezahlte Rechnungen anzeigen), 
* Create/display financial link (Finanzlink erstellen/anzeigen)
* add invoice or Foodsoft transactions to financial link
* Assign Foodsoft transactions to Bank transactions (Bank Konto Transaktionen zuordnen)

I remember @paroga mentioned that the source of the problem is that the invoice attachment files are stored in the database instead of as separate files which should be fixed in order to keep the Foodsoft performance also for increasing number of invoices with attachments.

yksflip:
Until now invoice attachments were stored in the database directly, this led to performance issues see #1037.
This PR migrates the invoice attachments to use active storage. Additionally multiple attachments are now supported, a small preview is rendered and the delete flow was adapted.

fixes #1037 

Screenshot from 2024-02-10 17-06-02
Screenshot from 2024-02-10 17-05-41
Screenshot from 2024-02-10 17-05-50

### DIFF
--- a/app/controllers/finance/invoices_controller.rb
+++ b/app/controllers/finance/invoices_controller.rb
@@ -79,11 +79,12 @@ class Finance::InvoicesController < ApplicationController
     redirect_to finance_invoices_url
   end
 
-  def attachment
-    @invoice = Invoice.find(params[:invoice_id])
-    type = MIME::Types[@invoice.attachment_mime].first
-    filename = "invoice_#{@invoice.id}_attachment.#{type.preferred_extension}"
-    send_data(@invoice.attachment_data, filename: filename, type: type)
+  def delete_attachment
+    attachment = ActiveStorage::Attachment.find(params[:attachment_id])
+    attachment.purge_later
+    respond_to do |format|
+      format.js
+    end
   end
 
   private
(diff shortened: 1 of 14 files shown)


lentschi:
@yksflip Code looks good to me :+1: 
Though I do have to admit I haven't tried it out and I wonder why #1037 is happening in the first place:
The fact that attachments are stored in the same db table alone shouldn't cause any performance issues, should it? Maybe this could have been solved by adding `attachment_data` to the `ignored_columns` and only load that field whenever we really want to deliver the file contents...?

Anyhow, moving it to active storage is much cleaner :+1:

yksflip:
thx for reviewing! Ah I see, yeah maybe I concluded a bit too quickly, that might have been enough to resolve the issue.

PR #1045, opened 2024-02-17
fix active storage links for attachments

mortbauer:
we had strange issues with the current implementation, best described here:

image

with this fix the issues are gone.


mortbauer:
Ok, rubocop is still failing but this is more or less code from rails activestorage itself.

PR #1050, opened 2024-03-05
move document data from document plugin to active storage

mortbauer:


mortbauer:
I actually realized that I had forgotten to commit the migration I created. 
I tried yours now, the problem I face is that the migrations from the plugins directory are actually not run. Do I have to use another task as db:migrate?

yksflip:
yes you'd either need to copy it manually or run `rake railties:install:migrations` (see `plugins/documents/README.md`)

yksflip:
I thought it would be good to add the migration also to the default path, as the documents plugin is enabled by default and also the other plugin migrations life there ...

PR #1062, opened 2024-04-10
Allow longer bank account descriptions

kidhab:
We recently got the following error:
```
An ActiveRecord::ValueTooLong occurred in bank_accounts#create:

Mysql2::Error: Data too long for column 'description' at row 1: INSERT INTO `bank_accounts` (`name`, `iban`, `description`) VALUES (...)
app/controllers/admin/bank_accounts_controller.rb:11:in `create'
```
Obviously someone tried to enter a long description for the bank account. At the moment the database field is of the type `varchar(255)`. This MR allows to enter longer descriptions.

### DIFF
--- a/db/migrate/20240410093313_change_bank_account_desc.rb
+++ b/db/migrate/20240410093313_change_bank_account_desc.rb
@@ -1,5 +1,5 @@
 class ChangeBankAccountDesc < ActiveRecord::Migration[7.0]
   def change
-    change_column :bank_accounts, :description, :text, 
+    change_column :bank_accounts, :description, :text
   end
 end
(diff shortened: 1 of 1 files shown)


yksflip:
hey kidhab, thanks for this MR! I think rubocop fails because there is a superfluous  `,`  in https://github.com/foodcoops/foodsoft/pull/1062/files#diff-28eff7cc0ab79d26401c7885e1ccb6dbb24c4ec0538e436071b48b060a8820c7R3
besides that, we can merge this.
Additionally we could write a test for the name / iban field, I'm unsure if there is a validation for those, we might want to have that too ..

PR #1066, opened 2024-04-28
Preselect more fields when creating new user

twothreenine:
- use default language from config to preselect language of new user
- add config option "profile_data_public_by_default"; if enabled, the privacy options ("phone/email/name is public") will be pre-checked when creating a new user

In my experience, admins easily forget to set all options correctly when creating a new user. This lead to many users having their language set to English although a translation exists (but often they don't know that they can change it or how) or unintentionally having their phone and email private (i.e. only visible to admins), which really hampers communication within the coop.

I chose to create only one config option for all three privacy options in order not to create too many config options.

One questionable consequence of this is that, if the config option is enabled, created users will have the setting "name is visible" set to true (visible in the admin/users/show view) even if nicknames are disabled, so "name is visible" is irrelevant. This also means, if you later enable nicknames, the names of the respective users will remain visible (unless changed in their profiles), which I think is good.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -45,8 +45,15 @@ class User < ApplicationRecord
 
   before_validation :set_password
   after_initialize do
-    settings.defaults['profile']  = { 'language' => I18n.default_locale } unless settings.profile
-    settings.defaults['messages'] = { 'send_as_email' => true }           unless settings.messages
+    unless settings.profile
+      settings.defaults['profile'] = { 'language' => FoodsoftConfig[:default_locale] || I18n.default_locale } unless settings.profile
+      if FoodsoftConfig[:profile_data_public_by_default]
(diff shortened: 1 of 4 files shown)


yksflip:
thanks @twothreenine!
If you think this are the better defaults for foodcoops, then why just not default it for all instead of adding another config option? Do you think this is something other foodcoops might dislike?
If possible I'd like to avoid adding more config options...

Besides that, it looks good to me! :)

twothreenine:
Food coops have different policies for opting-in / opting-out privacy options.
For example, in "IG Foodsoft" (a super-Foodsoft for the Austrian food coops) we wouldn't enable this setting as it would expose phone numbers and emails to thousands of users. There might also be other rather anonymous coops where contact details should only be visible to admins. But for many smaller coops, it would be useful.

PR #1070, opened 2024-07-12
fix: Add missing migration for articles table

kidhab:
"max_quantity" column was introduced with v4.9
This commit adds the missing migration


yksflip:
ouff, thanks for pointing this out. that was not on purpose actually. I'll remove it from the schema directly

Issue #1056, opened 2024-03-08
Critical bug: Article price changes when closing order


twothreenine:
There seems to be a bug regarding article_prices:

1. Change an article price in the accounting menu without `Globally update current price`, e.g. 1.00 -> 2.00 €
2. In the article menu and in current orders, the price remains unchanged, as well as in other closed orders for the same supplier.
3. But **when an order is closed now, the price suddenly changes** from 1.00 to 2.00 € (the last price this article was changed to in the accounting menu)

Changing a price in the receive menu does _not_ lead to this bug.

lentschi:
I think this has been the case for a long time. (I couldn't manage to run the old release version right now but I bet it was in there too; maybe I'll manage to check tomorrow).

As long as an order hasn't been closed, it's price is subject to changes in the article's master data form. When closing the order the article price becomes fixed to the article's latest `article_price`. I think that part has been intended.

However what this "latest `article_price`" is, can be quite confusing: As far as I can tell right now it's sorted by `article_price.created_at`. When editing the price in the accounting menu in turn, this date field is set to `orders.end` which is something the user can enter manually or even omit. If omitted it is just set to the current datetime.

twothreenine:
I can confirm this was not the case two months ago.

We have a supplier which articles I update monthly (roughly), but we have an order every week.
Dec 3rd price update
Dec 16th - Dec 20th some prices changed via balancing menu
Jan 6th: order PDF sent still shows the correct prices from the last update
Jan 20th next price update

lentschi:
> I can confirm this was not the case two months ago.

I don't know exactly what your workflow was in your foodcoop prior to the upgrade to version 8, but I now could locally start up and reproduce the bug in v7!

Maybe in your workflow you've never updated the price through the accounting menu if another order of the same supplier was in the 'open' state...? (Because this particular bug only occurs in that case.)

lentschi:
I've added a draft PR that would fix this specific error case. It's just an ugly workaround, but if nobody can think of where this might cause trouble, we _could_ merge it. :thinking: (Would have to be tested thoroughly though.)

As already discussed elsewhere, the current article (price) versioning is fundamentally flawed IMO. That's part of why we aim to [change the underlying db structure](https://github.com/foodcoopsat/foodsoft_hackathon/blob/a23f34f4649b051152647664b5df3afdb5d29ce2/db/migrate/20221026102300_alter_articles_add_versioning.rb) in our ongoing article unit project.

[...]

Issue #1058, opened 2024-03-11
Merge Article versioning, extended units and alternative article sharing 


yksflip:
Get #1073 merged 🙏 

1. [Article versioning](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md#1-article-versioning)
2. [Article list synchronization](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md#2-article-list-synchronization)
3. [Article unit system](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md#3-article-unit-system)

For details see the [architecture notes describing the changes](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md).

lentschi:
It now is ready for PR :smile: : see https://github.com/foodcoops/foodsoft/pull/1073

As this unfortunately is a very big PR (now hard to split), I will add some architecture documentation soonish.

lentschi:
I've now added the promised docs.  :blush: 

@yksflip Maybe you could change your issue description accordingly (maybe linking to the same docs as the PR)?

Issue #1030, opened 2023-09-30
Synchronize from external database fails with "out of range" error


kidhab:
A foodcoop at app.foodcoops.net tries to synchronize articles from the external (Sharedlists) database. It fails for one supplier with the following error message:
```
An ActiveModel::RangeError occurred in articles#update_synchronized:

  6000000000 is out of range for ActiveModel::Type::Integer with limit 4 bytes
  app/controllers/articles_controller.rb:194:in `block (2 levels) in update_synchronized'
```

Looking at the parameters from this request it shows entries like this one:
```
"388"=>{"name"=>"cranberry siroop", "shared_updated_on"=>"", "note"=>"Biologisch", "manufacturer"=>"TERSCHELLINGER", "origin"=>"NL", "unit"=>"500ml", "unit_quantity"=>"6000000000.0", "price"=>"0.0", "tax"=>"9.0", "deposit"=>"0.0", "article_category_id"=>"5"}
```

There are others entries which have `unit_quantity"=>"6000000000.0`

The XML file from the supplier lists the unit_quantity as `<sve>6</sve>` which is then saved to the Sharedlists database as `6.0`.

Any idea why this error happened?

yksflip:
I have almost no insights into sharedlist yet, but maybe I can help debugging by digging into it a bit ...

So the unit_quantity is stored correctly in the sharedlists db, right? So I guess the bug is somewhere on the foodsoft side synchronizing the articles.
 
Because in the request params is already the wrong unit_quantity, I guess that the bug must be somewhere before `update_synchronized` is called. Was the unit_quantity correct in the `sync` form view? I think that must be the screen that shows the differences of the updated articles ... I'd guess that it's already wrong here?

can you maybe also provide the article details before syncing?

yksflip:
I think this could happen when the foodcoop article unit is much smaller than the shared_list unit?
e.g. 
```      
article.update!(unit: '0.1ml', unit_quantity: 1)
shared_article.update!(unit: '100ml', unit_quantity: 1.0)
supplier.sync_all ......
```
will give the same out of range error

hamaryns:
Any progress on this?  We still can’t synchronize our prices…

yksflip:
thanks for pinging again! My messy mind completely forgot about it, sorry!
I already managed to reproduce the error in ab1fc34
I'll look into it again!

hamaryns:
[...]

PR #1073, opened 2024-10-11
Article versioning, extended units and alternative article sharing (#1058)

lentschi:
In this PR (addressing #1058) we originally wanted to [extend article units](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md#3-article-unit-system) only, however during development it became clear, that other parts had to be refactored/rewritten too - so it now encompasses these **three major changes**:

1. [Article versioning](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md#1-article-versioning)
2. [Article list synchronization](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md#2-article-list-synchronization)
3. [Article unit system](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md#3-article-unit-system)

For details see the [architecture notes describing the changes](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md).

Since this PR is quite big, we've decided to - instead of overly using comments on the PR - create separate issues and discussions each labeled [extended units PR](https://github.com/foodcoops/foodsoft/issues?q=state%3Aopen%20label%3A%22extended%20units%20PR%22) directly here in the upstream repo.


twothreenine:
Solves #422 Improve unit handling
Solves #474 by allowing price input per price unit (I think the rounding discrepancy should be in a tolerable range now)
Solves #827 Minimum order quantity for certain articles
Solves #843 Allow to insert an integer / number in orders to set amount of a certain product
Solves #850: Full article versioning (still to implement: apply new version specifically to future/current/closed/settled orders, and/or specific order)
Fixes #1056 Critical bug: Article price changes when closing order 
Makes #710 obsolete (unit_quantity == 0 causes 500 when viewing group_order)

We've decided against supporting ordering partial cases, so #645 won't be solved, however if you choose a scalar unit, you can order loose quantities like 2.5 kg.

Sharedlist issues which will probably be affected:
#675 feature: (Un)check articles for article sync
#1030 Synchronize from external database fails with "out of range" error
#1033 Access control for shared suppliers 

Note: Implementation of #569 (Work with datafoodconsortium) could be done based on this PR.

to-do: Check issue numbers below 400

RayOei:
@lentschi I ran into an issue with the migration `20240726083740_alter_articles_add_versioning` step.
The part with 
`article_versions[0..-2].each do |obsolete_version|
        update("UPDATE order_articles SET article_version_id = #{quote latest_version['id']} WHERE article_version_id = #{quote obsolete_version['id']}")
`
Seems to replace the incorrect article_version_id which leads to duplicates and "disappearing" entries.
This halts the migration on the `add_index :order_articles, %i[order_id article_version_id], unique: true` step.

I tried this on actual current 'live' data used in v4.9.1 (I can therefore not share it here but I have logging of the migration).

Example: an entry with `article_price_id = 718` exists and one with `article_price_id=543` 
After migration this column is changed in `article_version_id` but there are now two entries with `718`. The corresponding `order_article.id` for the `543` record has become a second `718` entry and `543` has therefore disappeared as original.
The pre-migration `article_prices` with id `543` has `article_id=0`. As does the `718` entry. It looks like these records are therefore unusable and could be removed and the corresponding order_article entry with this id could also be removed, thus preventing turning into an, I assume, unintended duplicate? 
Was this also the intention? 
The special case here is that both belong to the same `order_id`. But I am unclear why the `543` entry doesn't keep the reference (with a article_id=0) as others do?

I checked a few other with this condition, where the id has been changed, but they don't seem to show this result, I didn't check all of them, though. Either way: those records seem useless as they refer to non existing articles?

robwa:
### Code

I consider a complete code review to be impossible. With such a large scope, the changes would have to be divided into several functionally complete and consecutive commits that can be reviewed one after the other. Subsequent decomposition would take too much time.

### Synchronisation (of article lists)

- There is now a ‘Synchronisation source’ field in the settings of a supplier. A hint is needed as to what form of input is expected there (a URL) and what common URLs to be used in this field look like. Is it only intended for the Foodsoft internal share links?
- The use case for the share link feature is unclear to me: Is it about one foodcoop maintaining the items of one supplier, another relying on it and synchronising with the item list of the first? Are there such cases? As far as I understand Sharedlists, it is mainly about synchronising article lists without an intermediate step using the wholesaler's lists (in the sense of a merchandise management system).
- In the installation I am familiar with, Sharedlists mainly uses mail import (e.g. Antakya, Odin) and FTP import in BNN format (e.g. Terra and possibly others such as Kornkraft/ Naturkost Nord). The Terra BNN, XLS or CSV files, for example, are also available via HTTPS, but a login is required first. The synchronisation feature is therefore unlikely to be usable in its current form for many existing food coops. This does not affect the manual import of files, provided Foodsoft supports the format.
- **I suggest implementing the previous FTP-BNN and mail functionality directly in Foodsoft (without Sharedlists) after consultation with existing foodcoops or wholesalers.** (If we agree on this, we should open an issue.)
- A note is required in the release notes or a comparable document that the existing links to the external sharedlists suppliers will be lost.

### Units

- The migration `20240726083743_create_article_units.rb` inserts ‘unit codes’ directly into the database. We should consider inserting the data as seed data at any time. This was noticed because the migration doesn't work with a SQLite database (use of `NOW()`).
- Further questions and comments on the units will follow.

lentschi:
> I ran into an issue with the migration `20240726083740_alter_articles_add_versioning` step

@RayOei Thanks for reviewing and trying it out with your data!
I can't quite follow what's going wrong. :thinking:  Maybe it would be simplest, if you could provide a rails seed file in a branch forked from the current `master`, after which - when switching to this PR's branch - the migrations would fail. I could then debug it on my machine.
(Or alternatively - if a seed file is too much effort - you could just provide the original primary and foreign keys in a tabular layout, so it would be easier to manually reproduce.)

lentschi:
> I consider a complete code review to be impossible. With such a large scope, the changes would have to be divided into several functionally complete and consecutive commits that can be reviewed one after the other.

Yes, I agree. Like I said elsewhere: The reason this PR has gone out of bounds, is that we started with an issue (article units) that was blocked by several other issues (namely article versioning and dependencies to sharedlists).

So in theory it could have been done in three separate PRs, which would have been easier to review (article versioning -> sharedlists replacement -> article units) - but unfortunately it would mean a lot of effort dividing the changes into functional units.

lentschi:
@RayOei And about your questions about synchronization - it tried to explain the general idea [here](https://github.com/foodcoops/foodsoft/blob/feature/1058-extended-article-units/doc/article_units_fork_architecture_changes.md#2-article-list-synchronization), but maybe it wasn't very clear.

Since - as was pointed out - this PR is so big, I think discussing things like this as global comments here will quickly lead to chaos, so I suggest either we need to find another way. I'll message you about it and we get back to GitHub as soon as we figured out how to best communicate here.

db/migrate/20240726083740_alter_articles_add_versioning.rb
```diff
@@ -0,0 +1,217 @@
...
+
+    # Remove orphaned order articles (db inconsistencies due to lack of foreign key constraints):
+    delete('DELETE FROM order_articles WHERE order_articles.article_version_id IS NULL')
+
+    # De-duplicate article version (db inconsistencies due to lack of unique key for created_at and article_id):
+    duplicate_article_versions = select_all(%{
+      SELECT article_id, created_at
+      FROM article_versions
+      GROUP BY article_id, created_at
+      HAVING COUNT(*) > 1
```
lentschi, line 83: This would partially "fix" #1114:

```suggestion
      HAVING COUNT(*) > 1 AND article_id <> 0
```

However, as it's not really in the scope of #1058, I'm reluctant to add it. Maybe this should be rectified in a separate PR first - also see https://github.com/foodcoops/foodsoft/discussions/1119

RayOei:
@lentschi I can, for some reason, not find your comment regarding a change: 
`-     HAVING COUNT(*) > 1
+      HAVING COUNT(*) > 1 AND article_id <> 0`

I assume this is in the current migration? I would agree with you that, although it may prevent the migration error possibly, it may hide or introduce other problems. 
As the source or the problem is older data which was not part of the main stream, I suggest to leave it as is.
The solution is clear in that the source data needs to be corrected before it can be migrated (for details see #1114).
I don't even think you _have_ to add it as a separate PR because of the before mentioned reason. YMMV.

lentschi:
> I can, for some reason, not find your comment regarding a change

I'm sorry, I'm not sure which comment you're referring to either. :thinking: 

> [...] I assume this is in the current migration? I would agree with you that, although it may prevent the migration error possibly, it may hide or introduce other problems.

The `AND article_id <> 0` part is in none of the PRs migrations and after our discussion in #1114 I chose not to add it there. (Not quite sure, why you're asking this here in the issue comments by the way and not in the respective issue [as agreed](https://github.com/foodcoops/foodsoft/discussions/1087#discussioncomment-12110384) btw...? :slightly_smiling_face: )

RayOei:
> why you're asking this here in the PR comments by the way

@lentschi a comment of yours was sent by Github with the mentioned code. The link in the comment directs to this issue, therefore I added a comment here. I am not sure GH is acting up or something??

lentschi:
> Maybe this should be rectified in a separate PR first.

Yes I agree - it should be.

> also see https://github.com/foodcoops/foodsoft/discussions/1119

<- Yes :+1: Let's continue any discussion about it there, I'd say.

robwa (approved):

mortbauer:
When i run this locally in docker I get the following error when I navigate to `https://<domain>/<fc>/stock_articles`
```
| F, [2025-03-03T20:00:46.287468 #66] FATAL -- : [687e045a-e520-4dda-8d75-fc3778fe20e7]
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7] ActionView::Template::Error (undefined method `name' for #<StockArticle:0x00005fbc213f3b70>):
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     1: %tr{:class => stock_article_classes(stock_article), :id => "stockArticle-#{stock_article.id}"}
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     2:   %td= link_to stock_article.name, stock_article
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     3:   %td= stock_article.quantity
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     4:   %td= stock_article.quantity_ordered
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     5:   %td.main_info= stock_article.quantity_available
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7] app/views/stockit/_stock_article.html.haml:2
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7] app/views/stockit/index.html.haml:71
foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7] app/views/stockit/index.html.haml:70
```

lentschi:
> When i run this locally in docker I get the following error when I navigate to `https://<domain>/<fc>/stock_articles`
> 
> ```
> | F, [2025-03-03T20:00:46.287468 #66] FATAL -- : [687e045a-e520-4dda-8d75-fc3778fe20e7]
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7] ActionView::Template::Error (undefined method `name' for #<StockArticle:0x00005fbc213f3b70>):
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     1: %tr{:class => stock_article_classes(stock_article), :id => "stockArticle-#{stock_article.id}"}
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     2:   %td= link_to stock_article.name, stock_article
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     3:   %td= stock_article.quantity
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     4:   %td= stock_article.quantity_ordered
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]     5:   %td.main_info= stock_article.quantity_available
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7]
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7] app/views/stockit/_stock_article.html.haml:2
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7] app/views/stockit/index.html.haml:71
> foodsoft-1  | [687e045a-e520-4dda-8d75-fc3778fe20e7] app/views/stockit/index.html.haml:70
> ```

Ah, thank you for raising this. I had tested stock articles before creating the PR, but apparently something has broken by a later change.
I created #1132 to tackle this.

lentschi:
> > link_to stock_article.name
> 
> Hmmmm `name` doesn't seem to exist as property with `stock_article` in the config/locales/xx.yml files.

No, but `article_version.name` is - the field has been moved there. `article.name` still exists in the translation files - to avoid breaking stuff. (Since there is a lot of magic happening in rails, it is hard to be certain that all instances of the implicit calls to `t('activerecord.attributes.article.name')` have been removed, so I think it safer to leave it there.)

steffen-unicode:
Thanks @lentschi for all the latest fixes.

We would like to start our developement based on these changes. Who is brave enough to push the merge button?

lentschi:
> Thanks @lentschi for all the latest fixes.

:+1: 

> We would like to start our developement based on these changes. Who is brave enough to push the merge button?

:partying_face: 
I was :blush:  - I think we've waited long enough now.

If this merge causes objections after all, @foodcoops/contributors, @foodcoops/operations, @foodcoops/triage, please let me know soon and I'll try to revert it right away!

Issue #1074, opened 2024-11-05
exclude stock orders from order sum in invoices and financial balances


mjavurek:
I like the feature of [stock orders](https://github.com/foodcoops/foodsoft/issues/389) and we often use it in our FC. During balancing, the stock orders should be excluded from the total amount of the order, because no existing ordergroup will be charged for it: When we add both order and stock delivery to the invoice ([example in foodsoft demo](https://app.foodcoops.at/demo/finance/invoices/13)), the stock order is considered twice: once in the order and a second time by the stock delivery. As a result, a loss for the FC appears with the amount of the stock order.

![grafik](https://github.com/user-attachments/assets/e3b23e83-7735-461f-ab70-e7cd73f5af24)

![grafik](https://github.com/user-attachments/assets/41014f59-0869-4d5b-836d-7a3f8baf1b0d)

mjavurek:
Implementation: https://github.com/foodcoops/foodsoft/pull/1075

PR #1075, opened 2024-11-10
correct invoice balance with stock orders and deliveries 

mjavurek:
According to issue https://github.com/foodcoops/foodsoft/issues/1074, stock orders and deliveries were not correctly considered in invoice balances (invoice view and unpaid invoices), leading to unbalanced invoices in case of deliveries and/or stock orders. 

This patch includes:

- generally ignore stock orders when calculating the groups-sum for an order, since no order-group is charged for it
- invoice view: for the total amount, use only the order sums without stock orders, but display both sums for orders if the sums differ (without/with stock order)
- invoice class: exclude stock orders from orders_sum, new deliveries_sum, consider both orders_sum and deliveries_sum in expected_amount

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -31,21 +31,29 @@ class Invoice < ApplicationRecord
   end
 
   def orders_sum
-    orders
-      .joins(order_articles: [:article_price])
-      .sum('COALESCE(order_articles.units_received, order_articles.units_billed, order_articles.units_to_order)' \
-        + '* article_prices.unit_quantity' \
-        + '* ROUND((article_prices.price + article_prices.deposit) * (100 + article_prices.tax) / 100, 2)')
+    sum = 0
+    for order in orders 
+      sum += order.sum(:groups) 
+    end
+    sum
   end
 
   def orders_transport_sum
     orders.sum(:transport)
   end
-
+  
+  def deliveries_sum
+    sum = 0
+    for delivery in deliveries
+      sum += delivery.sum
+    end
+    sum
+  end
+  
   def expected_amount
     return net_amount unless orders.any?
 
-    orders_sum + orders_transport_sum
+    orders_sum + orders_transport_sum + deliveries_sum
   end
 
   protected
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -220,7 +220,7 @@ class Order < ApplicationRecord
         end
       end
     elsif %i[groups groups_without_markup].include?(type)
-      for go in group_orders.includes(group_order_articles: { order_article: %i[article article_price] })
+      for go in group_orders.includes(group_order_articles: { order_article: %i[article article_price] }).where.not(ordergroup: nil)
         for goa in go.group_order_articles
           case type
           when :groups
--- a/app/views/finance/invoices/show.html.haml
+++ b/app/views/finance/invoices/show.html.haml
@@ -28,12 +28,15 @@
         %dt= heading_helper(Invoice, :orders) + ':'
         %dd><
           - @invoice.orders.order(:ends).each_with_index do |order, index|
+            - sum_og = order.sum(:groups) 
             - sum = order.sum
             - transport = order.transport || 0
-            - total += sum + transport
+            - total += sum_og + transport
             = ', ' if index > 0
             = link_to format_date(order.ends), new_finance_order_path(order_id: order)
-            = ' (' + number_to_currency(sum)
+            = ' (' + number_to_currency(sum_og) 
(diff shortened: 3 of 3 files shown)


mjavurek:
I noticed that the handling of a markup is not considered consistently in balancing invoices. In my opinion, both total amounts without and with markup should be displayed in case a markup is set. The total without markup should match the net invoice amount, and the total with markup should  be displayed to see the total amount that is received from the ordergroups. 

So my proposal is, in case of a markup > 0:
### invoice view 
* **orders and deliveries**: show only sums without markup to keep it clear (currently, sums with markup are displayed!)
* **total without markup**: ...
* **total with markup**: ...
* or: **total**: (without markup) / (with markup)
### unpaid invoices 
show two balance amounts: one without and one with markup, e.g.: **0 € / +9,50 €** which means that without markup, the grouporders sum up exactly to the invoice amount, and with markup, the foodcoop  has 9,50 €  "profit".

mjavurek:
In the _finance  / balance order_ view, the following modifications should make it more correct and transparent:
grafik
* generally show amounts with markup only if markup > 0
* in the order summary:
   * by now an invoice balance was shown, not considering any other orders or delivieries and therefore showing a wrong balance in these cases. therefore, the balances are removed from the summary and placed in the invoice field, considering the invoice balance instead of the order balance.
   * the stock order amount was added if > 0. stock order and group orders should sum up to the gross value.
   * transport costs are shown it there are some. The amount is split in group and FC amounts if a stock order exists.
* in the invoice summary:
   * the numbers of orders (if > 1) and the numbers of deliveries (if > 0) are shown to make it clear why order and invoice amounts differ in these cases
   * the invoice balance is shown, replacing the not always correct order balance previously shown in the order summary  
   * _edit invoice_ is now a button, and a _view invoice_ button is added.

For currency differences in balances, the helper function format_currency_difference() is introduced, showing negative values in red and positive values in green with en extra + sign.

mjavurek:
There is still an issue with the `order.profit()` method. It is only correct if there is only one order and no delivery associated with the invoice. The resulting value is stored in `order.foodcoop_result` in `order.close!()` and displayed for each closed order in the _finance/balancing_ list. 

Example: if there are two orders 
* order 1: 10 € 
* order 2: 20 € 
* invoice for order 1 and 2: 30 €

The overall profit is 0. In contrast, `order.profit` is -20 € (= 10 € - 30 €) for order 1 and -10 € (= 20 € - 30 €) for order 2.  

So basically, the overall profit should be split amongst all orders and deliveries. In my opinion, there is no perfect way to realize it, because it is generally an unsolvable problem without knowing details of the invoice about the parts of the invoice corresponding to the individual orders and deliveries. Example:
* order 1: 10 €, part of invoice: 15 €: profit -5 €
* order 2: 20 €, part of invoice: 20 €: profit: 0 €
* invoice total: 35 €, overall profit: -5 €

Basically I see 2 options:
1) divide the overall profit by the total number of orders and deliveries (simple)
2) split the overall profit according to the individual order or delivery amounts, like 
`order.profit = invoice.profit / invoice.expected_amount * order.sum` (more complicated, but still not more correct)

I am curious if somebody has an idea how to resolve this issue.

yksflip:
Hey @mjavurek,

Thank you so much for your efforts! It's great to see your initiative on this one.

As you're probably aware, our development capacities are very limited at the moment. On the one hand, it's really great to have your contribution, but on the other hand, I can't promise to review and merge it very soon. That's why we established a roadmap and a monthly community call to organize the work (see [Readme](https://github.com/foodcoops/foodsoft?tab=readme-ov-file#roadmap)). If you can make it, it would be really cool to have you on the next call so we can discuss how to organize the work.

Recently, the Austrian team submitted #1073, which will require some time to review, and we want to try to limit the amount of conflicts. I can try to see if there will be conflicts with your contribution. If so, we might need to rework it to fit with it; if not, we might be able to get it through quickly.

Some general remarks: It would be great if you could squash your commits into only a few, more descriptive ones. That would make it easier to review already. Conventional Commits (conventionalcommits.org) might be a good starting point for this.

mjavurek:
Thank you @yksflip for your response. I understand the difficulties, and I hope that despite of the difficulties, it will be possible to merge my pull request in the near future. I am quite sure that there will be conflicts in `app/models/order.rb` and `app/models/invoice.rb`, but it should be easy to resolve them. 

I did not manage to squash my commits. I did all commits in the edit mode on the github website, unfortunately I am not familiar with Git CLI or Git Desktop, which seems to be necessary to be able to squash commits.

PR #1076, opened 2024-12-05
group orders: display deposit, tolerance only if necessary

mjavurek:


mjavurek:
Todos: 
- add `.contains_deposit`  language files

mjavurek:
Actually, the markup is also calculated for the deposit included in the gross price: 

`((price + deposit) * (1+tax)) * markup` 

At least there should be an option to exclude the deposit from the markup, like: 

`(price * (1+tax)) * markup + (deposit * (1+tax))`

twothreenine:
Generally, I think those are good points. However, they will lead to merge conflicts and therefore more effort needed with #1073.

> An additional column is displayed for the deposit if there is at least one article in the order with deposit. This is important for the ordergroups so that they know that the price shown includes a deposit.

I'd prefer a different approach here: The more interesting number is the _price without deposit_, and adding another column will have negative effects on small (mobile) screens. That's why I'd suggest showing the price without deposit instead and adding a small symbol (e.g. a bottle) next to the price of articles with deposit, with an onclick+onhover tooltip that shows the deposit, like `excl. 0.50 € deposit`. (since onhover only would not be mobile-friendly)
This should not affect the calculation of sum / total sum. An additional `incl. deposit` [in total sum] would be nice, but would also make the layout even more problematic.

> Actually, the markup is also calculated for the deposit included in the gross price:
> 
> `((price + deposit) * (1+tax)) * markup`
> 
> At least there should be an option to exclude the deposit from the markup, like `FoodsoftConfig[:deposit_with_markup]`:
> 
> `(price * (1+tax)) * markup + (deposit * (1+tax))`

1. Can you please open an issue for this (if not already exists)? It think it should be done in a separate PR.
2. btw we also had a discussion on pricing models altogether: https://forum.foodcoops.net/t/deposit-not-counted/667/3

PR #1079, opened 2025-01-02
create fincancial link for invoice

mjavurek:
Currently, financial links can only be created for bank transactions. In this PR, a "add finance link" button is added to the invoice view unless there is already a finance link.
grafik

This is useful if invoices are not paid from the bank account of the foodcoop, but from a foodcoop member (e.g. by  credit card). In this case, a foodcoop transaction is required instead of a bank transaction to pay the invoice by crediting the invoice amount to the foodcoop member. After creating a finance link for the invoice, a foodcoop transaction can be easily generated and linked to the invoice.

grafik

### DIFF
--- a/app/controllers/finance/financial_links_controller.rb
+++ b/app/controllers/finance/financial_links_controller.rb
@@ -46,6 +46,9 @@ class Finance::FinancialLinksController < Finance::BaseController
     if params[:bank_transaction]
       bank_transaction = BankTransaction.find(params[:bank_transaction])
       bank_transaction.update_attribute :financial_link, @financial_link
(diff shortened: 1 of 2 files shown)


lentschi:
Thank you for your contribution, @mjavurek - looks reasonable!

**Just a general observation:**

There are currently no unit tests whatsoever for `FinancialLinksController`, right? (If there were, I'd ask you to add one for this use case.)

Since I'm not too deep into all of this BankTransactions/FinancialTransactions logic, I'm finding it hard to review such changes. Maybe someone knowing their way around them can find the time to add tests before we review/merge this?

PR #1080, opened 2025-01-10
Correct and add missing NL translations

RayOei:
Minor typo corrected
Added missing entries for NL

### DIFF
--- a/config/locales/nl.yml
+++ b/config/locales/nl.yml
@@ -922,7 +922,7 @@ nl:
         title: Financiële link %{number}
         type: Type
         remove_from_link: Verwijder link
-        remove_from_link_confirm: De transactie zelf wordt niet verwijderd, alleen de link. Wil je doorgaan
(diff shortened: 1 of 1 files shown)


RayOei:
@wvengen As a first attempt to contribute, I have updated the NL.yml translation as I noticed those missing. Didn't create an issue for it, though. Should I?

RayOei:
Ah.. OK. Will check that. Better not to upset the workflow 😁

RayOei:
@wvengen As a follow-up: needed to dig a bit 😁 (will make some notes on that for newbies) but have CrowdIn installed and working locally. I do not have permissions, though, to download the current state from CrowdIn so preparing a PR from that is also not possible. 
I assume uploading is also not allowed but I added my translations directly in CrowdIn.

wvengen:
Syncing Crowdin translations is something the maintainers need to do, so that is up to the development team.

yksflip (approved):
i don't understand dutch, but more translations should be good :D

Issue #474, opened 2017-05-02
add a property of 'supplier price' and a 'split' value


carchrae:
our co-op, and i think many others, order cases of items from our suppliers.  so, a case of 100 apples might be $50 for 100 apples.  while we show the price to our members as $0.50 per apple and unit quantity (UQ) of 100, the supplier price is $50 and this is what must be communicated to the supplier when ordering.

i have written a bunch of scripts that convert the supplier price into a price per item, but this isn't ideal.  i would prefer to have the original price stored and derive the co-op price from the original price / UQ.

here are some examples of the problems right now:

1. the rounding problem.  so that the co-op does not lose money, i round the price up.  so, if there are 66 pears in a box that costs $50, then price per pear is $0.7575757575 - or rather $0.76.  now, when i compute the supplier price from my derived price of $0.76, the case cost is listed as 66 UQ * 0.76 = $50.16.  the supplier may be confused at the price discrepancy.  (and sure i can see how the coop can slowly get rich!). 

2. if the price of the box changes, it requires more calculating to determine the food coop price.  for example, say some of the fruit is spoiled/bruised, and the supplier says the box will cost $30, then we cannot update the case price, but need to do more calculating.  

3. if the amount in the box is different, then we need to determine the updated cost based on the new quantity.

i propose that we add a 'case price' to the article model.  this price can then be used when producing the ordering pdf/fax to the supplier.  it can also be used when updating the costs of articles in a particular order.

wvengen:
Wow, good practical cases that I recognize from other fresh-based foodcoops.
(1) Adding a case or box price would be great to solve rounding errors. Need to think how to present this clearly without clutter..
(2) and (3) perhaps this may become part of the receive screen, where differences in what was ordered and what was received are handled. Now only member amounts are taken into account, but this could also take article price into account. Since the `order_article` already has its own `article_price`, we can easily assign a new price to it (without affecting the `article`'s price). Just need to make sure that it's clear from the UI why there may be different prices (e.g. message explaining or whatever).

carchrae:
on 1) i'll try a few things and see what i can come up with.  because i know a lot of logic depends on the current setup, i will try and keep the changes minimal.  

thanks for the reminder about `article_price` - i will have a deeper look at the code and see how the received discrepancies are handled in the received/settling order pages.  

[...]

PR #1064, opened 2024-04-21
Improve financial link feature & transaction links

twothreenine:
- add a config option "use_financial_links" for activating both finance links and foodcoop transaction features
- add migration: set this option to true if any financial link or foodcoop transaction exists
- add options to create a financial link and/or foodcoop transaction when balancing an order
- show links to finance links only to users with finance role (required anyway, but links were shown in home/ordergroup)
- add note column when adding a financial transaction to a link
- financial transactions: show link to group_order only in `home/ordergroup` and (additionally) dashboard; replace with link to order's balancing page in finance views
- finance link: show link in financial transaction note to respective order's balancing page (also when adding a financial transaction)
- finance link: replace misleading "delete" button for removing a transaction with "remove from link", also adapt confirm message
- adapt en/de locales, fix finance-related locale typos

Notes:
- I wonder if the additional if/else conditions per row (in multiple views) will have a noticeable effect on performance if there are a lot of rows. I couldn't test this as the performance is horrible anyway in my local docker environment.
- Foodcoop transactions don't show links to the respective order since they don't belong to a group_order and don't have a relation to `order` either. I thought about adding the relation, but I think it's unnecessary to add a column in the db just for foodcoop transactions, as you can go to the order with one additional click (by opening the financial link and clicking on a non-foodcoop transaction).
- I tried to outsource the code for the _transaction note with order link_ to a new template `finance/financial_transactions/_order_note`, but I couldn't call it from financial_links_controller. See comment for details.

Implementation of #848
Replaces #860

PR #1082, opened 2025-01-27
Fix transport cost related issues

twothreenine:
Bug: When creating a foodcoop transaction + finance link when settling an order, the group sum without transport costs was used for the foodcoop transaction, so whenever transport costs were involved, unbalanced finance links were created.

This commit changes:
- Accurate calculation of `ordergroups sum` of an order:
  - before: `total += goa.result * goa.order_article.article_version.fc_group_order_price`
  ignored transport costs and had [rounding issues](https://github.com/foodcoops/foodsoft/issues/474#issuecomment-2743958156), e.g. 3 ordergroups with 16.67 € each were incorrectly added up to 50.00 €
  - new: `total += go.total`
  use the actual group order total including transport costs
  - This value is displayed (among other places) in the balancing summary box. I chose not to include transport costs in `groups_without_markup` though, since that would be more complicated and also because transport costs are some kind of "extra costs."
- When creating a foodcoop transaction when settling an order, use the exact inverted sum of the group transactions to prevent rounding issues.
- Display transport costs above `The following were not ordered` to avoid confusion.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -228,9 +228,12 @@ class Order < ApplicationRecord
         when :groups
           total += go.total
         when :groups_without_markup
+          go_price_without_markup = 0
           for goa in go.group_order_articles
-            total += goa.result * goa.order_article.article_version.gross_group_order_price
+            go_price_without_markup += goa.result * goa.order_article.article_version.gross_group_order_price
           end
+          total += go_price_without_markup.round(2) || 0
+          total += go.transport if go.transport
         end
       end
     end


twothreenine:
I rebased it. Ready to review @yksflip @lentschi 

@mjavurek Perhaps you also want to take a look at this since you also worked on transactions.

RayOei:
Ah... thanks!

Never mind... it helps when you pull from the correct repo. I thought I was on the latest master but I was not in the correct local repo which wasn't updated/synced... 😊

lentschi:
> I rebased it. Ready to review

I'd rather wait for https://github.com/foodcoops/foodsoft/pull/1073 to be merged (since this is bound to happen any minute now :) ) and then review the rebased version of your PR. Sorry for not reviewing/merging it earlier though, I've missed the GitHub notification.

Is that okay for you, @twothreenine, or do you think this to be very urgent that needs to be in the upcoming 4.9.x release? (Is it an old bug or something new?)

twothreenine:
While it isn't fixed, unbalanced financial links will be created when there are transport costs (with the feature I introduced in #1064). The UI issues are old. So nothing critical, but a bit annoying.
I think I had checked that there wouldn't be any merge conflicts with #1073 so I'd prefer to get this out before we commit to a year of finishing release 5.0, but it's also ok for me to delay it.

twothreenine:
I rebased it, resolved the merge conflicts, and used `next` to satisfy RuboCop. Please have a look at that, I don't really know what I'm doing there.

app/models/order.rb
```diff
@@ -405,12 +405,15 @@ def update_price_of_group_orders!
...
-        group_order.ordergroup.add_financial_transaction!(price, note, user, transaction_type, financial_link, group_order)
-      end
+      next unless group_order.ordergroup
+
+      price = group_order.total * -1 # decrease! account balance
+      transaction = group_order.ordergroup.add_financial_transaction!(price, note, user, transaction_type, financial_link, group_order)
+      total += transaction.amount
     end
+    total
   end
```
lentschi, line 417: I'm personally not a big fan of having a method alter stuff and returning other stuff (except for maybe the object itself).

Disregarding that, if you'd like to keep it in there, I'd prefer this from a coding style perspective (untested!):

```suggestion
def charge_group_orders!(user, transaction_type = nil, financial_link = nil)
    note = transaction_note
    amounts = group_orders.includes(:ordergroup).map do |group_order|
      price = group_order.total * -1 # decrease! account balance
      group_order.ordergroup.add_financial_transaction!(price, note, user, transaction_type, financial_link, group_order)

      group_order.total
    end

    amounts.sum
  end
```

The `if group_order.ordergroup` seems to be superfluous: `group_orders.ordergroup_id` is not nullable, but maybe I'm missing something here. If so, it could be rewritten to use a `select` instead.

I'm summing the positive `total` instead of `price`, which would remove the necessity to call `* -1` _again_ in `close!`

However actually I'd even prefer not returning the sum here, but using this in `close!` instead (untested too!):

`total_group_charges = group_orders.map(&:total).sum`

You'd have to test it for performance of course, but I think it shouldn't do much harm there, as caching should alleviate things and it would be much more readable IMO.

lentschi (changes requested):
I'm not sure why I've been requested as the sole reviewer of this - I'm actually not too deep into financial links :blush: .

I've done a quick review and requested a little change, but maybe someone else can review too?

app/models/order.rb
lentschi: > When creating a foodcoop transaction when settling an order, use the exact inverted sum of the group transactions to prevent rounding issues.

Ah, I had missed that - by that you mean that the sum _should_ actually be _less_ precise...?
(When assigning a number with a higher precision than `FinancialTransaction.amount` allows, it is automatically rounded and you're using that effect, right?)
But can you again describe what your intention about that was in the PR description? (I have an assumption, but it would be good for other reviewers to know about that too and it's kind of hard to read the whole discussion about it in https://github.com/foodcoops/foodsoft/issues/474 to find out which change is for what.)

twothreenine:
> I chose not to include transport costs in groups_without_markup though, since that would be more complicated and also because transport costs are some kind of "extra costs."

Btw I realized that groups_without_markup should better be calculated regarding those rounding differences, too, because otherwise it looks confusing when markup = 0 as I noted in https://github.com/foodcoops/foodsoft/issues/474#issuecomment-2762409748.
Also, in the UI it's not `without extra costs`, but `without extra charge` (German: `ohne Aufschlag`), so it should include transport costs.
For this reason, I set this PR back to draft and just added a second commit. Not really happy with the code yet, suggestions welcome.

app/models/order.rb
lentschi: > It is nullable, there's no null: false. This is for the case of stock orders as far as I know.

Ah, you're right, sorry - I forgot about that logic (kind of quirky how stock orders currently work :roll_eyes: ). And I wonder if the `if stockit?` the shouldn't then be extended to include the `charge_group_orders!` as well. But nvm, that's not part of this PR.

> Exactly.
The group sum should show how much the groups are (resp. will be) actually charged.
Also, the financial link should always be balanced, so the food coop transaction should be the exact sum of the group charges.

Okay, then how about this?:

```ruby

# Query shorthand method to make things more readable:
def non_stock_group_orders
  group_orders.where.not(ordergroup_id: nil)
end

# Let the `charge_group_orders!` method return the `FinancialTransaction` instances themselves 
# as IMO that's the least unexpected return value:
def charge_group_orders!(user, transaction_type = nil, financial_link = nil)
  note = transaction_note
  non_stock_group_orders.includes(:ordergroup).map do |group_order|
    price = group_order.total * -1 # decrease! account balance
    group_order.ordergroup.add_financial_transaction!(price, note, user, transaction_type, financial_link, group_order)
  end
end

# ... then in `close!` use this to calculate the sum:
transactions = charge_group_orders!(user, transaction_type, financial_link)
# ...
amount: transactions.sum(&:amount) * -1
```

app/models/order.rb
```diff
@@ -228,9 +228,12 @@ def sum(type = :gross)
         when :groups
           total += go.total
         when :groups_without_markup
+          go_price_without_markup = 0
           for goa in go.group_order_articles
-            total += goa.result * goa.order_article.article_version.gross_group_order_price
+            go_price_without_markup += goa.result * goa.order_article.article_version.gross_group_order_price
           end
+          total += go_price_without_markup.round(2) || 0
```
lentschi, line 235: Why's the `|| 0` needed here?

app/models/order.rb
lentschi: Again - I think you should edit your PR description to something like to make this clearer for other people reviewing:

```
When creating a foodcoop transaction when settling an order, use the exact inverted sum of the group transactions to prevent rounding issues. This is actually _less_ precise than before, but has the advantage of mirroring the loss of precision happening elsewhere and to thus avoid confusion.
```

... Maybe you could elaborate on the "elsewhere" though.

PR #1084, opened 2025-02-03
Fix/run send to supplier as job

mortbauer:
This pull request depends on #1083.

* it will run the send_to_supplier mail in a job
* re-raise any errors in the mail path so it can be re-run

The reason for it is that we had an unnoticed issue with our email setup, which also lead to mails to suppliers not sent from the foodsoft. This errors are not even visible in the email errors overview site since only email problems to members are shown there.


wvengen (commented):
Makes sense! :+1: (looked at last two commits only here)

PR #1085, opened 2025-02-05
improve dev setup

mortbauer:
* add doker build action
* add test worklflow running in docker buildx
* use multistage build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.7.8
+FROM ruby:2.7.8 AS base
 
 RUN supercronicUrl=https://github.com/aptible/supercronic/releases/download/v0.1.3/supercronic-linux-amd64 && \
     supercronicBin=/usr/local/bin/supercronic && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -7,29 +7,48 @@ RUN supercronicUrl=https://github.com/aptible/supercronic/releases/download/v0.1
...
     RAILS_LOG_TO_STDOUT=true \
     RAILS_SERVE_STATIC_FILES=true
 
 WORKDIR /usr/src/app
 
-COPY . ./
+RUN --mount=type=cache,target=/var/cache/apt/ \
+    buildDeps='libmagic-dev mariadb-server nodejs' && \
+    apt-get update && \
+    apt-get install --no-install-recommends -y $buildDeps
```
wvengen, line 23: This is happening in the base stage, so e.g. mariadb-server and the other `buildDeps` are getting into the final image.

wvengen (commented):
Thanks for the PR!
This is something for the current dev team to decide. E.g. we already build container images on Docker Hub, so building it here is a bit superfluous.

Also, I'm not sure if I'd want to build a test image as part of the release workflow. I think passing `RAILS_ENV` as a variable is good, and then we could perhaps adapt the `bundle install` according to it (or even the `--without` argument value, defaulting to `development test`). Then the test workflow can use the same `Dockerfile`, but just run a different command (and if we'd need `.rubocop` files, we could mount them perhaps?)
I think this could be improved a bit more, e.g. precompiling assets in the build stage, then copying them to the dist. Invoking rubocop really belongs in the test workflow, not here (and I'd then also like to be able to run rspec using this image).

PR #1086, opened 2025-02-06
fix(ci): missing libmagic package

yksflip:
more of a quickfix to get the pipeline running again ... maybe good enough for now?
maybe  #1085 can be the proper solution once ready?


yksflip:
pipeline seems to run again, but some tests fail ... @mortbauer could they correlate with your latest changes?

lentschi (approved):

PR #1143, opened 2025-03-21
Fixes inconsistent `Gemfile.lock` and `db/schema.rb`

### LINKED ISSUE #1142: Inconsistent `Gemfile.lock` and `db/schema.rb`
**Current state**

Steps to reproduce (in the default docker environment):
- Check out an older commit like c7fc4758fa25d2326e6d742016ba6924c4e9728e
- Run `rake db:schema:load`
- Run `bundle install`
- Check out a more current commit (anything before the PR fixing this is merged) like 02853a5268742d0641f77d23c4c9f81b83d933ee
- Run `rake db:migrate`
- Run `bundle install`

-> then both `db/schema.rb` and `Gemfile.lock` have uncommitted changes ❌ 

**Expected state**

There are no uncommitted changes after executing the above steps.

lentschi:
Fixes #1142 

The `db/schema.rb` is actually correct in https://github.com/foodcoops/foodsoft/commit/02853a5268742d0641f77d23c4c9f81b83d933ee as far as I can tell, but it was incorrect before for some reason (probably due to an active record upgrade).

So in this PR I added a migration replaying the current schema state. (This migration won't change anything in a database where `schema:load` has already been run, but it will cancel out the uncommitted changes in `db/schema.rb`.)
To avoid such issues in the future, we might add a CI check that fails, if there are uncommitted changes after running migrations from the last commit in the target branch.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -624,6 +624,7 @@ GEM
     zeitwerk (2.6.12)
 
 PLATFORMS
+  ruby
   x86_64-linux
 
 DEPENDENCIES
--- a/db/migrate/20250322103203_fix_schema_inconsistent_with_migration_history.rb
+++ b/db/migrate/20250322103203_fix_schema_inconsistent_with_migration_history.rb
@@ -1,0 +1,13 @@
+class FixSchemaInconsistentWithMigrationHistory < ActiveRecord::Migration[7.0]
+  def up
+    change_column :article_versions, :created_at, :datetime
+  end
+
+  def down
+    change_column :article_versions, :created_at, :datetime, precision: nil
+  end
+
+  def change
+    change_column_null :suppliers, :supplier_category_id, false
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_07_26_083744) do
+ActiveRecord::Schema[7.0].define(version: 2025_03_22_103203) do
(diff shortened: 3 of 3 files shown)


RayOei:
Adding a check is not a bad idea. 
I am, however, having a weird(?) issue with the `db:schema:load` at the moment when I start with an empty database, where it fails with a constraint error on the `create_table active_storage_blobs`. I would expect it to create a new database. I was trying to replay this together with the mentioned `rspec` issue but at this stage I can't run it like GH flow would do. Anything that sounds familiar?

lentschi:
> I am, however, having a weird(?) issue with the db:schema:load at the moment when I start with an empty database, where it fails with a constraint error on the create_table active_storage_blobs. I would expect it to create a new database.

Have you ensured you've dropped the old database before? `db:schema:load` _should_ do that for you since rails 4, but I found that depending on your setup, it doesn't always live up to that expectation. (Hence the `DROP DATABASE development; CREATE DATABASE development` trick in [SETUP_DEVELOPMENT_DOCKER.md](https://github.com/foodcoops/foodsoft/blob/master/doc/SETUP_DEVELOPMENT_DOCKER.md#setup)

lentschi:
Okay, then no - it doesn't sound familiar, sorry. With an empty db I've never had such errors. It's hard to tell without debugging on your machine. 

I think - as this describes a separate bug, it should be...

a. Discussed in an issue, not in a PR's comments
b. Described in more detail: What are the commands you issue (starting without any stopped containers with the `docker-compose` commands from the README leading up until the command that produces the error you mentioned)?

yksflip:
this looks good to me! I think we should discuss @RayOei problem seperately from this

RayOei:
>  we should discuss @RayOei problem seperately from this

For sure, I only asked here for the off change this may have been seen related to the out-of-sync `schema.rb` as I was trying to review/replay this particular change.

PR #1151, opened 2025-04-03
Add information about project for mirroring at openCode

steffen-unicode:
The [publiccode.yml](https://yml.publiccode.tools/) is a file used in the context of open-source software projects, particularly in public administration and the development of software intended for the general public. This file serves to structure and document important information about the project.

In the context of the recent public funding we want wo mirror this repository at [openCode](https://opencode.de/de). The file is a requirement for all repositories on OpenCode.

### DIFF
--- a/publiccode.yml
+++ b/publiccode.yml
@@ -1,0 +1,58 @@
+# This repository adheres to the publiccode.yml standard by including this 
+# metadata file that makes public software easily discoverable.
+# More info at https://github.com/italia/publiccode.yml
+
+publiccodeYmlVersion: '0.2'
+name: Foodsoft
+url: 'https://github.com/foodcoops/foodsoft.git'
+landingURL: 'https://foodcoops.net'
+releaseDate: '2025-03-20'
+softwareVersion: v4.9.1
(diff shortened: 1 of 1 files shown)


publiccode.yml
```diff
@@ -0,0 +1,58 @@
...
+  localisationReady: true
+  availableLanguages:
+    - de
+    - en
+    - es
+    - fr
+    - nl
+    - tr
+description:
+  it:
```
lentschi, line 34: It's my first encounter with publiccode.yml and it's quite likely that I misunderstand something, but shouldn't this be `en:`, if all the sub values are in English?
Even if not, why only Italian?

publiccode.yml
```diff
@@ -0,0 +1,58 @@
...
+    genericName: Webshop
```
lentschi, line 35: Shouldn't this be `Foodsoft` instead of `Webshop`?

yksflip (commented):
besides that one remark of lentschi it looks good to me! Intersting to here about publiccode.yml and love the idea to mirror this to openCode!

PR #1159, opened 2025-04-10
Fix various CSV related bugs

robwa:
When working on CSV re-import I found several small CSV related bugs. Please check the individual commit messages on details.

Issue #1140, opened 2025-03-11
Improve coverage


RayOei:
Improve overall code coverage. It is currently a bit on the low side: https://coveralls.io/repos/foodcoops/foodsoft/badge.svg?branch=master
We should strive for at least 20% more?

RayOei:
Don't worry, I was not about starting to rewrite the tests today 😁 
This is more about approaches and focus and agreements on the matter. 
There are a lot of tools to support that. Choosing one makes a bit more sense if we know the direction of the frontend.

PR #1160, opened 2025-04-10
Improve CSV re-import

### LINKED ISSUE #1128: Improve re-import of article database
For example, in the course of semi-automatic unit migration, the article database of a supplier can be saved in CSV format. It can then be imported again.

Before the export, the database looked like this, for example:
Image

After the CSV restoration, some articles were missing (‘delisted’) and some of the container sizes were not restored in the same way:
Image

Good recoverability could be particularly desirable in the event of unintentional errors during semi-automatic unit migration (see also #1121).

robwa:
- Unavailable articles are not removed by default upon CSV import. There is a new option if one wants this behaviour. When activated, an old article database cannot be restored, because unavailable articles are missing after import.
- Assignment of CSV articles to Foodsoft articles is now done by a new and optional column "Foodsoft ID" in the CSV. Otherwise we couldn't re-import articles without order number.

Depends on #1157 and #1159 

Fixes #1128

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -29,7 +29,7 @@ class ArticlesController < ApplicationController
                                                                                         type: nil)
 
     if request.format.csv?
-      send_data ArticlesCsv.new(@articles, encoding: 'utf-8').to_csv, filename: 'articles.csv', type: 'text/csv'
+      send_data ArticlesCsv.new(@articles, encoding: 'utf-8', foodsoft_url: root_url).to_csv, filename: 'articles.csv', type: 'text/csv'
       return
     end
 
@@ -251,7 +251,8 @@ class ArticlesController < ApplicationController
   # Update articles from a spreadsheet
   def parse_upload
     uploaded_file = params[:articles]['file'] or raise I18n.t('articles.controller.parse_upload.no_file')
-    options = { filename: uploaded_file.original_filename }
+    options = { filename: uploaded_file.original_filename, foodsoft_url: root_url }
+    options[:delete_unavailable] = (params[:articles]['delete_unavailable'] == '1')
     options[:outlist_absent] = (params[:articles]['outlist_absent'] == '1')
     options[:convert_units] = (params[:articles]['convert_units'] == '1')
     @enable_unit_migration = (params[:articles]['activate_unit_migration'] == '1')
--- a/app/lib/articles_csv.rb
+++ b/app/lib/articles_csv.rb
@@ -20,7 +20,8 @@ class ArticlesCsv < RenderCsv
       Article.human_attribute_name(:note),
       Article.human_attribute_name(:article_category),
       Article.human_attribute_name(:origin),
-      Article.human_attribute_name(:manufacturer)
+      Article.human_attribute_name(:manufacturer),
+      @options[:foodsoft_url]
     ]
   end
 
@@ -44,7 +45,8 @@ class ArticlesCsv < RenderCsv
         article.note,
         article.article_category.try(:name),
         article.origin,
-        article.manufacturer
+        article.manufacturer,
+        article.id
       ]
     end
   end
--- a/app/lib/foodsoft_file.rb
+++ b/app/lib/foodsoft_file.rb
@@ -4,9 +4,11 @@ class FoodsoftFile
   # returns two arrays with articles and outlisted_articles
   # the parsed article is a simple hash
   def self.parse(file, options = {})
+    foodsoft_url = nil
     articles = []
-    SpreadsheetFile.parse file, options do |row|
-      next if row[2].blank?
+    SpreadsheetFile.parse file, options do |row, index|
+      foodsoft_url = row[18] if index == 1
+      next if index == 1 || row[2].blank?
 
       article = { availability: row[0]&.strip == I18n.t('simple_form.yes'),
                   order_number: row[1],
@@ -26,6 +28,7 @@ class FoodsoftFile
                   article_category: row[15],
                   origin: row[16],
                   manufacturer: row[17] }
+      article[:id] = row[18] unless foodsoft_url.blank? || foodsoft_url != options[:foodsoft_url]
       articles << article
     end
 
--- a/app/lib/spreadsheet_file.rb
+++ b/app/lib/spreadsheet_file.rb
@@ -10,11 +10,8 @@ class SpreadsheetFile
 
     row_index = 1
     s.each do |row|
-      if row_index == 1
-        # @todo try to detect headers; for now using the index is ok
-      else
-        yield row, row_index
-      end
+      # header detection must be done by using code (e.g. based on index)
+      yield row, row_index
       row_index += 1
     end
     row_index
--- a/app/models/supplier.rb
+++ b/app/models/supplier.rb
@@ -123,10 +123,15 @@ class Supplier < ApplicationRecord
     new_articles = []
 
     data[:articles].each do |new_attrs|
-      article = articles.includes(:latest_article_version).undeleted.where(article_versions: { order_number: new_attrs[:order_number] }).first
+      undeleted_articles = articles.includes(:latest_article_version).undeleted
+      article = if new_attrs[:id]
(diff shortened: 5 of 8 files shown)


yksflip (approved):
nice, looks good to me!

lentschi:
> Assignment of CSV articles to Foodsoft articles is now done by a new and optional column "Foodsoft ID" in the CSV. Otherwise we couldn't re-import articles without order number.

Hm... That might lead to unexpected behavior when Foodsoft A shares such a CSV with Foodsoft B. (The IDs of different articles might overlap.)
But the benefits might outweigh this disadvantage - what do you think @twothreenine ?

lentschi (changes requested):
Thank you for the `delete_unavailable` - seems like a good idea :+1: 

Just two minor things:

1. This includes changes from https://github.com/foodcoops/foodsoft/pull/1159 and https://github.com/foodcoops/foodsoft/pull/1157 - Please rebase the actual commits on master to make this easier to review and to avoid conflicts should there be changes to the other two PRs.
2. And - Just for consideration: There are no unit tests for this new feature. (But I guess it might be a lot of effort to add them as there aren't _any_ for CSV import, right? That was maybe why I myself didn't go into either that when implementing the article units PR.)

robwa:
> Please rebase the actual commits on master to make this easier to review and to avoid conflicts should there be changes to the other two PRs.

Done so.

> Just for consideration: There are no unit tests for this new feature. (But I guess it might be a lot of effort to add them as there aren't _any_ for CSV import, right? That was maybe why I myself didn't go into either that when implementing the article units PR.)

There seem to be at least some simple tests in article and supplier spec. I didn't dive deeper into them. Should I try to add some for this feature?

robwa:
> > Assignment of CSV articles to Foodsoft articles is now done by a new and optional column "Foodsoft ID" in the CSV. Otherwise we couldn't re-import articles without order number.
> 
> Hm... That might lead to unexpected behavior when Foodsoft A shares such a CSV with Foodsoft B. (The IDs of different articles might overlap.) But the benefits might outweigh this disadvantage

Thank you, good point, I didn't think about this before. I can think of a few ways in which we could make improvements:

- We could introduce options when exporting (so far there are none). For a start, it could be enough to activate the ID column only when exporting during unit migration (without the possibility of interaction).
- When importing, there could be a warning if IDs are present in the CSV (either generally or only if they cannot be mapped).
- When importing, there could be an option to ignore the IDs.

twothreenine:
> Unavailable articles are not removed by default upon CSV import.
> [from 78bf4dec5272356135a127e1968af075765dbe57] The parser decides by the existence of "yes" in the first column on the availability. Everything else is assumed to mean "no".

Did you test if the availability also gets updated? I think that has not been implemented yet and I don't see it in your changes.

And I'd suggest:
- to change the behavior to _everything is assumed to mean "yes", except if there's "no"_ (you can leave the availability blank --makes it easier for composing spreadsheets manually)
- CSVs shared between different Foodsoft instances can be in different locales. So we have to check for any locale of `no`.

> Assignment of CSV articles to Foodsoft articles is now done by a new and optional column "Foodsoft ID" in the CSV. Otherwise we couldn't re-import articles without order number.

It would be nice to compare article names, too (after comparing the ID and the order number).
Otherwise if you manually create a CSV of articles you want to add, and use neither order numbers nor IDs, you still can't import them.

twothreenine:
> Hm... That might lead to unexpected behavior when Foodsoft A shares such a CSV with Foodsoft B. (The IDs of different articles might overlap.)

I guess it would be best to include the Foodsoft URL somewhere in the exported CSV and compare it at the import, then ignore the IDs if it doesn't match. Perhaps you could put it in the title cell of the ID column?

robwa:
> > Unavailable articles are not removed by default upon CSV import.
> > [from [78bf4de](https://github.com/foodcoops/foodsoft/commit/78bf4dec5272356135a127e1968af075765dbe57)] The parser decides by the existence of "yes" in the first column on the availability. Everything else is assumed to mean "no".
> 
> Did you test if the availability also gets updated? I think that has not been implemented yet and I don't see it in your changes.

No, the availability doesn't get updated.


> And I'd suggest:
> 
>     * to change the behavior to _everything is assumed to mean "yes", except if there's "no"_ (you can leave the availability blank --makes it easier for composing spreadsheets manually)
> 
>     * CSVs shared between different Foodsoft instances can be in different locales. So we have to check for any locale of `no`.
> 
> 
> > Assignment of CSV articles to Foodsoft articles is now done by a new and optional column "Foodsoft ID" in the CSV. Otherwise we couldn't re-import articles without order number.
> 
> It would be nice to compare article names, too (after comparing the ID and the order number). Otherwise if you manually create a CSV of articles you want to add, and use neither order numbers nor IDs, you still can't import them.

For me, all of the mentioned points (updating availability, yes/no-behaviour, locales, comparison of further fields) are very good and we should consider implementing them. But I see them beyond the scope of #1128, which is on improving re-import of CSV after a unit migration.

As those issues existed already before this PR, could you open one or more separate issues on them @twothreenine?

twothreenine:
> As those issues existed already before this PR, could you open one or more separate issues on them @twothreenine?

Done (updated existing ones), see references above.

robwa:
> > Just for consideration: There are no unit tests for this new feature. (But I guess it might be a lot of effort to add them as there aren't _any_ for CSV import, right? That was maybe why I myself didn't go into either that when implementing the article units PR.)
> 
> There seem to be at least some simple tests in article and supplier spec. I didn't dive deeper into them. Should I try to add some for this feature?

While working on #1100 I realized there are at least some article import tests in the [article import gem](https://git.local-it.org/Foodsoft/foodsoft_article_import). So maybe we want to improve testing when merging the features from #1100?

lentschi (approved):
From a coding perspective this looks good to me now, so I approve the PR. (Also did a quick local smoke test, which was successful) :+1: 

---

**Some general thoughts though:**

As I discussed elsewhere with @twothreenine I find the CSV import/export particularly hard to develop as it has grown to fulfill two contradicting purposes:

1. Export/Import articles to/from a human readable format which may be used to either mass-edit them or manually update them from a supplier source.
2. Being able to perform a partial, per-supplier backup/recovery of the articles db without requiring programming skills and direct db access.

Number 1) was the reason why fields like `available` are even been _translated_ back and forth and why we chose to export units with their translated name instead of the UNECE code.

Introducing a _database_ ID however, is clearly part of 2) and diminishes the human-readability part. (The fact that the foodsoft URL is used as column header doesn't help.)
I think that in the long run we should maybe move 2) to a separate export/import functionality using a less error prone format such as JSON. (I18n should be kept out of that and there should be a format version field to avoid ugly errors when trying to import an old format. Let me know what you think - if you agree, I could create a backlog issue for that new feature.)

But since that would be a lot more effort than your changes here, I think this PR might be a good temporary compromise. What do you think @yksflip, @wvengen, @steffen-unicode  ?

lentschi:
> I am in favor of adding/improving tests, when working on features 👍

@RayOei 
So am I in general.
However, in this case, I'm afraid we'd have to put quite some effort into it as I think there are no existing **unit** tests for import/export whatsoever. 

@robwa 
> There seem to be at least some simple tests in article and supplier spec. I didn't dive deeper into them. Should I try to add some for this feature?

Yeah these are **integration** tests, right. It's up to you really. If I were you I wouldn't invest much time into extending old integration test code as we might shift to a new frontend soon-ish which will cause them all to break.

Writing **unit** tests for **new features** like this one on the other hand would be a very good thing to do, but quite a lot of initial effort, so I understand if you don't want to do that in this PR. :blush: 

Also see my post [here](https://github.com/foodcoops/foodsoft/issues/1140#issuecomment-2724479145)

lentschi:
> While working on https://github.com/foodcoops/foodsoft/issues/1100 I realized there are at least some article import tests in the [article import gem](https://git.local-it.org/Foodsoft/foodsoft_article_import). So maybe we want to improve testing when merging the features from https://github.com/foodcoops/foodsoft/issues/1100?

Sure - sounds good to me :+1:

PR #1161, opened 2025-04-10
chore: update sqlite3 to 1.4

yksflip:
resolves #1138

Issue #1110, opened 2024-08-02
Clarify: Article import: How to proceed with `convert_units` 


lentschi:
How should we proceed with the `convert_units` checkbox ("Keep current units, recompute unit quantity and price (like synchronize).") on CSV upload?
(Not really compatible with the new article system - Do we still need it?)

For reference see the [ported code in this fork](https://github.com/foodcoopsat/foodsoft_hackathon/blob/6a38f7e8cd15a925b814f15632fde62cee6e8397/app/models/article.rb#L202), the [code upstream](https://github.com/foodcoops/foodsoft/blob/b2deb0ed1f93b970a53685580802fed9e6c0e5c1/app/models/article.rb#L195) as well as the [open upstream issue](https://github.com/foodcoops/foodsoft/issues/1059) on that topic.

Issue #1092, opened 2024-07-05
Precision of entered and displayed units 


lentschi:
Re-raising a logical issue:

We've had a similar problem before and I thought it to be solved, but the following logical issue just occurred to me again:

Using the received form for the oranges from our current small.en.seeds (`supplier_order_unit: "Piece (12 kg)", group_order_unit: kg`):
When receiving an amount of 4 kg, the stored value will be displayed as 3.996 kg. That happens as the `units_received` are stored in `supplier_order_unit` (to be directly and safely comparable with `units_to_order` which is obviously stored in the same unit) and 4 / 12 is 0.3333333... but only 3 decimals are stored. So when trying to convert it back for re-editing/display 0.333 * 12 turns out to be 3.996.

The quick fix would be to simply increase precision (to 6 e.g.), but obviously, if ratios are to high, the problem will still occur.

This is related to, but I believe not quite the same, as foodcoops/foodsoft#1094.

Issue #1059, opened 2024-03-29
deprecate article unit convertion from ST to KI (german foodcoops legacy)


yksflip:
in `models/article.rb ` is this convert_units special case that converts articles with the unit 'KI' (Kiste) to 'ST' (Stück). The comment says, that it's some legacy feature for german foodcoops. With the upcoming article_versions, this get's super obsolete. Will anyone still relies on that or can we deprecate it after we merged article_versions?

https://github.com/foodcoops/foodsoft/blob/master/app/models/article.rb#L195
https://github.com/foodcoopsat/foodsoft_hackathon/blob/master/app/models/article.rb#L201

PR #1162, opened 2025-04-10
1030 synchronize from external database fails with out of range error

yksflip:
depends on #1161, which I needed to get rspec running :D

This should resolve the out of range errors that appeared in #1030 

I'm a little bit afraid to touch this convert_units funciton, as it's really a bit messy and quite essential. Maybe someone could do some manual testing too?

Also now reading the discussion around [#1110](https://github.com/foodcoops/foodsoft/issues/1110) I wonder if we even still need this?
And reading [#1092](https://github.com/foodcoops/foodsoft/issues/1092) makes me wonder if it would be safe change to use float instead of rational

### DIFF
--- a/app/models/article.rb
+++ b/app/models/article.rb
@@ -213,7 +213,7 @@ class Article < ApplicationRecord
         nil
       end
       if fc_unit != 0 && supplier_unit != 0 && fc_unit && supplier_unit && fc_unit =~ supplier_unit
-        conversion_factor = (supplier_unit / fc_unit).to_base.to_r
+        conversion_factor = (supplier_unit / fc_unit).to_base.to_f
         new_price = new_article.price / conversion_factor
         new_unit_quantity = new_article.unit_quantity * conversion_factor
         [new_price, new_unit_quantity]
--- a/spec/models/article_spec.rb
+++ b/spec/models/article_spec.rb
@@ -52,6 +52,13 @@ describe Article do
       expect(new_unit_quantity).to eq 1.5
       expect(new_price).to eq 120
     end
+
+    it 'does not exceed limits when converting from 100ml to 0.1ml' do
+      article1 = build(:article, supplier: supplier, unit: '0.1ml', unit_quantity: 1)
+      article2 = build(:article, supplier: supplier, unit: '100ml', unit_quantity: 1.0)
+      _, new_unit_quantity = article1.convert_units(article2)
+      expect(new_unit_quantity.to_f).to eq 1000
+    end
   end
 
   it 'computes changed article attributes' do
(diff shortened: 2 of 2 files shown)


lentschi:
> depends on https://github.com/foodcoops/foodsoft/pull/1161, which I needed to get rspec running :D

Merged - you should be able to rebase on master now

lentschi:
> Also now reading the discussion around https://github.com/foodcoops/foodsoft/issues/1110 I wonder if we even still need this?

:shrug: Yeah, good question. Maybe we should transfer it into a discussion? (I just haven't done that so far as having the issue has the advantage of it showing up on the board so we don't forget about it before the release.)

yksflip:
okay great, thanks for your comments! 
tbh. I do struggle to integration test this, as I don't fully understand how/if the csv import with unit-conversion behaves correctly. 
The original error came up in a sharedlist scenario, which just does not exist anymore. Using a shared supplier will have you use the new article-unit system anyway.

Another way convert_unit is called is from the csv import.
But when I check it and try to upload a csv I see the differences in the multi-edit form, but I cannot proceed to finish the import, as some js form error occurs: `The invalid form control with name=‘articles[0][article_unit_ratios_attributes][1][unit]’ is not focusable.`

Maybe this is another bug?
Steps to reproduce:
1. Select the Supplier nutmaker -> Articles -> Export CSV (Articles are not unit-migrated!) 
2. Change the first article from 1.0 kilogram to 0.1 kilogram and the price to smth else
3. Select Upload articles, select the file, check " Keep current units, recompute unit quantity and price (like synchronize). "
4. Click on Process Upload

lentschi:
> Export CSV (Articles are not unit-migrated!)

You mean you exported it with Foodsoft v4? 
(Because if you exported it with the current master schema, the Nuthome's articles _are_ in fact "unit-migrated". The "Migrate old article units..." button just shows due to [yet _another_ bug](https://github.com/foodcoops/foodsoft/issues/1163) :see_no_evil: ).

yksflip:
gotcha! Thanks for pointing this out :) 
I'll try it another time with a actually not migrated supplier :D 
Maybe it makes also more sense for this to be a patch for foodsoft v4.

lentschi:
> I'll try it another time with a actually not migrated supplier :D

No, I think that is unrelated. I just wanted to make sure, I got the reproduction case right. And I _can_ reproduce it with fully migrated articles too. :+1: 

However, I think the error you're describing, is what I meant by https://github.com/foodcoops/foodsoft/issues/1110 - the checkbox just can't work with the new system currently. (I didn't want to invest time into finding a hacky solution, if maybe our decision might be to just abandon the feature - also see https://github.com/foodcoops/foodsoft/issues/1059)

https://github.com/foodcoops/foodsoft/issues/1110 is on the board and prioritized.

The decision doesn't have to be that we abandon the feature though - if this is still used by a lot of foodcoops, I'm sure we can find a way to re-integrate it. It's just always hard to tell whether this is the case as we don't have any kind of tracing.

yksflip:
I see! Okay then we can discuss it further in #1110  and I'll try to adapt this patch to the latest foodsoft v4 version

lentschi (approved):

PR #1168, opened 2025-04-16
fix: text fields onchange search

henning-unicode:
Automatic search when typing into a text search field is broken in the current version. The bug was introduced in this commit: https://github.com/foodcoops/foodsoft/blame/e4a9932caa44fc8960ddb3a88956a6d02fda25f3/app/assets/javascripts/application_legacy.js#L102

This PR adds search for both `type=number` and `type=text` inputs. 

@lentschi Was the replacement of `type=text` a mistake? It would be great if you could check whether this works with the changes you introduced.

### DIFF
--- a/app/assets/javascripts/application_legacy.js
+++ b/app/assets/javascripts/application_legacy.js
@@ -99,7 +99,8 @@ $(function() {
     // Submission will be done after 500ms of not typed, unless data-submit-onchange=changed,
     // in which case it happens when the input box loses its focus ('changed' event).
     // (changeDate is for bootstrap-datepicker)
-    $(document).on('changed keyup focusin changeDate', 'form[data-submit-onchange] input[type=number][type=text]:not([data-ignore-onchange])', function(e) {
+    $(document).on('changed keyup focusin changeDate',
+    'form[data-submit-onchange] input[type=text]:not([data-ignore-onchange]), form[data-submit-onchange] input[type=number]:not([data-ignore-onchange])', function(e) {
         var input = $(this);
         // when form has data-submit-onchange=changed, don't do updates while typing
         if (e.type!='changed' && e.type!='changeDate' && input.parents('form[data-submit-onchange=changed]').length>0) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/application_legacy.js
```diff
@@ -99,7 +99,9 @@ $(function() {
     // Submission will be done after 500ms of not typed, unless data-submit-onchange=changed,
     // in which case it happens when the input box loses its focus ('changed' event).
     // (changeDate is for bootstrap-datepicker)
-    $(document).on('changed keyup focusin changeDate', 'form[data-submit-onchange] input[type=number]:not([data-ignore-onchange])', function(e) {
+    $(document).on('changed keyup focusin changeDate',
+    'form[data-submit-onchange] input[type=text]:not([data-ignore-onchange]), form[data-submit-onchange] input[type=number]:not([data-ignore-onchange])', function(e) {
+        console.log("foo");
```
lentschi, line 104: With less code & removing the `console.log`:

```suggestion
        $(document).on('changed keyup focusin changeDate', 'form[data-submit-onchange] input[type=number][type=text]:not([data-ignore-onchange])', function(e) {
```

lentschi (changes requested):

app/assets/javascripts/application_legacy.js
lentschi, line 104: Forgotten `console.log`:

```suggestion
```

lentschi (changes requested):

lentschi:
> Was the replacement of type=text a mistake? 

Whoopsie - Yeah, seems like it. (I hadn't known this was also used for non-numeric fields after my changes with the quantity fields.) Thank you for raising & fixing this!

> It would be great if you could check whether this works with the changes you introduced.

Tested & it does. :+1: 

For reference - here's the commit introducing the bug in our fork (before the squash):
https://github.com/foodcoops/foodsoft_hackathon/commit/5a29c5607cd23edddf6fe0c498547e7661093bbd

henning-unicode:
@lentschi thanks for the feedback! I made the changes you suggested

lentschi:
> Did you deleted the suggested change?

Yes, I did :nerd_face:

lentschi (approved):

PR #1169, opened 2025-04-19
create/edit article: sort article categories alphabetically

mjavurek:
actually, categories in the selection field are sorted by id. an alphabetical sorting would make it much easier to find the desired category. 
grafik


lentschi (approved):

lentschi:
@RayOei 
> just to be sure: this one is safe to merge, correct?

Ah, yes, thank you. I was waiting for @mjavurek to do so, but I guess he doesn't have the required permissions...?

Issue #1170, opened 2025-04-22
Add user documentation


robwa:
I suggest that we create user documentation together with Foodsoft and maintain it in a similar way to the tests. My feeling is that this should be

- be managed together with the Foodsoft code in the repository so that it matches the respective version,
- be created with a tool such as [VitePress](https://vitepress.dev/), which is made for this,
- be written primarily in English (and only translated if necessary), as this is the only way it can be maintained by all developers,
- be delivered with the deployment of Foodsoft, so that there can be direct links from Foodsoft to the documentation,
- be maintained and expanded in the same way as the tests are maintained and expanded when issues are worked on.

There are already some issues for which we could use such a system (e.g. #1106). Specifically, I see the need because I would like to maintain the documentation of the formats (e.g. the Foodsoft table format) outside the HTML forms when working on the import and export formats.

lentschi:
Great idea 👍

lentschi:
@mjavurek What do you think about this? Could we pull https://docs.foodcoops.net/de/home into https://github.com/foodcoops/foodsoft/tree/master/doc or is there an advantage in keeping it separate? (I don't know which technology is being used in https://docs.foodcoops.net/de/home, but @twothreenine indicated you might...?)

mjavurek:
The technology used in https://docs.foodcoops.net/de/home is a https://js.wiki/ 
The content is stored in MD syntax and mirrored to https://github.com/foodcoops/user-docs. @mariandl  was involved in the setup and should know more about it.

lentschi:
> The technology used in https://docs.foodcoops.net/de/home is a https://js.wiki/ The content is stored in MD syntax and "mirrored" to https://github.com/foodcoops/user-docs, see https://docs.foodcoops.net/de/documentation/development/howtowrite#wiki-js It is obviously an incremental backup and should be cleaned because there were several changes in the hierarchical structure during the writing process of the documentation and the (now abandoned) folders and files used before these changes are still there. Pages can be displayed on github due to the MD syntax, even embedded graphics are shown, but links to internal pages do not work on github. [@mariandl](https://github.com/mariandl) was involved in the setup and should know more about it.

Ah, great - thanks for the info!

So we could in theory to put it very shortly

1. move the generated sources from https://github.com/foodcoops/user-docs to [foodsoft/doc](https://github.com/foodcoops/foodsoft/tree/master/doc),
[...]

PR #1172, opened 2025-04-29
docs folder: Create readme.md

mjavurek:


doc/readme.md
```diff
@@ -0,0 +1,6 @@
+# Foodsoft Development Documentation
+
+Here are some documentations helpful for the foodsoft development. 
+
+A user documentation can be found at https://docs.foodcoops.net/
```
lentschi, line 5: Hm... considering the readme is written in English, I guess this should link to https://docs.foodcoops.net/en/home instead, but then again there's nothing there :thinking: 
So, to be honest, I'm not sure if this PR makes sense at this moment. Probably we should first decide which way we go with https://github.com/foodcoops/foodsoft/issues/1170 ...? What do you think?

lentschi: @mjavurek Since there hasn't been an answer to this since May 25, I'm converting this into draft. 
This also depends on the outcome of https://github.com/foodcoops/foodsoft/discussions/1252.

mjavurek: Still many people using the foodsoft are not aware that there is this documentation. Most of them understand German. The documentation enhances the usability of the foodsoft for both end users and foodcoop admins. So I would be glad if the link could be included here.
I hoped that somebody would get inspired to translate the German documentation to other languages if it is linked here.

lentschi: I see your point and maybe that's a better-than-nothing decision. :+1:  But I'd say let's discuss this as a group elsewhere - see https://github.com/foodcoops/foodsoft/discussions/1252#discussioncomment-15953930

lentschi: -> Converting it back into ready-for-review (but not merging it just yet).

PR #1179, opened 2025-05-10
fix(mollie): skip payment method if pricing is empty

### LINKED ISSUE #1178: [BUG]: Mollie API doesn't contain pricing information anymore
**Describe the bug**
An error was reported that the Mollie plugin didn't work anymore (500 error on /mollie/new). 
A nil error happens on `mollie_controller.rb  : payment_methods method.pricing`
It seems that the mollie api doesn't provide the pricing information that are used to calculate the transaction costs anymore.

**To Reproduce**
1. Enable mollie plugin and configure with api key
2. ordergroup -> credit acount ( f/payments/mollie/new)
3. shoud show the screen to select a payment method, but shows error

**Foodsoft Version**
- 4.9.1

yksflip:
hotfix for #1178 
it fixes the problem only partially. It's still not possible to calculate the transaction fees, but when disabling the option to charge transaction fees the plugin can already be used again.

### DIFF
--- a/plugins/mollie/app/controllers/payments/mollie_controller.rb
+++ b/plugins/mollie/app/controllers/payments/mollie_controller.rb
@@ -78,6 +78,8 @@ class Payments::MollieController < ApplicationController
   def payment_methods
     @payment_methods = fetch_mollie_methods
     @payment_methods_fees = @payment_methods.to_h do |method|
+      next [method.id, []] if method.pricing.blank?
+
       [method.id, method.pricing.map do |pricing|
         {
           description: pricing.description,


RayOei:
@yksflip What does this actually mean from a user perspective? Are the Mollie transaction fees not added to the total amount? 
Thanks for the fast action, btw 👍

RayOei:
Ps. is there a test key somewhere, so I can also try this?

RayOei (approved):
I can't replay this myself at this time but the change looks fine to me :-)

yksflip:
thanks for your quick response :) 
From a user perspective it's currently not possible to use the mollie plugin at all. with this possible it would be possible to use it with the setting "charge transaction fee" disabled.

RayOei:
@yksflip As I assume Roko.li (Steffen) will publish this patch: best to also set a new tag to this version. I don't have the permissions, I think?

PR #1184, opened 2025-05-16
Bugfix: Deleting article categories fails

### LINKED ISSUE #1175: [BUG]: Deleting article categories fails on main
**Describe the bug**
When trying to delete a article category an error is returned: 
`Category could not be deleted: Mysql2::Error: Unknown column 'articles.article_category_id' in 'where clause'
`
This seems to be a regression introduced with the addition of the article_version model @lentschi 

**Foodsoft Version**
What Version of Foodsoft are you running (can be found in the bottom-right). Are you running a modified Fork?
- main branch

lentschi:
*In theory* this fixes #1175 

(Associations in the `ArticleCategory` model had not been updated to the new `ArticleVersion` db structure.)

**However:** There has been a logical issue with `ArticleCategory` deletion - even in v4.9.1 (before the article unit upgrade). Steps to reproduce:

1. Create a new ArticleCategory 'TestCat'
2. Create a new Supplier 'TestSupplier123'
3. Create a new Article 'TestArticle123' in 'TestSupplier123' and set its category to 'TestCat'
4. Create a new Order for 'TestSupplier123'
5. Place a GroupOrder for this Order and 'TestArticle123' 
6. Close the Order
7. Delete 'TestArticle123' in the supplier's article list (now possible as the associated Order has been closed)
8. Delete 'TestCat' (now possible as all associated articles have been marked as deleted)
9. Open the closed order (e.g. through "Manage orders" -> "TestSupplier123" -> Show
-> Error: "undefined method name' for nil:NilClass" in /app/app/views/orders/_articles.html.haml:14 - `a.article.article_category.name` (The closed order's view tries to show the deleted Article, which would be possible as it has only been soft-deleted. However the associated ArticleCategory has been hard-deleted, so we see the error.)

I think the best fix for this would be to make ArticleCategory soft-deletable (same as `Article#mark_as_deleted`).
What do you think @henning-unicode?

Even if we do, I'd say we can continue using `check_for_associated_articles` in `ArticleCategory` as it would avoid not being able to select an category in the article edit form of a new article despite it still being visible as the current selection in an existing article.

**EDIT**

Fixes #1175 as well as the above mentioned issue - also see https://github.com/foodcoops/foodsoft/pull/1184#issuecomment-2902058794

### DIFF
--- a/app/controllers/article_categories_controller.rb
+++ b/app/controllers/article_categories_controller.rb
@@ -12,7 +12,8 @@ class ArticleCategoriesController < ApplicationController
   end
 
   def destroy
-    destroy!
+    resource.mark_as_deleted
+    redirect_to article_categories_path, notice: I18n.t('flash.actions.destroy.notice', resource_name: resource.class.model_name.human)
   rescue StandardError => e
     redirect_to article_categories_path, alert: I18n.t('article_categories.destroy.error', message: e.message)
   end
@@ -20,6 +21,6 @@ class ArticleCategoriesController < ApplicationController
   protected
 
   def collection
-    @article_categories = ArticleCategory.order('name')
+    @article_categories = ArticleCategory.undeleted.order('name')
   end
 end
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -3,7 +3,7 @@ class ArticlesController < ApplicationController
 
   before_action :load_article, only: %i[edit update]
   before_action :load_article_units, only: %i[edit update new create]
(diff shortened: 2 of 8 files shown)


lentschi:
As henning-unicode hasn't answered so far, what do you think @yksflip ?

lentschi:
> sound logical to me to use soft-delete in this case. lgtm

:+1: Done implementing soft-delete

lentschi:
> > sound logical to me to use soft-delete in this case. lgtm
> 
> 👍 Done implementing soft-delete

Meaning: Please review again :slightly_smiling_face:  (I've added code after your original review @yksflip )

yksflip (approved):

PR #1185, opened 2025-05-17
Bugfix: Receive order: can't add article to list of received articles

### LINKED ISSUE #1180: [BUG]: Receive order: can't add article to list of received articles
**Describe the bug**
It's not possible to add an additional article to the list of received articles.

**To Reproduce**
1. Go to 'Orders' -> 'Manage Orders'
2. Pick an order with status _closed_ and click on _Receive_
3. Below the article list click in the drop-down list _Add article_
4. Choose an article from the drop-down list
5. Result is an alert "A problem has occured.", no article is added to the list

**Expected behavior**
The chosen article is added to the article list above

**Screenshots**

Image

**Foodsoft Version**
v4.9.1

**Logfile**
n/a

**Additional context**
n/a

lentschi:
Fixes #1180

### DIFF
--- a/app/assets/javascripts/receive-order-form.js
+++ b/app/assets/javascripts/receive-order-form.js
@@ -113,15 +113,15 @@
         // TODO implement adding a new article, like in deliveries
       }).on('change', (e) => {
         var $input = $(e.target);
-        var selectedArticleId = $input.val();
-        if(!selectedArticleId) {
+        var selectedArticleVersionId = $input.val();
+        if(!selectedArticleVersionId) {
           return false;
         }
 
         $.ajax({
           url: this.newOrderArticlePath,
           type: 'post',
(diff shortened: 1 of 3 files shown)


henning-unicode:
We tested it and it fixed the bug! Thanks!

However we were wondering if the unit conversion button should also appear for newly added articles, currently it only appears for the articles that were already ordered.

lentschi:
@henning-unicode 

> However we were wondering if the unit conversion button should also appear for newly added articles, currently it only appears for the articles that were already ordered or after you save and go back.

Ah, you're right - I had missed that :+1: . Should [now be fixed](https://github.com/foodcoops/foodsoft/pull/1185/commits/772ab83eed066c47d923c0ba6e3df7d43dfe1733) too.

henning-unicode:
@lentschi great, it works now!

PR #1187, opened 2025-05-27
Add FTP support and B85 export for automated orders

### LINKED ISSUE #1126: Send orders to suppliers
At the moment you have to close an order, generate a pdf document or a fax and then send it manually to a supplier. We can automate this process by sending a order directly to the order system of the supplier.

robwa:
### Features Added

- **FTP Validation**: Added specific validation for orders to meet B85 requirements, including order number length, unit conversion, and quantity checks.
- **FTP Ordering**: Enabled sending of orders via FTP, based on supplier configuration. Email remains the default option.
- **Supplier Updates**: Introduced `remote_order_method` and `remote_order_url` fields to suppliers for supporting remote FTP ordering.
- **B85 Export**: Implemented B85 mailbox order export format for FTP orders, widely used by German organic wholesalers.

### Tests
- Unit tests added for validation, FTP ordering, and B85 formatting to ensure robustness of new functionality.

### Notes
- FTP order format is not configurable so far. It can easily be added as soon as we want to support other FTP ordering formats than B85.
- B85 export format lives beside the other export formats (CSV, PDF, etc.). Moving the export formats to a separate plugin can be thought of but is beyond the scope of this PR from my point of view.

Closes #1126.

### DIFF
--- a/spec/integration/order_spec.rb
+++ b/spec/integration/order_spec.rb
@@ -57,4 +57,55 @@ feature Order, :js do
     order.reload
     oa.reload
   end
+
+  context 'with registered order method' do
+    let(:user) { create(:user, groups: [create(:workgroup, role_orders: true)]) }
+    let(:supplier) { create(:supplier, article_count: 1, remote_order_url: 'ftp://user:pass@example.com/path') }
+    let(:order) { create(:order, supplier: supplier) }
+    let(:custom_method_name) { :custom_order_method }
+    let(:custom_formatter) do
+      Class.new do
+        def initialize(order)
+          @order = order
+        end
+
+        def to_remote_format
+          'ORDER_DATA'
+        end
+
+        def remote_file_name
+          'order.txt'
+        end
+      end
+    end
+    let(:ftp_mock) { instance_double(Net::FTP) }
+    let(:current_time) { Time.current }
+
+    before do
+      allow(Time).to receive(:now).and_return(current_time)
+      allow(Net::FTP).to receive(:open).and_yield(ftp_mock)
+      allow(ftp_mock).to receive(:login)
+      allow(ftp_mock).to receive(:putbinaryfile)
+
+      Supplier.register_remote_order_method(custom_method_name, custom_formatter)
+      supplier.update(remote_order_method: custom_method_name)
+    end
+
+    after do
+      Supplier.instance_variable_get(:@remote_order_formatters).delete(custom_method_name)
+    end
+
+    it 'can use registered order method' do
+      expect(supplier.remote_order_formatter).to eq(custom_formatter)
+    end
+
+    it 'successfully sends order to supplier via FTP' do
+      order.send_to_supplier!(user)
+
+      expect(Net::FTP).to have_received(:open).with('example.com')
+      expect(ftp_mock).to have_received(:login).with('user', 'pass')
+      expect(ftp_mock).to have_received(:putbinaryfile).with(anything, 'order.txt')
+      expect(order.remote_ordered_at.to_i).to eq(current_time.to_i)
+    end
+  end
 end


Gemfile
```diff
@@ -66,6 +66,7 @@ gem 'terser', '~> 1.1'
 
 # we use the git version of acts_as_versioned, and need to include it in this Gemfile
 gem 'acts_as_versioned', git: 'https://github.com/technoweenie/acts_as_versioned.git'
+gem 'foodsoft_b85', path: 'plugins/b85'
```
lentschi, line 69: This means the plugin is enabled by default, but [here](https://github.com/foodcoops/foodsoft/pull/1173#issuecomment-2995968924) we agreed it shouldn't be.
So I suggest this PR to be marked **draft** for now, lest anyone should review, right?

lentschi (changes requested):

Gemfile
robwa: I marked it as _draft_ as I'm still working on it.

Regarding enabled state: What actually means *enabled*? In Foodsoft, there are two ways to *disable* a plugin:
- soft disable: most plugins support a config flag (typically called `use_<plugin-name>`) which can be set to `false`
- hard disable: removing the plugin from `Gemfile`

I'm not sure if there are any hard-disabled plugins in Foodsoft yet. It would mean, we cannot run the tests easily.

robwa: Yes, very good, thank you for your answer. In fact, I had overlooked the deactivated plugins in Gemfile, sorry. So I moved the plugin to the test section, that seems to work fine.

app/models/order.rb
```diff
@@ -307,10 +307,37 @@ def close_direct!(user)
...
+
+    formatter_class = supplier.remote_order_formatter
+    raise "No formatter registered for remote order method: #{supplier.read_attribute_before_type_cast(:remote_order_method)}" unless formatter_class
+
+    local_temp_file = Tempfile.new
+    begin
+      local_temp_file.write(formatter_class.new(self).to_remote_format)
+      local_temp_file.rewind
+      # BE + 6-digit customer number + last 3 digits of order ID
+      remote_filename = "BE#{format('%06d', supplier.customer_number.to_i)}.#{format('%03d', id % 1000)}"
```
yksflip, line 331: I'm just wondering how this filename is made up? Is this part of the B85 spec?
I've a slight feeling that this rather belongs to the plugin or it should be somewhere configurable. But should be also okay for now.

plugins/b85/README.md
```diff
@@ -0,0 +1,26 @@
+# FoodsoftB85
+
```
yksflip, line 2: We recently decided about the plugin maintenance and agreed that non-default plugins should have a list of contact persons in the Readme file (See https://github.com/foodcoops/foodsoft/tree/master/plugins#plugin-maintenance)
Could you add a contact to this file whomever we/user can contact if the plugin does not work anymore as expected?

plugins/b85/foodsoft_b85.gemspec
```diff
@@ -0,0 +1,21 @@
...
+
+# Maintain your gem's version:
+require 'foodsoft_b85/version'
+
+# Describe your gem and declare its dependencies:
+Gem::Specification.new do |s|
+  s.name        = 'foodsoft_b85'
+  s.version     = FoodsoftB85::VERSION
+  s.authors     = ['Foodsoft Team']
+  s.email       = ['foodsoft@foodcoops.net']
```
yksflip, line 11: I think this should be rather your contact / unicodes?

yksflip (commented):
I didn't check the b85 spec/logic in detail, but it looks very good to me in general. I like that the remote_order_formatter is extensible now!
For the plugin it would be good if you could add your / a contact person from unicode. There is now direct obligation with that. But it would be just good to have someone to ask in case smth breaks in a couple of years ...
If @lentschi  doesn't have any other remarks this would be ready to merge for me :)

spec/models/order_spec.rb
```diff
@@ -187,4 +187,53 @@
...
+      let(:order) { create(:order, supplier: supplier) }
+
+      before do
+        supplier.update_column(:remote_order_method, :unknown_method)
+        supplier.reload
+      end
+
+      it 'raises an error when no formatter is registered' do
+        expect { order.upload_via_ftp }.to raise_error(RuntimeError, /No formatter registered for remote order method: unknown_method/)
+      end
```
lentschi, line 236: Maybe it would make sense to add another `context` '**with** registered order method' here?
(I know, that developers can see a [specific example in the plugin's spec](https://github.com/foodcoops/foodsoft/pull/1187/files#diff-e6334318218698b583a81cb25c43de351ebfa89482d9386f2517887c7b5a1935R21), however - since the plugin is non-default, it might make sense to have an abstract sample/test here too...?)

app/models/order.rb
robwa: Good catch, changed with 436231c.

yksflip (approved):

yksflip:
@lentschi  your vote is stll "requested changes", can you change this to approve or is there still smth. that should be adapted?

lentschi:
> @lentschi your vote is stll "requested changes", can you change this to approve or is there still smth. that should be adapted?

Sorry, I was on holidays - changing it now as the two remaining issues have been resolved in the mean time. :slightly_smiling_face:

lentschi (approved):

Issue #1007, opened 2023-06-12
Include a maximum number of articles to be sold 


yksflip:
feature request by [ardjan](https://forum.foodcoops.net/u/ardjan)

> the Adam fork already included an option to include a maximum number of products to be sold in the regular product upload/ordering system. Now we are improvising with the Stock functionality to achieve the same functionality, but that is quite cumbersome.

Should be already in here somewhere: https://github.com/foodcoop-adam/foodsoft

yksflip:
https://github.com/foodcoop-adam/foodsoft/commit/0ac5c9c37b6af78f922ff86e6b3a302ac505c2ed

PR #1201, opened 2025-07-09
feat: maximumx order quantity for order article

yksflip:
implements #1007

### DIFF
--- a/app/assets/javascripts/application_legacy.js
+++ b/app/assets/javascripts/application_legacy.js
@@ -18,6 +18,7 @@
 //= require_self
 //= require big
 //= require units-converter
+//= require validation-utils
 //= require migrate-units-form
 //= require article-form
 //= require unit-conversion-field
--- a/app/assets/javascripts/article-form.js
+++ b/app/assets/javascripts/article-form.js
@@ -13,6 +13,7 @@ class ArticleForm {
       this.supplierUnitSelect$ = $(`#${this.unitFieldsIdPrefix}_supplier_order_unit`, this.articleForm$);
       this.unitRatiosTable$ = $('#fc_base_price', this.articleForm$);
       this.minimumOrderQuantity$ = $(`#${this.unitFieldsIdPrefix}_minimum_order_quantity`, this.articleForm$);
+      this.maximumOrderQuantity$ = $(`#${this.unitFieldsIdPrefix}_maximum_order_quantity`, this.articleForm$);
       this.billingUnit$ = $(`#${this.unitFieldsIdPrefix}_billing_unit`, this.articleForm$);
       this.groupOrderGranularity$ = $(`#${this.unitFieldsIdPrefix}_group_order_granularity`, this.articleForm$);
       this.groupOrderUnit$ = $(`#${this.unitFieldsIdPrefix}_group_order_unit`, this.articleForm$);
@@ -60,7 +61,7 @@ class ArticleForm {
       const tax = parseFloat(this.tax$.val());
       const deposit = parseFloat(this.deposit$.val());
       const grossPrice = (price + deposit) * (tax / 100 + 1);
-      const fcPrice = grossPrice  * (this.priceMarkup / 100 + 1);
+      const fcPrice = grossPrice * (this.priceMarkup / 100 + 1);
       const priceUnitLabel = this.getUnitLabel(this.priceUnit$.val());
       this.fcPrice$.find('.price_value').text(isNaN(fcPrice) ? '?' : I18n.l('currency', fcPrice));
       this.fcPrice$.find('.price_per_text').toggle(priceUnitLabel.trim() !== '');
@@ -88,7 +89,7 @@ class ArticleForm {
         this.loadRatios();
         this.undoPriceConversion();
         this.undoOrderAndReceivedUnitsConversion();
-      } catch(err) {
+      } catch (err) {
         e.preventDefault();
         throw err;
       }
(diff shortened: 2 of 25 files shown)


yksflip:
> I tried it out locally and got an unhandled error right away:
I fixed it by adding a null check.

> Also, I think there's (at least) some logic missing in `app/assets/javascripts/article-form.js` 
Indeed! I've added it almost identitcal to max-quantity ..

> and `app/assets/javascripts/group-order-form.js`
I think that is not needed as I used the `quantity_available` in group_order.rb ( https://github.com/foodcoops/foodsoft/pull/1201/files#diff-b946eb642b5614e4a0519a74fc97aec4c0bcec98db9c99c790ea7c048349b5acR63 ) that was already implemented for the stock-order feature I think

lentschi:
> I think that is not needed as I used the quantity_available in group_order.rb ( https://github.com/foodcoops/foodsoft/pull/1201/files#diff-b946eb642b5614e4a0519a74fc97aec4c0bcec98db9c99c790ea7c048349b5acR63 ) that was already implemented for the stock-order feature I think

Ah, right, seems to work :+1: 

But maybe we should also change the error message when entering an invalid value manually (exceeding max):
image
(It now only mentions min and granularity)

app/assets/javascripts/article-form.js
```diff
@@ -204,6 +206,28 @@ class ArticleForm {
       this.updateCustomUnitWarning();
     });
     this.onSupplierUnitChanged();
+
+    // Add validation for minimum/maximum order quantity
+    this.initializeOrderQuantityValidation();
```
lentschi: Why did you add client side validation to this form? I think since other form fields are validated by the backend, this makes this kind of inconsistent. If I remove that line, it works just as fine - after clicking "Update article", I see the following error:

image

Of course frontend validation is always a nice-to-have in terms of usability, but I was hoping to only add this when building a PWA style client to not further complicate the already ugly jquery code. What do you think?

app/assets/javascripts/validation-utils.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+
+
+    /**
+     * Sets up validation for a form input
+     * @param {jQuery} input$ - The input element
+     * @param {Object} options - Validation options
+     */
+    static setupInputValidation(input$, options = {}) {
```
lentschi: Method seems to be unused - AI slop?

app/assets/javascripts/validation-utils.js
```diff
@@ -0,0 +1,155 @@
...
+                errorSpan$.hide();
+            }
+            input$.removeClass('validation-error');
+        }
+    }
+
+    /**
+     * Hides all validation messages
+     */
+    static hideAllValidationMessages() {
```
lentschi: Method seems to be unused - AI slop?

app/views/group_orders/_form.html.haml
```diff
@@ -116,6 +116,7 @@
             - quantity_data['used_quantity'] = @ordering_data[:order_articles][order_article.id][:used_quantity]
             - quantity_data['price'] = @ordering_data[:order_articles][order_article.id][:price]
             - quantity_data['minimum_order_quantity'] = @ordering_data[:order_articles][order_article.id][:minimum_order_quantity] unless @ordering_data[:order_articles][order_article.id][:minimum_order_quantity].nil?
+            - quantity_data['maximum_order_quantity'] = @ordering_data[:order_articles][order_article.id][:maximum_order_quantity] unless @ordering_data[:order_articles][order_article.id][:maximum_order_quantity].nil?
```
lentschi: If I'm not mistaken, `@ordering_data[:order_articles][order_article.id][:maximum_order_quantity]` is never set, right? (At least the above fields are all set in [GroupOrder.load_data](https://github.com/foodcoops/foodsoft/blob/d262291ce48447dc404c99ce129c6f3b796efb70/app/models/group_order.rb#L60), but maximum_order_quantity isn't.)

app/assets/javascripts/application_legacy.js
```diff
@@ -18,6 +18,7 @@
 //= require_self
 //= require big
 //= require units-converter
+//= require validation-utils
```
lentschi: I think the AI added some unused methods for these validation-utils (I added review comments for where I seem to have found dead code) and also the solution it provided supersedes some old logic, which it also failed to remove (e.g. the `%span.numeric-step-error` in the group order form template no longer seems to be required).

Instead of reviewing the AI code in depth (which seems cumbersome, as I think there's quite some AI slop), here's my quick go on this:

https://github.com/foodcoops/foodsoft/pull/1232/files#diff-ade87a685fd4a0e008033752780385cce4bbfff5d1355fab4b35cafffc2b3080R44

lentschi (changes requested):
> I've added backend and client side validations for this now.
The javascript client-side validation were quite complicated, good old friend ai helped a little bit out here. Not sure if it needs to be a custom validation like this. see app/assets/javascripts/validation-utils.js
maybe there is a better / more idiomatic way to do this?

I've added some code comments.

In addition, two general issues I'd like to throw in:

1. Currently when two users open the group order form at the same time, then one adds a quantity that is below `maximum_order_quantity` and the second user tries the same but it's exceeding `maximum_order_quantity` when _combined_ with the first user's quantity, the second user just gets a "The order couldn’t be updated due to a bug." So maybe add some custom error flash [here](https://github.com/foodcoops/foodsoft/blob/d262291ce48447dc404c99ce129c6f3b796efb70/app/controllers/group_orders_controller.rb#L38) if the `maximum_order_quantity` is exceeded? (It would be better to have the form error inside the form, but I guess that's more difficult to implement.)
2. I haven't tried, but have you checked, if "using up" tolerance, can't be causing to somehow exceed `maximum_order_quantity`?

PR #1203, opened 2025-07-25
Fix/1178 mollie plugin

RayOei:
Fixes #1178

- Update for changed Mollie API call (permanent fix for hotfix ec170719)
- Fix handling of fee which did not include tax
- Add tax for Mollie fee to configuration
- Update maintainer info
- Bump Mollie API to v4.17
- Fix missing translations
- Improve Mollie return information
- Add more logging for better debugging (also in production)
- Add translation for payment status
- Add mechanism for local endpoint testing, without the need to change code
- Set Mollie plugin version to 1.0.1

Refactoring handling of payment methods - Only iDEAL supported at this stage, more to follow

PR #1206, opened 2025-07-26
V4.9.2

### LINKED ISSUE #1181: [BUG]: Message Group views are missing translations
**Describe the bug**
These translations for message group views are missing in all languages:

- admin.messagegroups.index.title 
- admin.messagegroups.index.first_paragraph
- admin.messagegroups.index.second_paragraph
- admin.messagegroups.index.new_messagegroup
- admin.messagegroups.index.title

**To Reproduce**
1. Go to 'Administration' -> 'Message groups'
2. Check view description, it says "First Paragraph" and "Second Paragraph"
3. Check invalid page title 
4. Check missing button label

**Expected behaviour**

- Views display correct translation values instead of dummy/no label

**Screenshots**

Image

**Foodsoft Version**
v4.9.1

**Logfile**
n/a

**Additional context**
n/a

RayOei:
Bumped to 4.9.2

- Layout improvements and fixes in transaction overview
- Added missing translation entries
- Improves #1178
- Includes hot fix https://github.com/foodcoops/foodsoft/commit/ec170719dd60b76e4ec32fc213465d6976b01b16

Note: merging to main release branch/trunk for future `v4.x` releases.

Update:

Cherry-pick #1169
Cherry-pick #1178
Cherry-pick # 1166
Fix #1179 
Fix #1181 
Fix #1080


RayOei:
@twothreenine @lentschi FYI: see https://github.com/foodcoops/foodsoft/discussions/1199#discussioncomment-13606271 and onwards on the juggling of branches. 
Goal is to end-up with a `v4.x` branch/trunk for future `v4` releases.
PR #1203 is also targeted to merge onto `v4.x`.

RayOei:
Cherry picking is done separately. I am reviewing those as we speak 😁 As soon as I re-push that branch, I encountered an issue, I will add the commits as well. Only a few but some do bother the FCAN users at this time. As long as v4 is not abandonded adding (minor) improvements is not a bad thing, in my book anyway.
The branches discussion was about being able to clearly separate `v4.x` and `v5.x`. As there were a few different ideas floating around in branches in the past, it got a bit messy. We settled now on using the `v4.x` one for future updates on `v4`. 

Part of the fixing/improving here was also meant as a way to get a bit better acquinted with the code. It is not imperative but does provide a few minor tweaks. I did find an issue yesterday, I left a Mollie specific item in there by accident, which I am going to address soon. Again also part of me getting to understand the plugin idea a bit better. 

As a side note: while cleaning up the release4 confusion I also killed the `Mollie` PR as part of the cleaning. I will reintroduce that soon, as that does provide a needed fix for payments and missing fees.

RayOei:
@lentschi Any idea why this spec is failing? It doesn't locally.

RayOei:
1) no this should be the 'final' product
2) This is a follow-up version from 4.9.1 with some fixes and improvements. As Git shows it is a continuation of 4.9.1 which will merge back to 4.x.
3) This was earlier discussed with yksflip as there were originally two similar, already existing, 4.x branches and this was chosen as a way to clean up. Which would mean having a 4.x branch and a 5.x (which is master).

As for the test failing: "Order sends mail if min_order_quantity has been reached". It doesn't fail locally. It is part of the code I haven't looked into so I don't yet know what the flow should be and how this could fail. But it is somewhat confusing that it all passes when run locally. That should not be different, unless there is a reason it is and I missed that?

lentschi:
@RayOei 

> This is a follow-up version from 4.9.1 with some fixes and improvements

What exactly are those "some fixes and improvements" you're referring to here? (What was the process to pick them and who picked them?)

Let me mention one random example:
https://github.com/foodcoops/foodsoft/pull/1078 seems to be in `master` and in `v4.9.2`.
It's in `master` because I myself reviewed and tested it for 5.0.
But who did that for 4.x.?

IMO 4.x should _only_ contain **urgent** hotfixes (and I don't think the sample I picked is one of them - I think that's rather a nice-to-have).
Everything else should only be released in 5.0 - that's where all our focus should be right now!

(And even if we did pick some commits into 4.x, then it's not a good idea to do it in a bulk PR. It should be separate cherry-pick PRs then.)

lentschi:
@RayOei I had to remove the branch `v4.9.2` (which automatically closed this PR), because I did a real v4.9.2 release.
Branches never should have potential release version tags' names.

I copied the branch's reference to the new branch [bulk-changes-for-4.9.x](https://github.com/foodcoops/foodsoft/tree/bulk-changes-for-4.9.x) so you don't lose your work. However, [I still think doing it all in bulk isn't a good idea](https://github.com/foodcoops/foodsoft/pull/1206#issuecomment-3896236547).

Issue #1002, opened 2023-05-26
Upgrade to Ruby 3.x


yksflip:
as mentioned in https://github.com/foodcoops/foodsoft/pull/979 it's already time to upgrade the ruby version again ... ruby 2.7 is now [EOL](https://www.ruby-lang.org/en/news/2023/03/30/ruby-2-7-8-released/)

yksflip:
* `less-rails` depends on `therubyracer` and seems not to support the newer `mini_racer` gem
   see: https://github.com/metaskills/less-rails/issues/160

yksflip:
ahh there's an issue for that already:  #928 
I'll give it a shot and try to convert the `.less` files to `.scss`

yksflip:
I found this for bootstrap: https://github.com/twbs/bootstrap-sass
But I think we have to upgrade to v3, in v2 smth seems to break with sass-rails and other gems ...

I know upgrading bootstrap sounds terrible, but I did a quick and dirty run and I think it's doable. Just touch every single view :o

sneak peek:

![Image](https://github.com/foodcoops/foodsoft/assets/16109235/8f6dfa91-66d0-4f6e-bf18-f2283e874502)

RayOei:
Based on [this](https://www.ruby-toolbox.com/categories/Active_Record_Versioning) it seems that [Paper_trail](https://github.com/paper-trail-gem/paper_trail) (and its [gem repo](https://rubygems.org/gems/paper_trail)) is de-facto a standard by the looks of it. I don't know what the differences are, though.

henning-unicode:
@RayOei Thanks! I think for just getting to Ruby 3 we should keep acts_as_versioned as a new gem would need us to cover a migration path for already created wiki pages with acts_as_versioned.

RayOei:
> getting to Ruby 3 we should keep acts_as_versioned 

Sure 👍 
I was merely looking for alternatives if the old gems would be a problem. It may still be useful later as being dependent on a gem which in reality hasn't changed in 13 years (or so) is likely not the best choice moving forward. 
But that should be a new issue for later to resolve 😁

PR #1208, opened 2025-08-02
On #1002: Fixes search reset button design issues

lentschi:
see https://github.com/foodcoops/foodsoft/pull/1146#discussion_r2231139003 - specifically:

> TODO Search Button breaks after reset: Seems to be caused by the call list.search('') in list.reset.js as it removes the btn class from the button. I have no idea how listjs works, why it removed the class and how this can be fixed. Maybe someone can help out?

This seems to fix it for me.


henning-unicode:
@lentschi great thanks!

PR #1233, opened 2025-10-31
Closes #1108 - Create api v2 supporting new article (version) fields

### LINKED ISSUE #1108: Create api v2 supporting new article (version) fields
see https://github.com/foodcoops/foodsoft/commit/ed9192c

lentschi:
- [X] Create API version 2
- [x] Move shared_suppliers endpoint to new API v2
- [x] Copy old endpoints to API v2 as well, but adapt them to support new article unit fields replacing old fields like `unit_quantity`

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -321,7 +321,7 @@ Naming/MemoizedInstanceVariableName:
 # AllowedNames: as, at, by, cc, db, id, if, in, io, ip, of, on, os, pp, to
 Naming/MethodParameterName:
   Exclude:
-    - 'app/controllers/api/v1/base_controller.rb'
+    - 'app/controllers/api/base_controller.rb'
     - 'app/controllers/api/v1/user/group_order_articles_controller.rb'
     - 'app/helpers/application_helper.rb'
     - 'app/helpers/orders_helper.rb'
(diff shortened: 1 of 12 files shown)


twothreenine:
I'll share the link to this preview build in the Austrian forum so people can take a look at the new unit system, since the master build doesn't work.

lentschi:
> I'll share the link to this preview build in the Austrian forum so people can take a look at the new unit system, since the master build doesn't work.

Huh - and this branch build works - interesting. :thinking: 
Still, I wonder if it's not a bit risky to share the link as long we haven't figure what's causing the issues on master - there might be some some things failing in this branch too.

henning-unicode:
@lentschi are you looking for someone to review this? I think I could do it.

lentschi:
>  are you looking for someone to review this? I think I could do it.

@henning-unicode Yes, _please_ - that would be great!! :pray:

lentschi:
@henning-unicode 

> will try to review ASAP, probably some time next week

Sure, whenever you find the time. This has been lying around since the 1st of November, so no hurry. :slightly_smiling_face:

RayOei:
Well.. it does to some extent: when trying to move to this PR I have to migrate and that breaks. But that will be true for every PR on v5. Either way: reading up on migrations I think this is not a scenario that ActiveRecord supports. It is too late to add a repair migration as some fields no longer exist with v5.

PR #1246, opened 2025-11-12
Resize input fields in group order form

twothreenine:
I tried to make the quantity & tolerance input fields wider, couldn't test it locally though.
Deploy to server failed, too (`denied: installation not allowed to Write organization package`)


twothreenine:
> I think a 'decimal' class does not exist

I looked into browser tools and noticed that in the article edit form, some input fields like _price_ use the class `decimal`. I'm not sure what it does.
(from 4.9)
grafik
`<input class="numeric decimal required input-mini" required="required" aria-required="true" type="number" step="any" value="11.18" name="article[price]" id="article_price">`

Anyway, I mainly wanted to test whether the test deployment works for new PRs, so I started this tiny draft PR.

Issue #1225, opened 2025-10-10
[BUG]: Multiple styling issues after bootstrap upgrade


lentschi:
**Description**
After the upgrade of bootstrap in https://github.com/foodcoops/foodsoft/pull/1146, some styling issues turned up:

| no. | View | Before | After | Remarks |
| ------ | ------ | ------ | ----- | --- |
| 1. |**Article form** (in articles index, or when editing in receive or accounting) | ![image](https://github.com/user-attachments/assets/5e55395b-3756-46f7-81db-10898c0be687) | <img width="604" height="850" alt="Image" src="https://github.com/user-attachments/assets/722e554f-db6d-4125-9508-b68adc983067" />  |  On some rows, the label is not on the same level as the value (e.g. "FoodCoop price") and sometimes the columns are not aligned with the others (e.g. "Name" compared to "Delivery unit"). The modal header ("Edit article") seems a bit off too (compare with the [bootstrap demo](https://getbootstrap.com/docs/4.0/components/modal/): There the title and the closing icon are on the same y-axis.) |
| 2. | **Edit all articles** |  ![image](https://github.com/user-attachments/assets/d0269fd7-fb84-4636-b285-d8a7d26ad650) | <img width="717" height="395" alt="image" src="https://github.com/user-attachments/assets/09bad934-5b08-4e0a-b516-ef8affa231e4" /><img width="905" height="166" alt="image" src="https://github.com/user-attachments/assets/3e6850f5-f433-41b1-a87b-fbb9b523cc51" /> | Fields in units popover misaligned; cog icons in index misaligned |
| 3. | **Create new order** | ![image](https://github.com/user-attachments/assets/f4f9bb6f-9795-49cb-baee-e3d2b119a01f) | ![image](https://github.com/user-attachments/assets/bf2a6bfc-9460-469f-8a90-ed5f7fe95a38) | Some of the header form fields should probably be on the same line to save space when trying to scroll down to the articles etc. (**Note:** This is just my opinion. If you disagree, please turn it into a discussion.) |
| 4. | **Orders view - sorted in groups** as well as **Accounting with expanded article**  | ![image](https://github.com/user-attachments/assets/2b6256bc-4c29-4aee-8506-84063acfef8f) | ![image](https://github.com/user-attachments/assets/fbdd0030-1458-48c4-9cb0-722c587f4d50) | Quantity field layout broken  |
[...]

PR #1247, opened 2025-11-14
fix: Multiple styling issues after bootstrap upgrade

henning-unicode:
Fixes some of the issues in #1225 
- fixes quantity, tolerance and delta input widgets
- fixes article-info hover in group-order edit view

It also includes a small fix for a button group with a dropdown toggle which had two different colors


lentschi (approved):
> @lentschi I wrote it in the issue:
>>  https://github.com/foodcoops/foodsoft/pull/1247 should hopefully fix issues 4-6 in the table

Ah, sorry - I had missed that.

Issues 4-6 are now fixed according to the descriptions in the issue table, yes! 👍 

(There are still some related issues left IMO, like the text in the quantity/tolerance inputs being too small or the Euro sign line-wrapping unexpectedly, but those can be fixed in a separate PR -> I'll approve this one.)

PR #1256, opened 2025-11-28
Fix styling issues after rails and bootstrap upgrade

lentschi:
Fixes multiple styling issues after the rails and bootstrap upgrade in https://github.com/foodcoops/foodsoft/pull/1146.

This PR includes fixes for the remaining issues mentioned in https://github.com/foodcoops/foodsoft/issues/1225, but also includes others, that were not listed (I didn't want to go into the effort of doing before/after screenshots for everything.)

Please look into the commit messages - I tried to make them as verbose as possible to explain what I was fixing.

The article form columns should now align properly. However I think the left column (containing the labels) is a bit to small - at least for German texts, which is why I added auto-hyphenation. Making the labels wider, would be better but that's a mere nice-to-have.

I'm aware that many of those fixes are a bit hacky and the UI is still _far_ from perfect - but perfection is something we will only achieve by finally replacing the old jquery UI completely, which is too much to do for the 5.0 release.

### DIFF
--- a/app/helpers/orders_helper.rb
+++ b/app/helpers/orders_helper.rb
@@ -134,18 +134,20 @@ module OrdersHelper
                                                   data: data,
                                                   disabled: order_article.result_manually_changed?,
                                                   autocomplete: 'off'
-    span_html = if order_article.result_manually_changed?
-                  content_tag(:span, class: 'input-prepend input-append intable d-flex',
-                                     title: t('orders.edit_amount.field_locked_title', default: '')) do
-                    button_tag(nil, type: :button, class: 'btn unlocker') {
-                      content_tag(:i, nil, class: 'glyphicon glyphicon-lock')
-                    } + input_html
-                  end
-                else
-                  content_tag(:span, class: 'btn-group numeric-step d-flex') { input_html }
-                end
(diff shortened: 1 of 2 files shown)


app/helpers/orders_helper.rb
```diff
@@ -135,10 +135,10 @@ def receive_input_field(form)
                                                   disabled: order_article.result_manually_changed?,
                                                   autocomplete: 'off'
     span_html = if order_article.result_manually_changed?
-                  content_tag(:span, class: 'input-prepend input-append intable',
+                  content_tag(:span, class: 'input-prepend input-append intable d-flex',
```
henning-unicode: `input-prepend` and `input-append` classes don't exist in Bootstrap 3 anymore. They have been replaced by `input-group-addon` see: https://getbootstrap.com/docs/3.3/components/#input-groups Just something I noticed, not sure these classes make sense here.

lentschi:
@henning-unicode Thanks for your first review, I've responded to one of your comments and merged your PR on mine resolving the other -> Please re-review :)

henning-unicode:
@lentschi changes look good! I've resolved the last open comment. I think this is good to be merged.

PR #1270, opened 2026-02-06
Fix/1178 mollie plugin

### LINKED ISSUE #1178: [BUG]: Mollie API doesn't contain pricing information anymore

RayOei:
Fixes #1178

Most importantly:  bump Mollie API to v4.17 as previous version is no longer supported by Mollie
Update for changed Mollie API call (permanent fix for hotfix https://github.com/foodcoops/foodsoft/commit/ec170719dd60b76e4ec32fc213465d6976b01b16)
Fix handling of fee which did not include tax
Add tax for Mollie fee to configuration
Refactoring handling of payment methods - Only iDEAL supported at this stage, more to follow
Improve Mollie return information
Fix missing translations
Add more logging for better debugging (also in production)
Add translation for payment status
Add mechanism for local endpoint (with SSL/TLS) testing, without the need to change code
Set Mollie plugin version to 1.0.1
Added language schema
Update maintainer info
Updated plugin readme

Note: Although this should work on its own, the better overall Mollie result is achieved together with #1206

### DIFF
--- a/config/locales/nl.yml
+++ b/config/locales/nl.yml
@@ -69,7 +69,7 @@ nl:
         mime: MIME-type
         name: Naam
       financial_transaction:
-        amount: Transactie bedrag
+        amount: Transactie
         created_on: Datum
         financial_transaction_class: Financiële-transactieklasse
         financial_transaction_type: Financiële-transactietype
--- a/plugins/mollie/README.md
+++ b/plugins/mollie/README.md
@@ -1,23 +1,32 @@
-FoodsoftMollie
-==============
+# FoodsoftMollie
 
-This project adds support for various online payment methods using Mollie to Foodsoft.
 
-* Make sure the gem is uncommented in foodsoft's `Gemfile`
-* Enter your Mollie account details in `config/app_config.yml`
+This plugin adds support for various online payment methods to Foodsoft, using [Mollie](https://www.mollie.com), a Dutch payment provider who offers this service in the European Economic Area (EEA).
+
+> Currently, `v1.0.1` ONLY iDEAL payments are supported
+
+## Setup
+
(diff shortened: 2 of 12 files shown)


lentschi (changes requested):
I'm only writing because no other reviews seem to be forthcoming. 🤔 

I'm glad to help with any urgent issues (and this seems to be one?), but since Mollie isn't used in Austria, I don't have any experience with it and so a different reviewer should dive into it.

But here are still some basic issues I see with this PR:

1. If I'm not mistaken, this should be merged into `v4.x`, not `v4.9.2`. I know there's an [ongoing discussion about this,](https://github.com/foodcoops/foodsoft/discussions/1199) , but until a decision is reached there, we should stick with the way it has been handled so far and that seems to be: Merge hotfixes for the last release (`4.9.1`) into the branch `v4.x`, then create release tags in that branch.
2. It seems like you change a lot of different things in a single PR. Again I don't know anything about Mollie, so this might be fine. But generally, if possible, it would be better to create several smaller PRs out of it, making them easier to review and test. (Are all those changes really required only to fix https://github.com/foodcoops/foodsoft/issues/1178 or are there other changes piggybacking?)

PR #1274, opened 2026-02-09
fix(grouporder): improves order footer for mobile devices

### LINKED ISSUE #676: Still unable to save an order on mobile device
Issue #665 reported the order button to be hidden on mobile devices.

PR #668 suggested a solution which works but still leaves space for errors e.g. after clicking on a single product row or after adding or removing a product to the cart:

image

image

MuffinFox:
Since the overlapping of article info with order button on group-order page is still a very annoying bug when using foodsoft on mobile devices, this PR is a quick fix for small devices.

It fixes display: none of article-info for small devices, as already implemented, but in wrong order. 
Additionally it aligns order sum + button left for easier access.

URL:
/demo/group_orders/new?order_id=1

Fixes #676

### DIFF
--- a/app/assets/stylesheets/bootstrap_and_overrides.css.scss
+++ b/app/assets/stylesheets/bootstrap_and_overrides.css.scss
@@ -230,14 +230,6 @@ table {
   width: 200px;
 }
 
-/* Hide the orders article info for small screens
-   to prevent the "save order" button to disappear */
-@media only screen and (max-width: 950px) {
-  tr.order-article:hover .article-info {
-    display: none;
-  }
-}
-
 #order-footer {
   width: 100%;
   right: 0;
@@ -272,6 +264,21 @@ tr.order-article .article-info {
 
 tr.order-article:hover .article-info {
   display: block;
+}
+
+
+/* Hide the orders article info for small screens
+   to prevent the "save order" button to disappear */
+@media only screen and (max-width: 950px) {
+  tr.order-article:hover .article-info {
+    display: none;
+  }
+
+  #order-footer {
+    #total-sum {
+      float: left;
+    }
+  }
 }
 
 tr.order-article {


lentschi (commented):
Thank you for your PR, @MuffinFox!
However, I think this might be better solved by [this solution](https://github.com/foodcoops/foodsoft/pull/1276) - Also see [the issue](url) I now raised for this.
What do you think?

(Needless to say, this is also just a quick patch and the whole frontend urgently needs to be remastered - see  https://github.com/foodcoops/foodsoft/discussions/1265 on that)

MuffinFox:
Yes, its just a quick fix to be able to submit orders on mobile. i agree that the frontend needs a rewrite.
I tried your solution and focuswithin was not working for me on android/chrome.

A minor change i also addressed in my PR is that i moved the button to the left on mobile, otherwise its is a bit lost.

lentschi:
> @MuffinFox
> 
> > I tried your solution and focuswithin was not working for me on android/chrome.
> 
> That's odd. 🤔 For me it works on Android with both, Vanadium (Chromium) 145 as well as Fennec (Firefox) 147. Also MDN states that this _should_ work with current versions of all major browsers:
> 
>     * `:focus-within`: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:focus-within#browser_compatibility
> 
>     * `pointer: coarse/fine`: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/pointer#browser_compatibility
> 
> 
> Are you sure that when you tried, the buttons/inputs were _actually_ focused? (Maybe you're trying this on a fork / older version of foodsoft, that's still using [preventDefault](https://github.com/foodcoops/foodsoft/pull/1276/changes#diff-ade87a685fd4a0e008033752780385cce4bbfff5d1355fab4b35cafffc2b3080L32) or similar preventing focus on tap?)

@MuffinFox As you haven't replied to my question for a few weeks and [my PR](https://github.com/foodcoops/foodsoft/pull/1276) supersedes this one, I'm closing it. If however you can still confirm my PR is not working and/or yours is better for some reason, feel free to reopen and mention me! (I can still revert mine.)

PR #1287, opened 2026-03-05
Add qr-code link to supplier invoices sum in unpaid invoices view

mjavurek:
An URL  is added to the amount of individual invoices and to the sum of invoices of each supplier in the unpaid invoices view that leads to a qr-code that can be used in the banking app to pay these invoices.

### DIFF
--- a/app/views/finance/invoices/unpaid.html.haml
+++ b/app/views/finance/invoices/unpaid.html.haml
@@ -13,6 +13,7 @@
       - invoices_text << invoice.number
       = link_to finance_invoice_path(invoice) do
         = format_date invoice.date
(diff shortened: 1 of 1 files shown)


twothreenine:
Have you checked what happens when a too long info (reference text) is submitted? Will it be truncated or rejected entirely?

twothreenine:
It would be quite important to be able to pay single invoices or some, but not all invoices from a supplier.

In case there are any invoices that are not ready to be paid (created but not received yet, still need to be checked, etc.)
(Btw, it would also be great to have a boolean field "ready to be paid" for each invoice, and show a warning when selecting a not-ready invoice for payment.)

PR #1291, opened 2026-04-02
Make v4.x -> v5.x migrations more tolerant to existing db issues

lentschi:
During my experiments migrating real life foodsoft dbs I stumbled across three issues:

1. Precision of `units_to_order`, `units_billed`, `units_received` and `price`  was not big enough to hold old values -> increased it from 11 to 12.
2. When multiplying `price` with `unit_quantity` (new `price` value is always stored in supplier order unit), some crazy values still wouldn't fit -> limit those with 999999 as nothing a foodcoop supplier sells will ever be _that_ expensive.
3. When setting `Supplier.supplier_category_id` to non-nullable (see https://github.com/foodcoops/foodsoft/pull/1143 ), ensure that existing nullable fields are first updated with the first category in the db.


lentschi:

Issue #1271, opened 2026-02-07
Order placement menu: Adapt units/amounts in footer to new unit system


lentschi:
### Discussed in https://github.com/foodcoops/foodsoft/discussions/1236

**TODO:** Clarify requirements (Especially what is the minimum to work for the 5.0 release)

PR #1295, opened 2026-04-04
On #1271 Fix superfluous multiplier symbol before SI units

lentschi:
Fixes a minor design issue with #1271: SI units shouldn't be prefixed with the multplier symbol.
E.g. `12 × piece`, but **not** `12 × kg` (Just `12 kg` is enough).


lentschi:

PR #1298, opened 2026-04-24
Optionally display demo mode warning in header

### LINKED ISSUE #1297: [FEAT]: Optionally display demo mode warning in header
A new optional app config option `display_demo_warning` should be added (default: `false`).
If set to `true`, a warning ("demo mode!") will be displayed right next to the foodsoft logo.

**Intention:** Users working in a demo instance should be made aware of this at all times, so they are not mislead thinking they're on a real live instance.

**Note:** This new setting should not change anything else about how foodsoft is running. After all, it is _already_ possible to run foodsoft in a way that is essentially a demo mode, simply by configuring it in a certain way (e.g. having `config.action_mailer` point to a mailcatcher instance instead of sendmail). This issue is only about having a means to inform users about the _purpose_ of the instance.

**Example image:**

Image

lentschi:
Implements #1297 - Optionally display demo mode warning in header

### DIFF
--- a/app/assets/stylesheets/bootstrap_and_overrides.css.scss
+++ b/app/assets/stylesheets/bootstrap_and_overrides.css.scss
@@ -72,6 +72,14 @@ $nonessentialDim: 35%;
     color: #ffffff;
     background-color: $mainRedColor;
   }
+
+  .demo-mode {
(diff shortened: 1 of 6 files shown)


lentschi:

PR #1299, opened 2026-05-20
fix: remove double dutch translation of unece unit

Fitog4:
Uploading article files containing article units that appear more than once in the UNECE unit translation files causes a the upload to crash. We noticed this problem when trying to upload a file using the "zakje" unit, which currently appears as translation for both XPO and XSH.
We propose to replace the translation of one of the two with the term "builtje", which we see being used by retailers in the Netherlands.

### DIFF
--- a/config/units-of-measure/locales/unece_nl.yml
+++ b/config/units-of-measure/locales/unece_nl.yml
@@ -44,7 +44,7 @@ unece_units:
   XCN:
     name: container
   XPO:
-    name: zakje
+    name: builtje
   XSH:
     name: zakje
   XCB:


lentschi (approved):

lentschi:
Ah, thanks for uncovering and fixing that, @Fitog4!

I raised a bug issue to properly address this in the import algorithm itself:
https://github.com/foodcoops/foodsoft/issues/1300

(I think even if there are duplicate unit names in the translations, the import shouldn't fail. It should just use the first UNECE unit it can map. Of course that's not the best solution either, but I don't see a better way given the current CSV format: It's used as a means of importing/exporting part of the DB, but _at the same time_ serves as a human-readable spreadsheet representation of the supplier's articles, which is why we can't just put the raw UNECE units in there.)

PR #1301, opened 2026-05-28
chore(build): improve Dockerfile and build swagger docs

### LINKED ISSUE #998: generate swagger.yml in dockerfile (rswag)
add smth like `RAILS_ENV=test rails rswag && \` to the Dockerfile to gerenate swagger.yml at build time.

follow up for #969, Dockerfile seem's to be broken somehow in the current version now (old bundler stuff?) ...
Waiting for rails upgrade to move on with this

henning-unicode:
- use multi-stage Dockerfile
- add building of swagger docs to Dockerfile
- replace temp mysql server in assets build step with nulldb

fixes #998 

build time should be about the same as it is on master now

Still open:
- [ ] api-docs are currently served on the top level not scoped to foodcoops => other issue?
- [ ] further testing needed, especially with multi-foodcoop setups

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:3.4.6
+FROM ruby:3.4.6 AS base
 
 RUN supercronicUrl=https://github.com/aptible/supercronic/releases/download/v0.1.3/supercronic-linux-amd64 && \
     supercronicBin=/usr/local/bin/supercronic && \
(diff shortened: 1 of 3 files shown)


wvengen (commented):
I just did a quick read through, and to me it seems fine. It is definitely an improvement, using the nulldb adapter!

In other projects, I use e.g. `DATABASE_URL=postgresql:dummy`, but I think with Foodsoft there is more interaction with the database in places (single/multi foodcoop handling), which may invalidate that approach here. I would like to check that we really need the nulldb dependency?

In any case, smart!

henning-unicode:
> In other projects, I use e.g. `DATABASE_URL=postgresql:dummy`, but I think with Foodsoft there is more interaction with the database in places (single/multi foodcoop handling), which may invalidate that approach here. I would like to check that we really need the nulldb dependency?

I would also prefer to get it to work without an additional dependency. I tried setting `DATABASE_URL=mysql2://nohost` but that does not work as it tries connecting to the db when building the docker image:

```
11.60 /usr/src/app/vendor/bundle/ruby/3.4.0/gems/activerecord-7.2.3.1/lib/active_record/connection_adapters/mysql2_adapter.rb:33:in 'ActiveRecord::ConnectionAdapters::Mysql2Adapter.new_client': There is an issue connecting with your hostname: nohost. (ActiveRecord::DatabaseConnectionError)
11.60
11.60 Please check your database configuration and ensure there is a valid connection to your database
```

Either something in rails or deeper in our code base needs to be changed to get this working without nulldb. As we already used a temporary mysql server in the Dockerfile I think it is fine to use nulldb.

wvengen (approved):
Thanks for explaining.
Yes, definitely a step forward!

lentschi:
@henning-unicode 

Great, thank you! :star_struck: 

> further testing needed, especially with multi-foodcoop setups

I'll try to test it with a multicoop setting asap & then post my review!

lentschi (approved):
I tried it out locally and it works great 👍 

As to the points you still consider open:

>Still open:

> api-docs are currently served on the top level not scoped to foodcoops => other issue?

-> Yes I do think that deserves another issue and is not super-pressing.

> further testing needed, especially with multi-foodcoop setups

Well if you think so, I won't object - more testing is always good 🙂 
What I've successfully tested locally:

Booted up a multi-foodcoop prod setup (with clones of some real-life DBs).
-> I could access https://localhost/foodcoopname, login and navigate around. ✔️ 
-> I could access https://localhost/api-docs ✔️ 

Were there any multi-foodcoop issues you were expecting?

If you really do think further testing is in order, please mark the PR as "Draft" for now. (Otherwise I'll just merge it right away. 😉 )

henning-unicode:
@lentschi thanks! PR can be merged! I was just not sure how /api-docs is handled when you have multiple coops. But I did not change the endpoint as was remarked in the issue as this is another issue.

PR #1302, opened 2026-05-31
Fixes #1300 - Adapt unit translations to avoid duplicates

### LINKED ISSUE #1300: [BUG]: Double translations cause error in article upload
**Describe the bug**
When two different UNECE codes translate to the same article name, the article import throws an error when trying to reverse-map those units.

**To Reproduce**
see https://github.com/foodcoops/foodsoft/pull/1299

**Expected behavior**
No error is thrown. Import should just choose the first of the two UNECE units without complaining.

**Foodsoft Version**
- unreleased (upcoming 5.0.0)

lentschi:
Adapt unit translations to avoid duplicates - Fixes #1300

### DIFF
--- a/config/units-of-measure/locales/unece_de.yml
+++ b/config/units-of-measure/locales/unece_de.yml
@@ -28,7 +28,7 @@ unece_units:
   XAL:
     name: Ball
   XBN:
-    name: Ballen
+    name: Ballen, unkomprimiert
   XBF:
     name: Ballon
   XSY:
@@ -46,7 +46,7 @@ unece_units:
   XPO:
     name: Beutel
   XSH:
-    name: Beutel
+    name: Beutelchen
   XCB:
     name: Bierkasten
   X43:
@@ -80,7 +80,7 @@ unece_units:
     aliases:
     - bd
   XTS:
-    name: Bündel
+    name: Ballen
   XBE:
     name: Bündel
     abbreviation: Bdl.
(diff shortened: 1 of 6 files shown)


twothreenine:
I outlined my issues with this approach [here](https://github.com/foodcoops/foodsoft/issues/1300#issuecomment-4628321674).

Regardless from that, I see that you fixed some obviously incorrect translations regarding [YDK](https://github.com/foodcoops/foodsoft/pull/1302/changes#diff-4f0d7b821f5c1919eaedfb03ad5e495f96c3309218ce9f66985df99b847b3b02R978-R980) and [YDQ](https://github.com/foodcoops/foodsoft/pull/1302/changes#diff-4f0d7b821f5c1919eaedfb03ad5e495f96c3309218ce9f66985df99b847b3b02R1272-R1274). About the [5I](https://github.com/foodcoops/foodsoft/pull/1302/changes#diff-4f0d7b821f5c1919eaedfb03ad5e495f96c3309218ce9f66985df99b847b3b02R1257-R1259) one I'm not sure. If the approach is scrapped, these fixes should still be done.

lentschi:
@twothreenine I had responded to the discussion in the issue, but there weren't any replies lately.

So I'm bypassing rules and merging this as other reviews don't seem to be forthcoming and we have a demo coming up where this needs to be fixed.

We can simply revert/adapt this PR if the need should arise!

PR #1308, opened 2026-06-11
feat: add oidc login

firlevapz:
Adds support for generic OIDC login for e.g. keycloak, authentik or other SSO-providers


decentral1se:
@firlevapz wow, this would be an incredible development for Foodsoft. I think I've been trying to find someone to do this for like 6 years :sweat_smile:

Did you also factor in a "merging" step for existing accounts? Existing foodsoft deployments need a way to merge accounts on the SSO when they log in with the new SSO system for the first time.

I hope this can eventually be merged into Foodsoft One Day :tm: :pray:

PR #1311, opened 2026-06-13
On #766: Fixes bug with floating points in self-service

lentschi:
The self-service functionality had been cherry-picked from a pre-unit-changes state, where availability was always an integer. In the new state it threw a JS error, if availability included a float.


lentschi:

PR #1312, opened 2026-06-19
Fixes #1306 - incorrect article prices in order received mail

### LINKED ISSUE #1306: [BUG]: The calculated prices for products in the order confirmation emails are incorrect.
**Describe the bug**
The calculated prices for products in the order confirmation emails are not correct. It doesn't show the unit price in the email and seems to take the price for the supplier's unit (e.g. a box of bananas) instead of a kilo. The total price is correct (see screenshot).

Image

**Foodsoft Version**
4.9.1

lentschi:
Fixes #1306

Price calculation was invalid in the `order_received` mail: ordered group order amount was multiplied with price per supplier order unit, which only works if group order unit and supplier order unit don't differ.


lentschi:

PR #1314, opened 2026-06-19
Fixes #1304 - sync/upload: wrong ID was used to identify articles to delete

### LINKED ISSUE #1304: [BUG]: Article import: "Delete articles not in uploaded file." option not working
**Describe the bug**
Old articles not appearing in the newly uploaded file are not deleted, when checking the "Delete articles not in uploaded file" checkbox. 

**To Reproduce**
1. Go to Supplier/articles -> Articles -> Upload articles
2. Check "Delete articles not in uploaded file."
3. No error appears, only the old articles still appear in the list

**Expected behavior**
All old articles should be deleted, that do not appear in the uploaded file. As our offer changes every offer, I now first delete all the existing articles before uploading any new articles. This means that an unnecessary amount of new article entries keep being created every week and that the changes between the previous and current version of the articles do not show up.

**Screenshots**

**Foodsoft Version**
v4.9.1

**Logfile**
If applicable, add relevant log information.

**Additional context**

lentschi:
Fixes #1304

`articles.id` (incorrect, imprecise) instead of `article_versions.id` (correct) was used to identify articles to delete

### DIFF
--- a/app/views/articles/_sync.html.haml
+++ b/app/views/articles/_sync.html.haml
@@ -9,7 +9,7 @@
     %ul
       - @outlisted_articles.each_with_index do |article, index|
         %li
-          = hidden_field_tag "outlisted_articles[#{index}]", article.id
+          = hidden_field_tag "outlisted_articles[#{index}]", article.latest_article_version.id
           = article.name
           - if article.in_open_order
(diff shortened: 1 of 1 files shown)


lentschi:

PR #1321, opened 2026-06-27
Fixes #1319 - Removes bootstrap 2 workaround

### LINKED ISSUE #1319: [BUG]: Opening article modal behavior, esp. in Firefox
**Describe the bug**
In Firefox, the article modal sometimes opens partly outside the viewable screen area.

**To Reproduce**
1. Use Firefox (desktop)
2. Go to the articles menu
3. Go to a supplier's articles list (> 10 articles)
4. Scroll down *a bit* within the article list
5. Try to edit an article: The modal opens partly outside the screen, so you can only interact with the bottom part of the modal, and can't scroll up. Only if you've scrolled _so far_ down in the articles list that the modal opens completely outside the viewable area, then the page scrolls up automatically.

In Chrome, the page always scrolls up to the top, where the modal is.
I didn't test other browsers.

**Expected behavior**
The modal should open in the viewable area. Ideally, the page shouldn't even have to scroll up (if you go through a list of articles, you don't want to lose focus where you are in the list).

**Screenshots**
Image

**Foodsoft Version**
What Version of Foodsoft are you running (can be found in the bottom-right). Are you running a modified Fork?
- v5.0.0-rc0 (Austrian demo deployment)

**Potential solution**
I noticed that other modals, like in Admin/Finance/Bank Gateways open in the viewable area, without the page scrolling up. There, the modalContainer div use `class="modal fade in"`, while the article modals use `class="modal fade modal-xl in"`. Removing the `modal-xl` class (in browser tools) puts the modal right in the viewable area. I don't see any size difference of the modal.

lentschi:
Fixes #1319 by removing a bootstrap 2 workaround that is no longer required for bootstrap 3 and even caused issues with it.

### DIFF
--- a/app/assets/stylesheets/bootstrap_and_overrides.css.scss
+++ b/app/assets/stylesheets/bootstrap_and_overrides.css.scss
@@ -558,16 +558,6 @@ i.package.icon-only {
   margin: 0;
 }
 
-// bootstrap 2 doesn't support differnt modal sizes -> let's provide our own modal-xl:
-.modal.modal-xl {
-  width: 80vw;
-  max-width: 650px;
-  position: absolute;
-  left: 0;
-  right: 0;
-  margin: 0 auto;
-}
-
 // allow to have indicator text instead of input with same markup
 .control-text {
   margin-top: 5px;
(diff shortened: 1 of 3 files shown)


lentschi:

PR #1323, opened 2026-06-28
Fix privilege escalation via mass assignment in workgroup update

tonghuaroot:
## Problem

`Foodcoop::WorkgroupsController#update` passes `params[:workgroup]` straight to `update`, and the app sets `config.action_controller.permit_all_parameters = true`, so every nested parameter is treated as permitted. The action is reachable by any member of the workgroup (`authenticate_membership_or_admin` admits members, not just admins), and `role_admin` is a plain boolean column on `groups`.

A member can therefore submit `workgroup[role_admin]=true` and grant their workgroup the global admin role. `User#role_admin?` is true if any of the user's groups has `role_admin`, and `authenticate('admin')` gates every administrative function on it, so this is a vertical escalation from an ordinary workgroup member to site administrator. The same primitive can also set `role_finance`, `role_orders`, `role_article_meta` and `role_suppliers`.

## Fix

Restrict the assignment to the fields the member edit form (`shared/_group_form_fields`) actually exposes (`name`, `description`, `user_tokens`) with a strong-parameters allowlist. This matches the pattern already used by `home_controller#update_profile`. Role assignment stays in the admin area (`Admin::WorkgroupsController`, gated by `authenticate_admin`).

Removing the global `permit_all_parameters = true` and giving each controller an explicit allowlist would close this whole class of exposure (the comment in `config/application.rb` already marks that as a TODO), but that is a larger change; this PR fixes the reachable escalation directly.

## Verification

Reproduced and fixed end-to-end against a local deploy from `docker-compose-dev.yml` (Ruby 3.4.5, Rails 7.2, MariaDB, `db:seed`). As a non-admin member of one workgroup, `PATCH .../foodcoop/workgroups/:id` with `workgroup[role_admin]=true`:

- before the fix: the flag persists, the member's session reaches `/:foodcoop/admin`;
- after the fix: `role_admin` is dropped, the workgroup stays non-admin, and `name`/`description`/`user_tokens` edits still work.

I reported this privately by email; opening the fix here as suggested in that thread.

### DIFF
--- a/app/controllers/foodcoop/workgroups_controller.rb
+++ b/app/controllers/foodcoop/workgroups_controller.rb
@@ -12,10 +12,16 @@ class Foodcoop::WorkgroupsController < ApplicationController
 
   def update
     @workgroup = Workgroup.find(params[:id])
-    if @workgroup.update(params[:workgroup])
+    if @workgroup.update(workgroup_params)
       redirect_to foodcoop_workgroups_url, notice: I18n.t('workgroups.update.notice')
     else
       render action: 'edit'
     end
+  end
+
+  private
+
+  def workgroup_params
+    params.require(:workgroup).permit(:name, :description, :user_tokens)
   end
 end


lentschi (approved):

lentschi:
@tonghuaroot Thank you _very much_ for raising this issue, your profound analysis and your fix! :heart: 
I merged it into `master` and cherry-picked it into `4.9.x` . Also I just released [v4.9.2](https://github.com/foodcoops/foodsoft/releases/tag/v4.9.2) which contains your fix.

PR #1327, opened 2026-07-03
Fixes #1326 - Article versioning: reinstate "global price"

### LINKED ISSUE #1326: [BUG]: Some article prices changed after migrating to 5.0.0-rc0
**Describe the bug**
Looking at a supplier's articles, _without_ having "migrated" any old units yet (so, all articles have custom units for now), I see that some articles have different prices than the same articles in our v4.8.1 instance.

For example, this supplier (krautkoopf/suppliers/131 if you have db access).

in v4.8.1:
Image
Image

in v5.0.0-rc0:
Image
Image

We've had only one order for this supplier, which was settled in May, all before the migration to 5.0.
After delivery (mid-May), the price of `Sura Kees` (article ID 10216) was changed from `4.75` to `5.59` in the accounting menu, without changing the global price. But after migration to 5.0, this price change was applied to the global price.

I suppose that, during the migration to 5.0, the latest price is taken as global price, even if it is not the global price.

I also tried changing prices in the accounting menu in 5.0 and settling orders. There weren't any unwanted changes in global prices, so I guess this really just happens during migration.

**Foodsoft Version**
v5.0.0-rc0

lentschi:
Fixes #1326 

Article versioning: reinstate "global price"

Reasoning see here:
https://github.com/foodcoops/foodsoft/issues/1326#issuecomment-4874677721

### DIFF
--- a/db/migrate/20240726083740_alter_articles_add_versioning.rb
+++ b/db/migrate/20240726083740_alter_articles_add_versioning.rb
@@ -73,6 +73,66 @@ class AlterArticlesAddVersioning < ActiveRecord::Migration[5.2]
       ))
     end
 
+    # create versions resetting price/tax/deposit/unit_quantity to what (until now) was
+    # referred to as "global price", if the latest version differs:
+    articles = select_all(%(
+      SELECT articles.*,
+        latest_article_versions.price AS latest_price,
+        latest_article_versions.tax AS latest_tax,
+        latest_article_versions.deposit AS latest_deposit,
+        latest_article_versions.unit_quantity AS latest_unit_quantity,
+        latest_article_versions.created_at AS latest_modification
+      FROM articles
+      LEFT JOIN (
+        SELECT `article_versions`.*
+        FROM `article_versions`
+        LEFT OUTER JOIN article_versions later_article_versions ON later_article_versions.article_id = article_versions.article_id
+          AND later_article_versions.created_at > article_versions.created_at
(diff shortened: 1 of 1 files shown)


henning-unicode (commented):
SQL look good. You could test that it runs on sample data that would cause an issue without this change?

lentschi:
@henning-unicode 

> You could test that it runs on sample data that would cause an issue without this change?

Well I tested it manually. But I guess you mean, I should write an rspec test for it, right?

Do you have an idea as for how to do that? :thinking: 

I had thought about it - as the migration is quite complex and I feel it _should_ be tested - but I don't see _how_ to best do it:

One could easily test a pure data migration by preparing test data and then calling `migration.up`.
However, this migration is a mixture of data and structure changes that are intertwined and I don't see a way to unravel that. 

So the only thing I can think of is copying the schema as it is before this migration, bring that up in an rspec `before do` by dropping all tables first, fill it with faulty data, test `migration.up`, check if everything is fine and finally run `ActiveRecord::Tasks::DatabaseTasks.load_schema_current(:test)` in `after do` to keep all the tests after this one from crashing.

In any case, while this bugfix comparatively small and the test for it would be quite complex (plus it should also include other data migrations): Would it be okay for you to postpone the test, moving it to a separate 5.0 issue?

henning-unicode:
> Well I tested it manually. But I guess you mean, I should write an rspec test for it, right?

No I just meant manual testing :)

henning-unicode (approved):

PR #1331, opened 2026-07-11
Fixes #1329 - Increase scale and precision of the `result` in `group_order_articles`

### LINKED ISSUE #1329: [BUG]: Precision error in group order results
**Describe the bug**
When changing the received amounts per order group sometimes a slightly different is stored and displayed (rounding issue).

*Internal Reference:* Originally reported in the [Austrian discourse forum](https://forum.foodcoops.at/t/bestellabrechnung-menge-aendern/10354/2).

**To Reproduce**
1. Create an article with a `supplier_order_unit` of piece.
2. Set its contents to 230g and the `billing_unit` to gram.
3. Order something (doesn't really matter how much and how many order groups) of this article and close the order.
4. In billing, expand the article and enter `242`  for one of the order groups.

-> Value is automatically changed to `241.96` ❌

**Screen video**

*Note:* The server seemed to have been quite slow when I recorded this -> watch till the end 😉 :

https://github.com/user-attachments/assets/43cc4a77-7df8-4d2e-9514-600dd44ed7c6

**Expected behavior**
Value stays `240`. ✔️ 

**Foodsoft Version**
- 5.0.0-rc0

**Additional context**
See discussion and linked workarounds in https://github.com/foodcoops/foodsoft/issues/1092 as well as another fix in https://github.com/foodcoops/foodsoft/pull/1291 - I guess we'll need to do this for `group_order_articles.result` as well (and probably for other fields...?)

lentschi:
Fixes #1329: Increase scale and precision of `group_order_articles.result` and `group_order_articles.result_computed`

Also see discussion about possible better solutions to this in #1092

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2025_06_01_093453) do
+ActiveRecord::Schema[7.2].define(version: 2026_07_11_191316) do
   create_table "action_text_rich_texts", charset: "utf8mb4", collation: "utf8mb4_general_ci", force: :cascade do |t|
     t.string "name", null: false
     t.text "body", size: :long
@@ -75,7 +75,7 @@ ActiveRecord::Schema[7.2].define(version: 2025_06_01_093453) do
 
   create_table "article_versions", id: :integer, charset: "utf8mb4", collation: "utf8mb4_general_ci", force: :cascade do |t|
     t.integer "article_id", null: false
-    t.decimal "price", precision: 12, scale: 6, default: "0.0", null: false, comment: "stored in `article_versions.supplier_order_unit`"
+    t.decimal "price", precision: 11, scale: 6, default: "0.0", null: false, comment: "stored in `article_versions.supplier_order_unit`"
     t.decimal "tax", precision: 8, scale: 2, default: "0.0", null: false
     t.decimal "deposit", precision: 8, scale: 2, default: "0.0", null: false
     t.datetime "created_at"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -75,7 +75,7 @@
 
   create_table "article_versions", id: :integer, charset: "utf8mb4", collation: "utf8mb4_general_ci", force: :cascade do |t|
     t.integer "article_id", null: false
-    t.decimal "price", precision: 12, scale: 6, default: "0.0", null: false, comment: "stored in `article_versions.supplier_order_unit`"
+    t.decimal "price", precision: 11, scale: 6, default: "0.0", null: false, comment: "stored in `article_versions.supplier_order_unit`"
```
henning-unicode: Why is ``price`` (and other fields further down) changed? The migration only changes ``result`` and ``result_computed``?

henning-unicode:
Can not test as I can not find the path to reproduce it. 

Issue mentioned:

> In billing, expand the article and enter 242 for one of the order groups.

What does "In billing" mean? I clicked on accounting from the managed orders view, but that seems wrong. Where and for what field should "242" be entered?

db/schema.rb
lentschi: Ah, thanks for spotting that! :pray: 

> Why is price (and other fields further down) changed?

My local development db seems to have been in an invalid state which is why persisting the schema produced this result.

Should be fixed now.

lentschi:
> Where and for what field should "242" be entered?

Yeah, sorry that was maybe a bit unclear - I now added a short screen video to the issue description.

henning-unicode:
> > Where and for what field should "242" be entered?
> 
> Yeah, sorry that was maybe a bit unclear - I now added a short screen video to the issue description.

Thanks! I could test it now and confirm that it fixes the issue.

henning-unicode (approved):

PR #1335, opened 2026-07-13
Fix article price field layout

### LINKED ISSUE #1332: [BUG]: Edit all articles: article price layout  broken
**Describe the bug**
In the "edit all articles" form the price input and dropdown have a faulty layout.

**To Reproduce**
1. Open "edit all articles" form (or "synchronize" which is the same basically) for any supplier.

-> The price input is too narrow and the open arrow of the dropdown is shown underneath the contained option text.

**Expected behavior**
Price input wide enough for reasonable prices like 12.45€. Dropdown arrow not overlapping text underneath.

**Screenshots**
After the bootstrap upgrade:

Image

Before the bootstrap upgrade:

Image

**Foodsoft Version**
- 5.0.0-rc0

**Additional context**
This seems to be happening since https://github.com/foodcoops/foodsoft/pull/1146.
@henning-unicode Could you maybe take a look at this?

aouxwoux:
Fixes #1332.

The bulk edit and synchronization tables now reserve enough space for normal price values and keep the price-unit selector from collapsing in the flex row.

Validation:
- `git diff --check`

### DIFF
--- a/app/assets/stylesheets/article.scss
+++ b/app/assets/stylesheets/article.scss
@@ -36,3 +36,11 @@
 .sync-table .extra-unit-fields .control-group {
   margin-bottom: 20px;
 }
+
+.input-group .article-price-input {
+  width: 6em;
+}
+
+.article-price-unit {
+  min-width: 7em;
+}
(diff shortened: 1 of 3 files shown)


lentschi (changes requested):
@aouxwoux Thank you for your contribution! :heart: 

It's better, but still not perfect:
image
<- The dropdown should be about the same width as the one in the unit column. If the selection options still won't fit, it should use an ellipsis just as the one in the unit column does.

aouxwoux:
oh yeah that makes sense, will push a change

PR #1336, opened 2026-07-17
Fix CVE-2026-54904: Upgrade ruby to 3.4.7

lentschi:
see https://github.com/advisories/GHSA-h8w8-99g7-qmvj Concurrent::AtomicReference#update can enter a permanent busy retry loop when the current value is Float::NAN.


lentschi:
Critical fix, so I'm merging it without waiting for a review

PR #1340, opened 2026-07-21
Fixes #1330 - Invalid space before conversion button

### LINKED ISSUE #1330: [BUG]: Group order create/edit form: Invalid space before conversion button
**Describe the bug**
Group order: Layout of amount controls appear broken depending on column content and screen width.

**To Reproduce**
1. Create a supplier with a single article with few characters in its name (e.g. "Test").
2. Create an order for this supplier.
3. Open the new group order form in a relatively wide window (> 1800px).

-> Layout breaks (Note the space before the convert button)

**Expected behavior**
Layout doesn't break.

**Screenshots**
Image

**Foodsoft Version**
- 5.0.0-rc0

**Additional context**
I assume this is only happening since https://github.com/foodcoops/foodsoft/pull/1146, but haven't checked yet. (I think I partially fixed stuff in https://github.com/foodcoops/foodsoft/pull/1256 - maybe this is a side-effect - didn't check that either.)
@henning-unicode Could you maybe take a look at this?

veronika-unicode:
Fixes #1330

Buttons and input field are now displayed without white space

grafik

### DIFF
--- a/app/assets/stylesheets/bootstrap_and_overrides.css.scss
+++ b/app/assets/stylesheets/bootstrap_and_overrides.css.scss
@@ -759,6 +759,10 @@ td.mr-1 > *:last-child {
   display: block;
 }
 
+.numeric-step {
+  min-width: 50px;
+}
+
 
 .form-group .select2-container {
   width: 200px;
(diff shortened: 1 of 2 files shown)


lentschi (changes requested):
I'm not against hacky `width: xx !important` solutions in general as far as foodsoft is concerned (After all, the whole FE needs to be rewritten soon, as we discussed elsewhere 😅 ).

However, this hack seems to make matters worse (Conversion button disappears when window is resized):

https://github.com/user-attachments/assets/1f0eff86-40b0-43fe-8c06-17ae76a17e38

henning-unicode:
@lentschi I hope that I've fixed it. Thanks for reviewing!

lentschi (approved):

PR #1346, opened 2026-08-05
fix: do not display order url if only email order method available

### LINKED ISSUE #1343: Hide field "Order URL" if email is the only order method
In #1187, sending orders via FTP was introduced as a plugin. As long as no other order methods are enabled, email remains the only available order method. Therefore, the [Order URL](https://github.com/foodcoops/foodsoft/blob/master/app/views/suppliers/_form.html.haml#L24) field serves no purpose and should be hidden to avoid confusing users.

Image

henning-unicode:
Fixes #1343

A LLM was used to create this PR. Code was reviewed and tested.

Please test @steffen-unicode 

### How to test

- Open a edit or new form for a supplier
- Only the method dropdown should be displayed not the Order Url field
---
- Activate b85 plugin
- Open a edit or new form for a supplier
- Order method dropdown should include FTP/b85 and Order URL field should be displayed if FTP/b85 is selected.
- Set order url
- Save
- Order method and url should be set.
---
- Edit supplier again
- Set method to email
- Order url field should be hidden

### DIFF
--- a/app/views/suppliers/_form.html.haml
+++ b/app/views/suppliers/_form.html.haml
@@ -26,3 +26,15 @@
   %div
     = f.button :submit
     = link_to t('ui.or_cancel'), suppliers_path
+
+- content_for :javascript do
+  :javascript
+    $(function() {
+      var urlFormGroup$ = $('#supplier_remote_order_url').closest('.form-group');
+      var methodSelect$ = $('#supplier_remote_order_method');
+      var toggleUrlField = function() {
+        urlFormGroup$.toggle(methodSelect$.val() !== 'email');
+      };
+      methodSelect$.on('change', toggleUrlField);
+      toggleUrlField();
+    });


lentschi (commented):
Hm... Just a note: This still wouldn't disable/hide the order URL field if the `B85` extension was enabled, but the current supplier had `:email` selected, right?
(It would fulfill the requirements as per issue, but I guess hiding/showing the URL field by javascript depending on which value is selected as `remote_order_method`, would even be better...? In addition, the select itself could be hidden, if there's only one value to choose from.)

henning-unicode:
@lentschi thanks for the idea. I changed it to an implementation using javascript. The url field is now hidden/shown depending on the selected order method.

order url is not cleared if it was set but order method is changed to email, as there seem to be tests that expect this behaviour.

lentschi (approved):

